PR #5, opened 2018-12-10
Capabilities: Binding controller

jmprusi:
Initial code for the binding controller, still work in progress. 

This controller takes the responsibility of traversing the whole dependency tree of CRDs and consolidating it into a single "Consolidated" CR object for each Binding object.

This requires some refactoring to split it into several different functions with proper meaning.

TODO:
- [x] Add support for Plans CRD
- [x] Add support for Limits CRD
- [x] Merge the IntegrationMethod object and the API object
- [x] Add support for Mapping Rules CRD
- [x] Find the proper way for testing operators and write tests
- [x] If the consolidated object exists: Check if it needs to be updated.
- [x] Refactoring: everything is inside the reconciliation loop, extract that logic into its own package/functions
- [x] Add Sort functions to compare 
- [ ] Set proper status for failed, or already reconciled.
- [ ] Docs: How to run, test. 
- [ ] Docs: CRDs examples.


jmprusi:
I need to improve several things and "sort" the structs so we can compare them and document it. But due to this PR is big enough already (sorry!), I would like some "early" review! :) @miguelsorianod  @eguzki

pkg/controller/binding/binding_controller.go
```diff
@@ -0,0 +1,215 @@
...
+			log.Printf("Reconcile: Consolidated config %s/%s has been updated.", currentConsolidated.Namespace, currentConsolidated.Name)
+		}
+	}
+	// Return without errors and resume the reconcile loop
+	return reconcile.Result{}, nil
+}
+
+func compareConsolidated(consolidatedA, consolidatedB *apiv1alpha1.Consolidated) bool {
+	//Let's compare only the Spec
+	A, _ := json.Marshal(consolidatedA.Spec)
```
eguzki: why serialize to json byte stream? `reflect.DeepEqual` should compare structs deeply.

eguzki: does not apply. For our structs, two array of elements may be equal even when do not follow same order. Conceptually `Set` data structure should be used, unfortunately not available in Go.

jmprusi: Yup! we need to write sort functions for those, check: https://github.com/3scale/api-operator/blob/4856b4d1f34fdf22f4202ffc1e4447166983b75f/pkg/apis/3scale/v1alpha1/types.go#L55

PR #10, opened 2018-12-13
Upgrade operator-sdk to 0.3.0

miguelsorianod:
This PR upgrades operator-sdk version to 0.3.0.

The PR Includes upgrade changes to:
* The skeleton source-code of the operator, not related to custom CRD-specific code
* The CRD-specific code
* Auto-generated templates. They have changed because the apimachinery library has been updated as part of the operator-sdk new dependency versions

It seems it is currently not trivial to perform an upgrade of the operator-sdk application (do not confuse it with the operator-sdk itself, which the installation of it is done OUTSIDE of this repository and is up to each user of this repository to install it by himself/herself).
Even when updating the Gopkg.toml and Gopkg.lock with an upgraded version of the operator-sdk, this does NOT change the existing operator-sdk application source code (the skeleton or possible changes that might be required in the CRDs as a consequence of an upgrade of the operator-sdk version).

The way I found to upgrade it has been:
* I copied my local up-to-date .git repository to a different directory
* I removed my local copy . The reason I've done this is because the only way that seems to be possible right now to have a new operator-sdk application source code skeleton is by recreating the project, and you cannot recreate the project with an existing name.
* I recreated the operator-sdk application by executing 'operator-sdk new 3scale-operator'
* I moved the local .git repository to the newly created operator-sdk application
* I restored removed files that were expected to be done by us (for example the pkg/3scale directory) and other files like Makefile, README.md, ... that are not created when you create a new operator-sdk application
* Because the AMP custom resource CRD did not have any content yet, I have not restored it and I recreated it again from zero but now using the new operator-sdk version. In this way we can know what has changed at CRD-specific level when you have an empty CRD cc @jmprusi this might interest you when migrating the API operator to 0.3.0 (I separated the CRD-specific changes in their own commit)

Also, remember to update your locally installed operator-sdk repository pointing to 0.3.0 and installing the new binary version.

If you think/know there's a better way of performing the upgrade process in future version upgrades please let me know. Upgrading will be much harder when we have more CRD-specific code.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,12 +2,12 @@
 
 
 [[projects]]
-  digest = "1:c4638ce2e73b35bb3b0a109d687d798159a1844e1174a80c1cc6d1895d2f6fe0"
+  digest = "1:fd1a7ca82682444a45424f6af37b1e0373f632e5a303441b111558ae8656a9b7"
   name = "cloud.google.com/go"
   packages = ["compute/metadata"]
   pruneopts = "NT"
-  revision = "74b12019e2aa53ec27882158f59192d7cd6d1998"
-  version = "v0.33.1"
+  revision = "0ebda48a7f143b1cce9eb37a8c1106ac762a3430"
+  version = "v0.34.0"
 
 [[projects]]
   digest = "1:75d2b55b13298745ec068057251d05d65bbae0a668201fe45ad6986551a55601"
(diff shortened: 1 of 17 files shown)


miguelsorianod:
I opened an issue to the operator-sdk project https://github.com/operator-framework/operator-sdk/issues/852 because it seems that after updating the Gopkg.toml versions the vendored operator-sdk command fails to execute the 'build' subcommand. The same problems happens if I create a new empty operator-sdk application

miguelsorianod:
Closing PR because operator-sdk v0.3.0 upgrades the kubernetes API version to 1.12 and at this moment we need to support Kubernetes 1.11 (Used by OpenShift 3.11), so we'll stick with operator-sdk v0.2.1 for the moment. The following PR contains the necessary changes to fully support v0.2.1 operator-sdk https://github.com/3scale/3scale-operator/pull/14

PR #14, opened 2019-01-21
Change operator dependencies

miguelsorianod:
This PR:
 * Fixes some minor operator-sdk v0.2.1 direct dependency. operator-sdk in master was already v0.2.1 in master but not all direct operator-sdk dependencies matched what the "operator-sdk print-deps --as-file" generated
 * Removes unnecessary imports from Gopkg.toml
 * Fixes openshift/api version in Gopkg.toml to the revision commit that currently matches the v3.11 branch (no tag exists in the openshift/api repository)
 * Regenerates Gopkg.toml
 * Updates Makefile to use a locally installed operator-sdk command instead of a vendored one. This has been done because it is not supported by operator-sdk (see https://github.com/operator-framework/operator-sdk/issues/852). This breaks our current CI/CD integration and in a future PR will be fixed
 * Updates README.md with an explanation of prerequisites at the time of writing this and a quick start section

PR #15, opened 2019-01-21
Local operator sdk command on ci

miguelsorianod:
This PR adds a run step to the CircleCI 'build' job that installs operator-sdk v0.2.1 tag locally, which is the n used to be able to use the operator-sdk command.
It also updates all CircleCI Docker images that use Go to 1.11.

This PR is based on the contents of the branch that has been requested to be merged into master in the https://github.com/3scale/3scale-operator/pull/14 PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,6 +91,16 @@ jobs:
             - v1-gopkg-cache-{{ arch }}-{{ .Branch }}
 
       - run:
+          name: Install operator-sdk
+          command: |
+              mkdir -p ${GOPATH}/src/github.com/operator-framework
+              cd ${GOPATH}/src/github.com/operator-framework
+              git clone https://github.com/operator-framework/operator-sdk --branch v0.2.1
+              cd operator-sdk
+              make dep
+              make install
+
+      - run:
           name: Install dependencies
           command: |
             make vendor
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -139,7 +139,7 @@ jobs:
             172.30.1.1:5000/$project/amp-system:latest=quay.io/3scale/porta:nightly --insecure
   generator:
     docker:
-      - image: circleci/golang:1.10
+      - image: circleci/golang:1.11
     working_directory: /go/src/github.com/3scale/3scale-operator
     steps:
       - checkout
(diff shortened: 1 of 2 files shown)


miguelsorianod:
As an offtopic theme, I keep seeing that sometimes the 'generator' CircleCI job that generates the templates fails. It fails because sometimes has an OOM when executing the following command:
make clean test -j 2 --directory pkg/3scale/amp

This command should limit the number of tasks to execute at the same time to 2, but debugging the CircleCI job I've seen that it does not limit it and it executes a number of huge jobs at the same time. After investigating I've seen that CircleCI runs GNU Make 4.1, whereas in my machine (a Fedora 29), GNU Make 4.2 is being executed.
After looking at differences in the versions of GNU Make 4.1 and 4.2 it seems that in 4.1 the -j flag value is ignored if MAKEFLAGS is set whereas in 4.2 that same value is used. Looking at the Make 4.2 changelog it seems that there was a bugfix related to the -j flag and MAKEFLAGS variable that could be related:
https://lists.gnu.org/archive/html/info-gnu/2016-05/msg00013.html

.circleci/config.yml
```diff
@@ -90,6 +90,17 @@ jobs:
             - v1-gopkg-cache-{{ arch }}-{{ checksum "Gopkg.lock" }}
             - v1-gopkg-cache-{{ arch }}-{{ .Branch }}
 
+      - run:
+          name: Install operator-sdk
+          command: |
+              mkdir -p ${GOPATH}/src/github.com/operator-framework
+              cd ${GOPATH}/src/github.com/operator-framework
+              git clone https://github.com/operator-framework/operator-sdk
```
eguzki: if you want to save `git checkout`, `git clone` has --branch flag where tags are also taken. Up to you, it is ok also to do clone and checkout.

eguzki:
check why tests failed and then merge

.circleci/config.yml
miguelsorianod: Changed. Using --branch leaves the status in HEAD dettached state, but it's not a problem if using it only on the CI because it is not going to be modified.

miguelsorianod:
> check why tests failed and then merge

The tests failed because of the Makefile problem with -j flag and MAKEFLAGS variable commented before. I removed the MAKEFLAGS variable that tries to auto-detect number of CPUs and should work correctly. Now if you want to execute the make rule locally with parallelism you have to specify -j <desiredparallelism> when executing Make

PR #16, opened 2019-01-22
Upgrade contents of files that change from the upgrade to v0.2.1

miguelsorianod:
Some files change when a new operator project is created in
v0.2.1 compared to previous versions (a notable change
is the change of logging methods). This commit updates
our existing files to match what is created in v0.2.1 for
non api/controlled related files.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,5 +1,7 @@
-FROM alpine:3.6
+FROM alpine:3.8
+
+RUN apk upgrade --update --no-cache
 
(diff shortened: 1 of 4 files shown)


miguelsorianod:
Logging in v0.2.1 uses new logging methods. Controllers code should use the new logging methods. Documentation on this change and the new method names can be seen in the following document:
https://github.com/operator-framework/operator-sdk/blob/master/doc/user/logging.md

PR #18, opened 2019-01-29
Refactor component options

miguelsorianod:
Refactor generator component options into required and nonrequired
    
Refactor options for apicast, backend, ha, memcached,
mysql, productized, redis, s3, system and wildcardrouter
components.

This has been done because at operator level you need to be able to set the secrets fields when constructing the components, to be able to reconcile the possible secrets that already exist.

### DIFF
--- a/pkg/3scale/amp/component/system.go
+++ b/pkg/3scale/amp/component/system.go
@@ -20,6 +20,11 @@ type System struct {
 }
 
 type SystemOptions struct {
+	systemNonRequiredOptions
+	systemRequiredOptions
+}
+
+type systemRequiredOptions struct {
 	adminAccessToken    string
 	adminPassword       string
 	adminUsername       string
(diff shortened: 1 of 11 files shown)


pkg/3scale/amp/component/system.go
```diff
@@ -133,59 +147,128 @@ func (s *SystemOptionsBuilder) StorageClassName(storageClassName *string) {
...
 	if s.options.mysqlRootPassword == "" {
-		return nil, fmt.Errorf("no mysql root password has been provided")
+		return fmt.Errorf("no mysql root password has been provided")
+	}
+
+	return nil
+}
+
+func (s *SystemOptionsBuilder) setNonRequiredOptions() {
+	defaultDatabaseURL := "mysql2://root:" + s.options.mysqlRootPassword + "@system-mysql/" + s.options.mysqlDatabaseName
```
eguzki: use `fmt.Sprintf`?

PR #19, opened 2019-01-29
Move secret field names and names into constants

miguelsorianod:
Refactor secret names and field names into constants
for apicast, backend, s3, system and zync components

### DIFF
--- a/pkg/3scale/amp/component/apicast.go
+++ b/pkg/3scale/amp/component/apicast.go
@@ -13,6 +13,12 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+const (
+	ApicastSecretRedisSecretName             = "apicast-redis"
+	ApicastSecretRedisProductionURLFieldName = "PRODUCTION_URL"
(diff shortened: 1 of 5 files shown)


pkg/3scale/amp/component/apicast.go
```diff
@@ -13,6 +13,12 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+const (
+	ApicastSecretRedisSecretName             = "apicast-redis"
+	ApicastSecretRedisProductionURLFieldName = "PRODUCTION_URL"
+	ApicastSecretRedisStagingURLFieldName    = "STAGING_URL"
```
eguzki: we could live without `Apicast` prefix, right? it is in `apicast.go` file, so context gives the remaining info. Specially because I see this constant is used just in this file

eguzki: same principle applies to other files

eguzki: does not apply because there could be name conflict as they all belong to the same package :+1:

PR #33, opened 2019-02-12
Add basic integration install e2e tests for APIManager

miguelsorianod:


pkg/apis/apps/v1alpha1/register.go
```diff
@@ -10,6 +10,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/runtime/scheme"
 )
 
+const (
+	APIManagerKind = "APIManager"
```
eguzki: ???

PR #37, opened 2019-02-13
Add APIManager full OpenAPI validation from current Go contents

miguelsorianod:


miguelsorianod:
Adds OpenAPI validation from current APIManager Go contents.

The fields of the Go contents will be reorganized but for the moment let's just add validation for how they currently are

PR #48, opened 2019-02-20
doc tenant resource types and secrets

eguzki:


miguelsorianod:
Seems good to me but if we are going to create a different reference document for each part of the operator then I'd use the 'reference.md' document to reference the others or put some link to the corresponding documents somewhere

eguzki:
can you elaborate? not sure about the changes you are proposing...

miguelsorianod:
I was referring that we should either:
 * Join everything in the same reference.md file
 * Move everything into its own reference-<specific>.md file. If this is the case we should reference this specific files from the main reference.md file, or from the main README.md or from the user-guide.

We'll change this in a future PR

Issue #31, opened 2019-02-11
Update 3scale labels used on AMP templates to be prometheus compliant


slopezz:
At the moment we are using the following labels on openshift objects created by 3scale AMP templates (example at https://github.com/3scale/3scale-amp-openshift-templates/blob/master/amp/amp-eval-tech-preview.yml#L3200) :

```
    labels:
      3scale.component: apicast
      3scale.component-element: staging
```

Which are correct from kubernetes point of view: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set 

```
Labels are key/value pairs. Valid label keys have two segments: an optional prefix and name, separated by a slash (/). The name segment is required and must be 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between. The prefix is optional. If specified, the prefix must be a DNS subdomain: a series of DNS labels separated by dots (.), not longer than 253 characters in total, followed by a slash (/).
```

But are not valid from prometheus point of view: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels 

```
The metric name specifies the general feature of a system that is measured (e.g. http_requests_total - the total number of HTTP requests received). It may contain ASCII letters and digits, as well as underscores and colons. It must match the regex [a-zA-Z_:][a-zA-Z0-9_:]*.

Label names may contain ASCII letters, numbers, as well as underscores. They must match the regex [a-zA-Z_][a-zA-Z0-9_]*. Label names beginning with __ are reserved for internal use.
```

And it is advisable using kubernetes labels that both complies kubernetes and prometheus syntax.

*What test I have done to validate it*

I added a service with label "3scale.component: system" (starting by digit "3" and containing a point "."):

```
apiVersion: v1
kind: Service
metadata:
  name: system-sphinx-test-regex-exporter
  namespace: prometheus-exporters
  labels:
    3scale.component: system
    app: sphinx-exporter
    environment: test
    name: system-sphinx-test-regex-exporter
    role: service
    template: sphinx-exporter
    tier: sphinx
  annotations:
    openshift.io/generated-by: OpenShiftNewApp
    prometheus.io/path: /metrics
    prometheus.io/port: '9247'
    prometheus.io/scrape: 'true'
spec:
  ports:
    - protocol: TCP
      port: 9247
      targetPort: 9247
  selector:
    app: system-sphinx-exporter
[...]

PR #60, opened 2019-02-26
Relabel 3scale elements

miguelsorianod:
This comes from the issue https://github.com/3scale/3scale-operator/issues/31.

This change renames the 3scale.component and 3scale.component-element labels to threescale_component and threescale_component_element labels to be compatible with Prometheus labels datamodel.

Adding this would make those labels compatible with prometheus. However, as you can see in the following Kubernetes link https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set , some automated Kubernetes elements/controllers add labels that use / as some part of the label, which is incompatible with the Prometheus data model. Our operator (which has controllers )also  might need (and probably will) to add labels to perform object management.
So in my opinion what really solves the reported problem in #31 is applying the appropiate relabel scrape config depending on the desired needs. I have no problem in applying this PR but that would not really solve the reported problem because potentially labels incompatible with the Prometheus datamodel could appear as explained.

Also cc @slopezz

### DIFF
--- a/pkg/3scale/amp/auto-generated-templates/amp/amp-eval-s3.yml
+++ b/pkg/3scale/amp/auto-generated-templates/amp/amp-eval-s3.yml
@@ -19,8 +19,8 @@ objects:
       openshift.io/display-name: AMP backend
     creationTimestamp: null
     labels:
-      3scale.component: backend
       app: ${APP_LABEL}
+      threescale_component: backend
     name: amp-backend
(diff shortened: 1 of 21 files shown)


miguelsorianod:
For the moment we are going to merge this change so we reduce the possibility of the problem happening, but as we discussed we should raise awareness on this.

PR #67, opened 2019-02-28
fix tenant doc

eguzki:


mikz:
Maybe it could be worth using the same tool as APIcast to verify the links in documentation as part of the CI process.

PR #74, opened 2019-03-05
Apimanager secret reconciliation changes

miguelsorianod:
This PR makes secrets not being updated when no differences are detected. Differences are calculated from a set of specific fields that are taken into account

### DIFF
--- a/pkg/controller/apimanager/apimanager_controller.go
+++ b/pkg/controller/apimanager/apimanager_controller.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"reflect"
 
+	"github.com/go-logr/logr"
+
 	"github.com/3scale/3scale-operator/pkg/3scale/amp/component"
 	"github.com/3scale/3scale-operator/pkg/3scale/amp/operator"
 	"github.com/3scale/3scale-operator/pkg/3scale/amp/product"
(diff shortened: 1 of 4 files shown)


pkg/controller/apimanager/apimanager_controller.go
```diff
@@ -151,111 +154,139 @@ func (r *ReconcileAPIManager) Reconcile(request reconcile.Request) (reconcile.Re
...
+	// We copy the secrets because we don't know the source of them. Might
+	// come from the Cache
+	currentCopy := current.DeepCopy()
+	desiredCopy := desired.DeepCopy()
+
+	// We convert StringData to Data because stringData cannot be read when
+	// obtained from the Kubernetes API and we need to compare the secret
+	// data
+	desiredCopy.Data = secretStringDataToData(desiredCopy.StringData)
+	if secretsEqual(currentCopy, desiredCopy) {
```
eguzki: `updateSecret` method name makes me think you are updating secrets. Why not check if secrets are equal and if not call `updateSecret`?

eguzki:
In general, I looks ok to me. 

On comment though, try to add changes that stick to the PR purpose. For example, try not to add changes that do not have to do with secret reconciliation, like new struct attributes of `ReconcileAPIManager`. It is ok for now.

pkg/controller/apimanager/apimanager_controller.go
miguelsorianod: That's what I'm doing later in https://github.com/3scale/3scale-operator/blob/834e7b6b050731578d08bbda4fd443c954f39168/pkg/controller/apimanager/apimanager_controller.go#L211

miguelsorianod: Maybe renaming the "updateSecret" method to "reconcileSecret" ?

miguelsorianod: I renamed the method updateSecret to reconcileSecret to make more clear that not always an update will be produced

PR #77, opened 2019-03-06
Create ProductVersion 2.5 for the Operator

miguelsorianod:
This PR adds a new ProductVersion for the 3scale API Management 2.5 release in the Operator productVersion field.

The possible update to PostgreSQL 10 will be added in another PR and the change in the templates too

### DIFF
--- a/doc/apimanager-reference.md
+++ b/doc/apimanager-reference.md
@@ -21,7 +21,7 @@ This resource is the resource used to deploy a 3scale API Management solution.
 
 | **Field** | **json/yaml field**| **Type** | **Required** | **Default value** | **Description** |
 | --- | --- | --- | --- | --- | --- |
-| ProductVersion | `productVersion` | ProductVersion | Yes | N/A | 3Scale API Management solution release version. Currently the values "2.4" and "upstream" are supported. **Warning**, "upstream" uses 3scale unstable nightly images and it is intended only for development purposes and to be used in non-productive environments |
+| ProductVersion | `productVersion` | ProductVersion | Yes | N/A | 3Scale API Management solution release version. Currently the values "2.5" and "upstream" are supported. **Warning**, "upstream" uses 3scale unstable nightly images and it is intended only for development purposes and to be used in non-productive environments |
 | WildcardDomain | `wildcardDomain` | string | Yes | N/A | Root domain for the wildcard routes. Eg. example.com will generate 3scale-admin.example.com. |
 | AppLabel | `appLabel` | string | No | `3scale-api-management` | The value of the `app` label that will be applied to the API management solution
 | TenantName | `tenantName` | string | No | `3scale` | Tenant name under the root that Admin UI will be available with -admin suffix.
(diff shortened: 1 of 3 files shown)


pkg/3scale/amp/product/release_2_5.go
```diff
@@ -0,0 +1,47 @@
...
+func (p *release_2_5) GetSystemPostgreSQLImage() string {
+	return "registry.access.redhat.com/rhscl/postgresql-95-rhel7:9.5"
+}
+
+func (p *release_2_5) GetSystemMemcachedImage() string {
+	return "registry.access.redhat.com/3scale-amp20/memcached"
+}
+
+func (p *release_2_5) GetWildcardRouterImage() string {
+	return "registry.access.redhat.com/3scale-amp22/wildcard-router"
```
miguelsorianod: I'm assuming no changes will be produced in WildcardRouter (I've seen there's no recent activity on it) so I guess we'll still keep using the same image than the one that was used in 2.4, which was 3scale-amp22.
Could you confirm that this would be the correct image to be used for the 2.5 release @mikz ?

mikz: Yes, that should be the correct image.

miguelsorianod: :+1: thank you for the confirmation

doc/apimanager-reference.md
```diff
@@ -21,7 +21,7 @@ This resource is the resource used to deploy a 3scale API Management solution.
 
 | **Field** | **json/yaml field**| **Type** | **Required** | **Default value** | **Description** |
 | --- | --- | --- | --- | --- | --- |
-| ProductVersion | `productVersion` | ProductVersion | Yes | N/A | 3Scale API Management solution release version. Currently the values "2.4" and "upstream" are supported. **Warning**, "upstream" uses 3scale unstable nightly images and it is intended only for development purposes and to be used in non-productive environments |
+| ProductVersion | `productVersion` | ProductVersion | Yes | N/A | 3Scale API Management solution release version. Currently the values "2.5" and "upstream" are supported. **Warning**, "upstream" uses 3scale unstable nightly images and it is intended only for development purposes and to be used in non-productive environments |
```
eguzki: 2.4 is not supported? I cannot deploy 2.4 with the operator?

miguelsorianod: I'd say no, because the operator is part of the 2.5 release and some needed APIs at Porta level have been added for the 2.5 release that are needed for the capabilities part.

miguelsorianod: Let me clarify... you could deploy the APIManager in a standalone way with 2.4 with the operator but we cannot guarantee it would not be used with the capabilities part, and operator is not part of the 2.4 release so I don't see the need to include it as an additional version to support.

miguelsorianod: In fact, in another PR I planned to remove the 2_4_release code but I've not done so because it's useful to use to have it for testing purposes until 2.5 is released

eguzki: ok

PR #83, opened 2019-03-11
Upgrade template images to 2.5

miguelsorianod:
This PR upgrades the templates contents to use AMP 2.5.0 preparing it for the upcoming changes.
This assumes correct image URLs, although changes might be needed if was not the case.


miguelsorianod:
Applying this PR makes the productized templates in this repository temporarily not usable until those images are available.
the non-productized templates can still be used and if there's a need to use the productized ones one can go to the https://github.com/3scale/3scale-amp-openshift-templates repository

Issue #96, opened 2019-03-18
Creating a metric with a hyphen in it's name causes the state to be always out of sync


jmprusi:
When creating a metric in the porta API that contains a hyphen, "my-metric" the values in porta for the metric display_name and system_name are different.

for example: 
```
display name: my-metric
system name: my_metric
```

We need to ensure that all the functions for handling metrics use the same fields and handle this "case" correctly.

PR #100, opened 2019-03-18
Use FriendlyName instead of SystemName for metrics

jmprusi:
fixes https://github.com/3scale/3scale-operator/issues/96

### DIFF
--- a/pkg/apis/capabilities/v1alpha1/api_types.go
+++ b/pkg/apis/capabilities/v1alpha1/api_types.go
@@ -234,11 +234,11 @@ func (api API) getInternalAPIfrom3scale(c *portaClient.ThreeScaleClient) (*Inter
 	// Grab the metrics from 3scale.
 	for _, metric := range service.Metrics.Metrics {
 		internalMetric := InternalMetric{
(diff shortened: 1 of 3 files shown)


pkg/apis/capabilities/v1alpha1/api_types.go
```diff
@@ -234,11 +234,11 @@ func (api API) getInternalAPIfrom3scale(c *portaClient.ThreeScaleClient) (*Inter
 	// Grab the metrics from 3scale.
 	for _, metric := range service.Metrics.Metrics {
 		internalMetric := InternalMetric{
-			Name:        metric.MetricName,
+			Name:        metric.FriendlyName,
 			Unit:        metric.Unit,
 			Description: metric.Description,
 		}
-		if internalMetric.Name != "hits" {
+		if internalMetric.Name != "hits" && internalMetric.Name != "Hits" {
```
eguzki: maybe better?
```golang
if strings.ToLower(internalMetric.Name) != "hits"
```

jmprusi: 👍

PR #103, opened 2019-03-19
Dockerfile to build operator

eguzki:
3scale operator image is build using our Dockerfile and Makefile target. Remove operator-sdk dependency for building docker images of the operator.

`/build/Dockerfile` has been left undeleted. This file was generated from operator-sdk scaffolding process and it might need it for some reason.

This PR allows Quay docker repo build images on each update in master branch and on new tags.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,34 @@
+FROM golang:1.11.5 AS build
+
+ENV DEP_VERSION=v0.5.0
+
+RUN curl https://raw.githubusercontent.com/golang/dep/${DEP_VERSION}/install.sh | sh
+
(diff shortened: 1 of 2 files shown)


mikz:
Why not just build it on CircleCI and push it to quay.io? 
We don't want to be reproducing and maintaining what is already provided by the SDK, right?

eguzki:
I am not totally convinced that is the way to go. In https://github.com/3scale/3scale-istio-adapter a dockerfile is used to build images. Provides you full flexibility and it is basically `go build` command to reproduce and maintain.

On the other hand, I will implement using circleCI and come back to this implementation when we require custom building process. :+1: @mikz

PR #114, opened 2019-03-26
Configurable system redis urls

miguelsorianod:
This PR:
 * Adds the SYSTEM_REDIS_URL parameter to all templates and not only to the HA one. Additionally Makes this parameter required without a default value only in the HA template.
 * Adds the SYSTEM_MESSAGE_BUS_REDIS_URL parameter to all templates. Additionally Makes this parameter required without a default value only in the HA template.
 * Sets the MESSAGE_BUS_URL attribute in the 'system-redis' secret gathering its value from the SYSTEM_MESSAGE_BUS_REDIS_URL in the case of templates. In the case of the operator it's also automatically generated if not specified
 * system's redis url is 'redis://system-redis:6379/1' by default (as before)
 * system's message bus redis url is 'redis://system-redis:6379/8' by default (as before)
 *  Adds the environment variable MESSAGE_BUS_REDIS_URL in the system-sidekiq and system-app pod's containers, gathering its value from the MESSAGE_BUS_URL attribute of the 'system-redis' secret
 * In the Operator side, adds reconciliation of the new MESSAGE_BUS_URL attribute of the 'system-redis' secret

The purpose of this changes is allowing a configuration of the system's message bus redis database and system's Redis database into the same logical database, to allow Redis Enterprise to work.

### DIFF
--- a/pkg/3scale/amp/component/system.go
+++ b/pkg/3scale/amp/component/system.go
@@ -408,6 +408,19 @@ func (system *System) buildSystemSphinxEnv() []v1.EnvVar {
 		envVarFromValue("DELTA_INDEX_INTERVAL", "5"),
 		envVarFromValue("FULL_REINDEX_INTERVAL", "60"),
 	)
+	result = append(result, system.SystemRedisEnvVars()...)
+	return result
+}
+
+func (system *System) SystemRedisEnvVars() []v1.EnvVar {
+	result := []v1.EnvVar{}
+
+	result = append(result,
+		envVarFromSecret("REDIS_URL", SystemSecretSystemRedisSecretName, SystemSecretSystemRedisURLFieldName),
+		envVarFromSecret("MESSAGE_BUS_REDIS_URL", SystemSecretSystemRedisSecretName, SystemSecretSystemRedisMessageBusRedisURLFieldName),
+		envVarFromSecret("REDIS_NAMESPACE", SystemSecretSystemRedisSecretName, SystemSecretSystemRedisNamespace),
+		envVarFromSecret("MESSAGE_BUS_REDIS_NAMESPACE", SystemSecretSystemRedisSecretName, SystemSecretSystemRedisMessageBusRedisNamespace),
+	)
 	return result
 }
 
@@ -440,16 +453,12 @@ func (system *System) buildSystemBaseEnv() []v1.EnvVar {
 
 		envVarFromSecret("SECRET_KEY_BASE", SystemSecretSystemAppSecretName, SystemSecretSystemAppSecretKeyBaseFieldName),
 
-		envVarFromSecret("REDIS_URL", SystemSecretSystemRedisSecretName, SystemSecretSystemRedisURLFieldName),
-		envVarFromSecret("MESSAGE_BUS_REDIS_URL", SystemSecretSystemRedisSecretName, SystemSecretSystemRedisMessageBusRedisURLFieldName),
-		envVarFromSecret("REDIS_NAMESPACE", SystemSecretSystemRedisSecretName, SystemSecretSystemRedisNamespace),
-		envVarFromSecret("MESSAGE_BUS_REDIS_NAMESPACE", SystemSecretSystemRedisSecretName, SystemSecretSystemRedisMessageBusRedisNamespace),
-
 		envVarFromSecret("MEMCACHE_SERVERS", SystemSecretSystemMemcachedSecretName, SystemSecretSystemMemcachedServersFieldName),
 
 		envVarFromSecret("BACKEND_REDIS_URL", "backend-redis", "REDIS_STORAGE_URL"),
 	)
 
+	result = append(result, system.SystemRedisEnvVars()...)
 	bckListenerApicastRouteEnv := envVarFromSecret("APICAST_BACKEND_ROOT_ENDPOINT", "backend-listener", "route_endpoint")
 	bckListenerRouteEnv := envVarFromSecret("BACKEND_ROUTE", "backend-listener", "route_endpoint")
 	result = append(result, bckListenerApicastRouteEnv, bckListenerRouteEnv)
--- a/pkg/3scale/amp/component/system.go
+++ b/pkg/3scale/amp/component/system.go
@@ -41,8 +41,11 @@ const (
 )
 
 const (
-	SystemSecretSystemRedisSecretName   = "system-redis"
-	SystemSecretSystemRedisURLFieldName = "URL"
+	SystemSecretSystemRedisSecretName                  = "system-redis"
+	SystemSecretSystemRedisURLFieldName                = "URL"
+	SystemSecretSystemRedisMessageBusRedisURLFieldName = "MESSAGE_BUS_URL"
+	SystemSecretSystemRedisNamespace                   = "NAMESPACE"
+	SystemSecretSystemRedisMessageBusRedisNamespace    = "MESSAGE_BUS_NAMESPACE"
 )
 
 const (
@@ -105,6 +108,9 @@ type systemNonRequiredOptions struct {
 	memcachedServers                       *string
 	eventHooksURL                          *string
 	redisURL                               *string
+	redisNamespace                         *string
+	messageBusRedisNamespace               *string
+	messageBusRedisURL                     *string
 	apicastSystemMasterProxyConfigEndpoint *string
 	apicastSystemMasterBaseURL             *string
 	adminEmail                             *string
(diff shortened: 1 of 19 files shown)


miguelsorianod:
In my opinion this PR performs what's intended in an immediate way but what we perform with it is adding an additional substantial maintenance burden on the templates, and in greater measure into the operator (because potentially might have to handle upgrades and reconciliation if we deemed that necessary, operators potentially perform more tasks than the templates).
Another reason I think that is that once introduce this into GA it makes changes to improve this really hard and almost impossible without breaking compatibility. It makes them hard because once we introduce the ability to configure those DBs via parameters we cannot control what is done with that by the users.

Basically we go from a scenario where:
 * The System's redis URL is not configurable (only in HA scenario) and the System's message bus redis URL is not configurable

Into a scenario where:
 * The System's redis URL is configurable in all  scenarios, and we additionally make the System's message bus redis URL configurable, even in a completely different physical database/instance.
   This adds more maintenance burden already high into the templates and even more into the operator

To me some better solutions would be:
 * Joining the message bus logical database into the same logical database than the redis url one in all scenarios and not only the Redis Enterprise one.
    Making this changes would need careful evaluation and would not be an immediate change because it involves some risks.
   System's team could give more details about this, but one of them I've been told is that key names could collision if not done right.
    Also the fact of making sure that a change of the existing logical DB number does not impact clients (if data is 100% ephemeral and is automatically regenerated shouldn't be a problem).
    This solution would reduce complexity at Templates, Operator and System level including the maintainability of them instead of actually increasing them. In fact, we would have the same DB scenario in Redis and in Redis Enterprise.
 * Only allowing the configuration of the database number. This would require renaming the Redis URL HA template parameter name (breaking compatibility). We would also have to decide whether only allowing changing the Message Bus logical DB name or also introduce the ability to modify the DB number of the system's redis DB. This adds three new parameters but would allow us to only give the option to set those logical DBs in the same physical/instance database, reducing complexity in that regard, although would increase it anyway.

Another solution would be just not changing anything for the moment, and in case anyone wants to configure Redis Enterprise, give them a manual procedure on how to change it (performing the procedure after having deployed the product), and think a good / one of the previously suggested solutions. 

cc @andrewdavidmackenzie too

guicassolato:
@miguelsorianod, maybe we can join the two logical databases and not to cause any damage. I see two ways of accomplishing that:
1. we just point both sidekiq (main user of `REDIS_URL`) and message_bus gem (user of `MESSAGE_BUS_REDIS_URL`) to the same logical db setting in the template for that and done;
2. we ensure (in system code) proper namespaces for each one of these components so keys never collide.

Problem with solution 1 is we don't own neither sidekiq, nor message_bus. At the very least, it would require us some thorough inspection of the code of those libraries to make sure they never collide. Perhaps they are already applying proper namespaces. Either way, we just don't have enough time before 2.5 to ensure that solution 1 works out of the box, and even less time we have to code solution 2 ourselves.

pkg/3scale/amp/component/system_options_builder.go
```diff
@@ -167,6 +179,10 @@ func (s *SystemOptionsBuilder) setNonRequiredOptions() {
 	defaultMemcachedServers := "system-memcache:11211"
 	defaultEventHooksURL := "http://system-master:3000/master/events/import"
 	defaultRedisURL := "redis://system-redis:6379/1"
+	defaultMessageBusRedisURL := "redis://system-redis:6379/8"
```
guicassolato:
```suggestion
	defaultMessageBusRedisURL := "redis://system-redis:6379/2"
```

so it matches the default behavior when `MESSAGE_BUS_REDIS_URL` is not defined.

pkg/3scale/amp/component/system.go
```diff
@@ -259,6 +269,26 @@ func (system *System) buildParameters(template *templatev1.Template) {
...
+			Name:        "SYSTEM_REDIS_URL",
+			Description: "Define the external system-redis to connect to",
+			Required:    true,
+			Value:       "redis://system-redis:6379/1",
+		},
+		templatev1.Parameter{
+			Name:        "SYSTEM_MESSAGE_BUS_REDIS_URL",
+			Description: "Define the external system-redis message bus to connect to",
+			Required:    true,
+			Value:       "redis://system-redis:6379/8",
```
guicassolato:
```suggestion
			Value:       "redis://system-redis:6379/2",
```

pkg/3scale/amp/component/system_options_builder.go
guicassolato: It can be empty by default as well.

pkg/3scale/amp/component/system.go
guicassolato: https://github.com/3scale/3scale-operator/pull/114#pullrequestreview-244229898

pkg/3scale/amp/component/system_options_builder.go
miguelsorianod: Updated to make the parameter not mandatory. The envvar in the system pods thus will be set, and will be set as the empty string

miguelsorianod:
Tested a deploy with the templates with the default parameters (namespaces and messagebus urls empty and the url being /1) and DBs 1 and 2 seem to contain data:

```
127.0.0.1:6379[2]> select 1
OK
127.0.0.1:6379[1]> keys *
 1) "stat:processed"
 2) "system-sidekiq-1-kvw5j:1:cbf44e294750"
 3) "stat:failed:2019-05-31"
 4) "cron_job:hourly_jobs:enqueued"
 5) "cron_job:hourly_jobs"
 6) "cron_job:weekly_jobs"
 7) "cron_job:daily_jobs:enqueued"
 8) "cron_job:weekly_jobs:enqueued"
 9) "cron_job:daily_jobs"
10) "cron_job:monthly_jobs:enqueued"
11) "cron_job:sphinx_index"
12) "cron_job:sphinx_delta"
13) "stat:failed"
14) "processes"
15) "cron_job:sphinx_delta:enqueued"
16) "cron_job:billing_jobs:enqueued"
17) "queues"
18) "BID-98eiDJbZ0jBrzg"
19) "cron_job:sphinx_index:enqueued"
20) "cron_jobs"
21) "cron_job:billing_jobs"
22) "cron_job:monthly_jobs"
23) "stat:processed:2019-05-31"
127.0.0.1:6379[1]> select 2
OK
127.0.0.1:6379[2]> keys *
1) "__mb_backlog_id_n_/__mb_keepalive__/$|$"
2) "__mb_backlog_n_/__mb_keepalive__/$|$"
3) "__mb_global_backlog_n"
4) "__mb_global_id_n"
```

miguelsorianod:
Also tested setting the urls for redis, both in the same db logical number (5) and namespaces set. It seems to work correctly too in templates:

```
127.0.0.1:6379[1]> select 1
OK
127.0.0.1:6379[1]> keys *
(empty list or set)
127.0.0.1:6379[1]> select 2
OK
127.0.0.1:6379[2]> keys *
(empty list or set)
127.0.0.1:6379[2]> select 5
OK
127.0.0.1:6379[5]> keys *
 1) "testnamespace:cron_job:weekly_jobs:enqueued"
 2) "testnamespace:stat:processed:2019-05-31"
 3) "anothernamespace:__mb_global_id_n"
 4) "testnamespace:system-sidekiq-1-xq8wh:1:188d54981e74"
 5) "testnamespace:cron_job:daily_jobs"
 6) "testnamespace:cron_job:sphinx_delta:enqueued"
 7) "testnamespace:cron_job:monthly_jobs:enqueued"
 8) "testnamespace:stat:processed"
 9) "testnamespace:BID-DqgX1_ln7MNTzw"
10) "testnamespace:cron_job:weekly_jobs"
11) "testnamespace:cron_job:billing_jobs:enqueued"
12) "testnamespace:cron_job:sphinx_index:enqueued"
13) "testnamespace:stat:failed:2019-05-31"
14) "testnamespace:processes"
15) "anothernamespace:__mb_backlog_n_/__mb_keepalive__/$|$"
16) "testnamespace:cron_job:sphinx_delta"
17) "testnamespace:cron_job:monthly_jobs"
18) "testnamespace:cron_job:sphinx_index"
19) "testnamespace:retry"
20) "anothernamespace:__mb_backlog_id_n_/__mb_keepalive__/$|$"
21) "testnamespace:cron_job:hourly_jobs:enqueued"
22) "testnamespace:stat:failed"
23) "testnamespace:queues"
24) "testnamespace:cron_jobs"
25) "testnamespace:cron_job:billing_jobs"
26) "anothernamespace:__mb_global_backlog_n"
27) "testnamespace:cron_job:hourly_jobs"
28) "testnamespace:cron_job:daily_jobs:enqueued"
```

PR #117, opened 2019-04-02
[rolling_updates] Add service_mesh_integration: true

hallelujah:
Closes THREESCALE-2165


hallelujah:
@miguelsorianod I think we should not use this anymore.
Porta has defaults settings for on-premises, and it should be used instead of manually changing the templates everytime

https://github.com/3scale/porta/blob/90284d4b19a05b06cb1e945eaa2e925e2a0b30e5/openshift/system/config/rolling_updates.yml

hallelujah:
Thank you @miguelsorianod 😍 
We should probably document how to use the generator so anybody can make a proper PR.

PR #124, opened 2019-05-15
Prepare usage of registry.redhat.io registry

miguelsorianod:
PR to prepare for the use of registry.redhat.io in the images for 2.6 releases and newer ones.

The old RedHat registry access.redhat.com will be deprecated and newer images are not going to be published on it. New images are going to be published in registry.redhat.io. This registry requires authentication.

This PR allows configuring the authentication credentials for the registry.redhat.io via a secret called 'threescale-registry-auth'. This secret is associated to the "amp" ServiceAccount that has been assigned to all the DeploymentConfig elements of the platform.
The final user is the only one that knows the credentials it has so the secret must be created by them with the specific fields needed and values needed before deploying the APIManager/the templates. Otherwise, the deploy will fail.
We can think of making the secret name configurable, but for the moment to already have something we can merge it this way.

In case of making it configurable I would:
In the templates we could:
 * Request the secret name (or have a hardcoded one) to be used via a new required parameter. The user should manually create the secret with the expected fields and values before deploying the templates

In the operator we could:
 * Have a required field in the CR that is the secret/set of secrets to be used by the ServiceAccount(s) that we assign to the DeploymentConfig pods (or have a hardcoded secret name). The user should manually create the secret with the expected fields and values before deploying the 3scale operator CR

### DIFF
--- a/pkg/3scale/amp/component/ampimages.go
+++ b/pkg/3scale/amp/component/ampimages.go
@@ -108,7 +108,7 @@ func (ampImages *AmpImages) buildObjects() []runtime.RawExtension {
 	systemMemcachedImageStream := ampImages.buildSystemMemcachedImageStream()
 	systemMySQLImageStream := ampImages.buildSystemMySQLImageStream()
 
-	quayServiceAccount := ampImages.buildQuayServiceAccount()
+	deploymentsServiceAccount := ampImages.buildDeploymentsServiceAccount()
 
 	objects := []runtime.RawExtension{
 		runtime.RawExtension{Object: backendImageStream},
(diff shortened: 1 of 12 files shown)


pkg/3scale/amp/component/ampimages.go
```diff
@@ -575,7 +575,7 @@ func (ampImages *AmpImages) buildQuayServiceAccount() *v1.ServiceAccount {
 		},
 		ImagePullSecrets: []v1.LocalObjectReference{
 			v1.LocalObjectReference{
-				Name: "quay-auth"}}}
+				Name: "threescale-registry-auth"}}}
```
eguzki: function still called `buildQuayServiceAccount`, kind of weird. isn't it?

miguelsorianod: fixed

PR #126, opened 2019-05-21
Operator sdk v0.8.0

eguzki:
- [x] operator sdk v0.8.0
- [x] Dependency management with Go modules. Dropping `dep` and Gopkg.toml dependency management system
- [x] breaking change: BindingSpec `APISelector` attribute, now `apiSelector`
- [x] Upgrading Go runtime to 1.12.5


pkg/apis/apps/v1alpha1/apimanager_types.go
```diff
@@ -11,43 +11,37 @@ import (
...
+	HighAvailability *HighAvailabilitySpec `json:"highAvailability,omitempty"`
 }
 
 // APIManagerStatus defines the observed state of APIManager
+// +k8s:openapi-gen=true
 type APIManagerStatus struct {
-	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
-	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
-	Conditions []APIManagerCondition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,4,rep,name=conditions"`
+	Conditions []APIManagerCondition `json:"conditions,omitempty" protobuf:"bytes,4,rep,name=conditions"`
```
eguzki: `patchStrategy` and `pathMerge` tags removed. `operator-sdk generate openapi` failed

Still unknown why

pkg/apis/capabilities/v1alpha1/api_types.go
```diff
@@ -39,15 +39,18 @@ type APISelectors struct {
 }
 
 // APIStatus defines the observed state of API
+// +k8s:openapi-gen=true
 type APIStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
+	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
```
eguzki: Should we keep this comment auto generated?? 

https://book.kubebuilder.io/beyond_basics/generating_crd.html

returns 404

miguelsorianod: We can remove it because it does not work anymore. It was useful because in case we wanted to add custom validation there explained how

eguzki:

deploy/crds/apps_v1alpha1_apimanager_crd.yaml
```diff
@@ -60,14 +65,22 @@ spec:
             system:
               properties:
                 database:
+                  description: TODO should union fields be optional?
```
miguelsorianod: For some reason the generator now adds some fields based on some comments that were not added before.
This should not appear on the crd

deploy/crds/apps_v1alpha1_apimanager_crd.yaml
```diff
@@ -60,14 +65,22 @@ spec:
...
                   properties:
                     mysql:
+                      description: Union type. Only one of the fields can be set
                       properties:
                         image:
                           type: string
                       type: object
                   type: object
                 fileStorage:
+                  description: TODO should this field be optional? We have different
```
miguelsorianod: Same about some comments appearing that did not appear before. This was really an internal comment

deploy/crds/apps_v1alpha1_apimanager_crd.yaml
eguzki: those are comments on struct and end up in crd. 

I think we should improve description instead of removing description from CRD's. The more info in the CRD's the better, right?

miguelsorianod: I agree, but isn't there a way to keep comments that won't appear as generated? those are useful to clarify/remember some information that does not necessarily belong to the description

eguzki: According to 
https://github.com/kubernetes/kube-openapi/blob/master/pkg/generators/openapi.go#L483

Adding `---` or `+` or `TODO` should ignore, but does not work. 

We will add extra new line to remove comment

PR #127, opened 2019-05-24
Fix backend init container sentinel env

alanmoran:


pkg/3scale/amp/component/backend.go
```diff
@@ -204,6 +204,8 @@ func (backend *Backend) buildBackendWorkerDeploymentConfig() *appsv1.DeploymentC
 						}, Env: []v1.EnvVar{
 							envVarFromValue("SLEEP_SECONDS", "1"),
 							envVarFromSecret("CONFIG_QUEUES_MASTER_NAME", BackendSecretBackendRedisSecretName, BackendSecretBackendRedisQueuesURLFieldName),
+							envVarFromSecret("CONFIG_QUEUES_SENTINEL_HOSTS", BackendSecretBackendRedisSecretName, BackendSecretBackendRedisQueuesSentinelHostsFieldName),
```
eguzki: It is better to use backend common env to have all env variables available

Something like this:

```
Env: append(backend.buildBackendCommonEnv(), envVarFromValue("SLEEP_SECONDS", "1")),
```

pkg/3scale/amp/component/backend.go
```diff
@@ -287,6 +289,8 @@ func (backend *Backend) buildBackendCronDeploymentConfig() *appsv1.DeploymentCon
...
```
eguzki: Same as above, something like this:
```
Env: append(backend.buildBackendCommonEnv(), envVarFromValue("SLEEP_SECONDS", "1")),
```

pkg/3scale/amp/component/backend.go
alanmoran: Fixed

pkg/3scale/amp/component/backend.go
alanmoran: Fixed

miguelsorianod:
Looks good to me

PR #131, opened 2019-05-29
[WIP] Implement Apicast Custom Resource for Operator

miguelsorianod:
This PR implements the APIcast custom resource to be able to deploy APIcast in a standalone way without the full 3scale infrastructure.

The idea of this PR is that if it is feasible we'll use Kubernetes native objects instead of OpenShift ones. There are two main blockers/research paths that we have to look at apart from the implementation:
 * Image management: Can this be improved? I see that in the templates we are using the image directly instead of ImageStreams in the case of apicast standalone. We don't want to go to ImageStreams because we would like to have apicast standalone as kubernetes native. Right now it is set with ImagePullPolicy to IfNotPresent and the image directly. I think the way it has been working until now might carry some surprises that we have not detected. We should look into that but I guess that we can leave it with the same parameters for the moment because it's how it's been working until now for the templates.
 * Configuration of the apicast management. An approach we thought about this is to have a secret to introduce them, and load the apicast pod envvars directly using the secret with the EnvFrom PodSpec attribute. What we found with this is that even if you change the secret contents the Pod is not aware of them. Maybe as a first step install-only that is enough but it might be problematic in the future. You can always manually redeploy the pod, but it seems that with kubernetes Deploments instead of DeploymentConfigs that is not trivial to do. We'll investigate whether there's a way to manually deploy a pod with Deployments or not.
 * Deployment management: See previous point related to manual redeployment of pods
 * Upgrade management: Decide how upgrades are being to be managed. This is closely related on how this will be done for APIManager and depends on that.

Also, at this moment the APIManager(3scale whole product) will still use its own APIcast definitions different than this new APIcast CR. In the future we might think about whether it would make sense reusing this CR or not, taking into consideration compatibility problems, dependencies, etc...


miguelsorianod:
Closing this issue.

The development continues in the new apicast-operator specific repository https://github.com/3scale/apicast-operator/pull/1

PR #135, opened 2019-05-31
Remove wildcard router from operator and templates

miguelsorianod:
This PR removes the WildcardRouter on the operator and templates side.

Making this change implies:
 * We are no longer using WildcardRouter in the operator
 * With this change, even deploying the productVersion 2.5 with the operator does NOT include the WildcardRouter. We could still deploy it with the wildcardRouter by using the 2.5-stable branch that won't have this modification.
 * The idea for the 2.6 release is to remove the productVersion "2.5" and only allow to use the productVersion value "2.6" (or an equivalent mechanism that only provides the 2.6 images).
 * From 2.6 release onwards we won't have WildcardRouter functionality with the operator, even when manually deployed in OpenShift 3.11. Manual route creation will be needed to add new tenants or we should provide another mechanism to automate it
 * We won't have WildcardRouter in templates either because the new release of Zync includes code to manage the routes to replace it. This means we'll need to provide some migration procedure for it in OpenShift 3.11 when upgrading from 2.5 to 2.6


miguelsorianod:
I tested to deploy the operator without the WildcardRouter and it seems to work (creating the needed additional routes when creating a service and/or a tenant)

PR #140, opened 2019-06-05
Kubernetes objects

eguzki:
Avoid the use of RawExtension and unnecessary conversions of objects, other than for templates

Original PR https://github.com/3scale/3scale-operator/pull/138


eguzki:
Tested this branch:
* Templates are generated as expected
* evaluation and s3 procedures when executed in operator working as expected.
** resource limits removed
** s3 configuration applied and removed RWX pvc's

PR #141, opened 2019-06-05
[build] update zync build

mikz:
zync uses Docker build now (https://github.com/3scale/zync/pull/197)

/cc @3scale/qe


miguelsorianod:
In https://github.com/3scale/3scale-operator/pull/142 PR I implemented a fix to the fact that we did not detect image builds failing from the BuildConfigs. There it can be seen that the deployment fails because it now detects that Zync failds to build the image because it does not have the changes that are going to be introduced in this PR

PR #142, opened 2019-06-06
Check for image build exit status in CircleCI

miguelsorianod:
This PR fixes the image build step in the nightly CircleCI task was being marked as correct when build images were failing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -125,7 +125,18 @@ commands:
             set -x
             oc cancel-build $(oc get bc --output=name)
             for bc in $(oc get bc --output=name); do oc start-build $bc --wait --follow & done
-            wait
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -125,7 +125,18 @@ commands:
             set -x
             oc cancel-build $(oc get bc --output=name)
             for bc in $(oc get bc --output=name); do oc start-build $bc --wait --follow & done
-            wait
+            echo "List of build image processes:"
+            for job in $(jobs -p); do
+              ps aux | grep $job
```
miguelsorianod: I tried to execute:
```
ps -o cmd fp
```
That works locally on my machine but seems that CircleCI uses another shell or shell version so for the moment this would be enough.

PR #143, opened 2019-06-10
OLM requires description and displayName fields to be present

bmozaffa:
OLM requires description and displayName fields to be present

Signed-off-by: Babak Mozaffari <bmozaffa@redhat.com>


bmozaffa:
@eguzki @jmprusi This PR is needed to successfully integrate with OLM, otherwise install plans fail, as these two fields are not present.

PR #147, opened 2019-06-11
Hardcode "s3" value for fileUploadStorage param in case S3 storage is selected.

matskiv:
"s3" is likely the only meaningful value for "FILE_UPLOAD_STORAGE" env var(in case S3 storage is used). Hard-coding it cleans up code and avoids possible confusion for the user.

Please check discussion on JIRA(https://issues.jboss.org/browse/THREESCALE-2774) before merging.
Close this PR if there is a reason for having this value configurable.

Ping @miguelsorianod :)

### DIFF
--- a/deploy/crds/apps_v1alpha1_apimanager_cr_s3.yaml
+++ b/deploy/crds/apps_v1alpha1_apimanager_cr_s3.yaml
@@ -12,4 +12,3 @@ spec:
         awsBucket: <bucket-name>
         awsCredentialsSecret:
           name: <credentials-secret-name>
-        fileUploadStorage: <file-upload-storage-name>
--- a/deploy/crds/apps_v1alpha1_apimanager_crd.yaml
+++ b/deploy/crds/apps_v1alpha1_apimanager_crd.yaml
@@ -88,13 +88,10 @@ spec:
                           type: object
                         awsRegion:
                           type: string
-                        fileUploadStorage:
-                          type: string
                       required:
                       - awsBucket
                       - awsRegion
                       - awsCredentialsSecret
-                      - fileUploadStorage
                       type: object
                     persistentVolumeClaim:
                       description: Union type. Only one of the fields can be set.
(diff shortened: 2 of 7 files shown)


miguelsorianod:
Hi @matskiv,

Thank you for your PR.

When changing the structure of a CR when the operator-sdk is used the regeneration of the CRDs and OpenAPI documentation is regenerated using the tools provided by it. To do it, instead of manually modifying the ```deploy/crds/apps_v1alpha1_apimanager_crd.yaml``` you should execute:
```
operator-sdk generate k8s
operator-sdk generate openapi # You should execute this one with the GO111MODULE environment variable unset
```

And commit the changed files that those command have modified. 
Remember also to be using the operator-sdk version mentioned in the README.md

Apart from that, we are also maintaining a set of OpenShift templates along with the operator. We should remove that from there too. To do that there's a method named "buildParameters" in the 4 pkg/3scale/amp/component/s3.go file where we should remove the definition of the ```FILE_UPLOAD_STORAGE``` OpenShift template parameter too.
Then after doing that we should regenerate the templates by going into:
pkg/3scale/amp and execute:
```
make clean
make -j4 all
```

To regenerate all of them. We should see that only the s3 related templates change by hardcoding the value and removing the template parameter. The changed OpenShift template yaml files should be better to place them in a separate commit to avoid noise with the other commits.

Thank you. If you have any doubt on how to do that please tell me.

miguelsorianod:
Hi @matskiv,

I've implemented the feedback I left you yesterday in my previous comment so don't worry to having to implement it.

Currently when a forked branch is created and then a PR submitted CircleCI tests are not currently working due to some credentials permissions/data. We will fix this as soon as we can but in order to receive it right now I created a new branch and PR maintaining your commit and authorship and  the additional implementations that had to be done and we'll merge that PR instead of this one once we verify the CircleCI tests are correctly executed there. The new PR is: https://github.com/3scale/3scale-operator/pull/150 you can follow the progress there.

I'm going to close this branch.

Thank you very much for your contribution.

PR #149, opened 2019-06-12
[WIP] Change system services ports

miguelsorianod:
As requested by @mikz, the system-{developer,master,provider} OpenShift Services user-facing port have been changed to port 80 in all of them.
The name of the user-facing port in the service has been replaced from "http" to the same name as the Port name exposed in the corresponding containers.

This change will need to be part of the migration procedure too.


miguelsorianod:
Not needed

PR #150, opened 2019-06-12
Remove configurable fileuploadstorage in operator and templates

miguelsorianod:
When in the S3 context (both in templates and operator), the only value that makes sense to have is "s3" as the FILE_UPLOAD_STORAGE environment variable needed by system so we remove the configurability of it.

Part of the PR was already implemented by @matskiv in the https://github.com/3scale/3scale-operator/pull/147 but when a forked branch is created and then a PR submitted CircleCI tests are not currently working due to some credentials permissions/data. We will fix this as soon as we can and apply this solution now.

I merged the commit submitted by @matskiv in the other PR maintaining the original commit and author and I've submitted new commits that contain the parts that were still pending to be implemented.

### DIFF
--- a/pkg/3scale/amp/component/s3.go
+++ b/pkg/3scale/amp/component/s3.go
@@ -242,11 +242,6 @@ func (s3 *S3) PostProcessObjects(objects []runtime.RawExtension) []runtime.RawEx
 func (s3 *S3) buildParameters(template *templatev1.Template) {
 	parameters := []templatev1.Parameter{
 		templatev1.Parameter{
-			Name:        "FILE_UPLOAD_STORAGE",
-			Description: "Define Assets Storage",
-			Required:    false,
-		},
-		templatev1.Parameter{
 			Name:        "AWS_ACCESS_KEY_ID",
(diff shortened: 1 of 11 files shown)


pkg/3scale/amp/component/s3.go
```diff
@@ -197,7 +195,7 @@ func (s3 *S3) addS3PostprocessOptionsToSystemEnvironmentCfgMap(objects []runtime
 		}
 	}
 
-	systemEnvCfgMap.Data["FILE_UPLOAD_STORAGE"] = s3.Options.fileUploadStorage
+	systemEnvCfgMap.Data["FILE_UPLOAD_STORAGE"] = "s3"
```
matskiv: I think this line can be removed.

miguelsorianod: hmm I don't think so because otherwise the element would not be created in the ConfigMap and the envvar could not be defined. Also the default would be local disk which we don't want in that context

matskiv: Oh, I misunderstood the code. I thought this was for template parameter (which got removed). My bad.

PR #153, opened 2019-06-17
Further enhancements to the OLM integration

bmozaffa:
Use of pie chart to show deployment progress
Add wildcard domain value to OLM page

Signed-off-by: Babak Mozaffari <bmozaffa@redhat.com>

### DIFF
--- a/deploy/olm-catalog/3scale-operator/3scale-operator.v0.3.0.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/3scale-operator/3scale-operator.v0.3.0.clusterserviceversion.yaml
@@ -32,6 +32,29 @@ spec:
       version: v1alpha1
       description: API Manager
       displayName: API Manager
+      resources:
+        - kind: DeploymentConfig
+          version: apps.openshift.io/v1
+        - kind: PersistentVolumeClaim
+          version: v1
+        - kind: Service
+          version: v1
+        - kind: Route
+          version: route.openshift.io/v1
+        - kind: ImageStream
+          version: image.openshift.io/v1
+      specDescriptors:
+        - displayName: Wildcard Domain
+          description: Wildcard domain as configured in the API Manager object
+          path: wildcardDomain
+          x-descriptors:
+            - "urn:alm:descriptor:com.tectonic.ui:label"
+      statusDescriptors:
+        - displayName: Deployments
+          description: API Manager Deployment Configs
+          path: deployments
+          x-descriptors:
+            - "urn:alm:descriptor:com.tectonic.ui:podStatuses"
     - kind: API
       name: apis.capabilities.3scale.net
       version: v1alpha1
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ require (
 	contrib.go.opencensus.io/exporter/ocagent v0.4.9 // indirect
 	github.com/3scale/3scale-porta-go-client v0.0.3
 	github.com/Azure/go-autorest v11.5.2+incompatible // indirect
-	github.com/RHsyseng/operator-utils v0.0.0-20190616014937-557e58cee892
+	github.com/RHsyseng/operator-utils v0.0.0-20190617184533-a633dc5f85a3
 	github.com/appscode/jsonpatch v0.0.0-20190108182946-7c0e3b262f30 // indirect
 	github.com/coreos/prometheus-operator v0.26.0 // indirect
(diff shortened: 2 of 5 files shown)


bmozaffa:
@eguzki see screenshot. I also added the wildcard domain to this page. Let me know if you think that's appropriate, or if there are other attributes or values directly or indirectly reflected in the APIManager CR that the user should see on the front page.

Screen Shot 2019-06-16 at 6 53 05 PM

eguzki:
This is cool :) 

`wildcardDomain` is for sure very helpful. Other interesting deployment properties I can think about are:
* `spec.resourceRequirementsEnabled` attribute (resource limits set)
* Main components images:
  * `spec.apicast.image`
  * `spec.backend.image`
  * `spec.system.image`
  * `spec.zync.image`
* `spec.highAvailability` attribute
* `spec.system.fileStorage` object to see if S3 has been configured or not.
* Mysql or Postgress has been selected for the deployment
  * `spec.system.database.mysql` or `spec.system.database.postgresql`

Among all of them. `resourceRequirements` and `highAvailability` are good candidates to be included. I think we can merge this PR as it is now and we can add more having this PR as template.

wdyt @miguelsorianod ?

miguelsorianod:
Hi Babak,

Thank you for the PR. It seems very useful.

However, I'm concerned about the stability of applying the Watch on DeploymentConfigs and the addition of the update status.
At this moment I'm not sure whether the code is prepared to support that. I'm concerned about consistency problems when several events are generated at the same reconciliation loop occurrence.

Right now only one reconciliation event was generated for every loop execution and now if I'm understanding this correctly, for every reconciliation loop occurrence, potentially as much events as the number of DeploymentConfigs that the APIManager managed + the update status event + the APIManager CR itself updates can potentially occur.

When I have some time I'll try to generate some situations to see whether it works as expected. As I said I'm concerned about the capability of our current code to manage that correctly, avoiding inconsistencies (what if the operator deployment has more than one instance and handles events simultaneously?), lost changes, infinite loops etc...

bmozaffa:
Hi Miguel,

I understand your concern, but will say that you have to support multiple and ongoing reconciliation calls anyway. That is how the operator SDK is designed to work and it is called a reconciliation loop. You will get occasional reconcile calls for various reasons, even if you don't watch deployment configs.

Regards, Babak

miguelsorianod:
Hi Babak,

I know, but I'm not sure at this moment the code are prepared to add the behaviour that performs what you say without causing instabilities/undesired effects in the reconciliation.

We are able to receive multiple events from CR changes itself and we only generate one event per reconciliation call at this moment. I'm not sure whether the code is able to manage multiple triggered events per reconciliation call and/or generated from other resources at this moment. I'll take a look at it as soon as possible.
I completely agree that we should support it, we are going that way because for upgrades/reconciliation of some fields that will be needed and the fact of being able to support multiple and ongoing reconciliation calls to is an existing planned task

eguzki:
Paradoxically we have implemented "capabilities" feature, which is in TP, of  the operator following reconciliation pattern and, on the other hand, "Install" feature, which is in GA, does not implement reconcilliation logic (except for secrets).

Definitely this is something we have to fix ASAP.

I will try to test this PR and check closely the effect of multiple reconciliation calls.

bmozaffa:
Hi @eguzki and @miguelsorianod ,

FWIW, I did of course deploy the operator to our OCP lab and test the effects of this PR, and everything worked fine.

The code in this PR watches the deployment as a secondary resource owned by the CR. That means every reconcile call is still tied to the CR, and the request object looks the same as it looked before. You will not be getting a different "input".

I also reviewed the API Manager reconcile call, and see no issue at all, where reconciliation is concerned. I do, however, have some unrelated comments, which you can find here: https://github.com/bmozaffa/3scale-operator/pull/1#pullrequestreview-252043825

Regards, Babak

miguelsorianod:
Hi Babak,

Thank you very much for reviewing some parts of the reconciliation code. It is very helpful to have feedback from other people.

Yesterday I reviewed the code too, seeing how events were and performing some tests and it indeed seems to be working correctly and behaving as expected with the new behaviour. Lots of more events than before are generated than before but that's completely expected as it keeps tracks of DeploymentConfig changes now, and in the near future we'll probably add lots of more watches for all the elements that the APIManager creates/will reconciliate.

I'll merge this PR. I also have looked at the unrelated comments you left and will look to perform some changes in another PR

PR #159, opened 2019-06-20
2.6 image changes

miguelsorianod:


miguelsorianod:
cc @3scale/qe , From the 2.6 release, registry.redhat.io is used for the images. In order for the smoke tests and other tests that you might have you need to create an OpenShift docker secret named "threescale-registry-auth" with registry.redhat.io credentials before deploying 3scale as commented here https://issues.jboss.org/browse/THREESCALE-2292. Otherwise, it will fail to pull the images.

PR #173, opened 2019-07-02
refactor template factories

eguzki:
Remove list of templates from factory. Each template factory register itself to be available to the CLI.


eguzki:
`options` not used at all

Maybe they should be removed?

miguelsorianod:
I think we can remove them.

If we were to add them probably would not be in their current form (array of strings)

eguzki:
options removed

PR #186, opened 2019-07-09
Add route customhost update

miguelsorianod:
This commits adds the route/custom-host update verb into the zync-que-role that is needed by the Zync component that has been requested by Zync team.

This change has a big impact where you need a privileged user to deploy 3scale from this point on.
This change should be temporary until we find a different way in Zync to do the desired tasks, but we add it so we can proceed with our internal testings for the moment.

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -66,6 +66,7 @@ rules:
   - routes/custom-host
   verbs:
   - create
+  - update
 - apiGroups:
   - route.openshift.io
   resources:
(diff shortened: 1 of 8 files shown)


miguelsorianod:
I'm closing this PR and open a new one where we only merge it into the 2.6-stable-prod branch (the only one used for the productized version of templates/operator)

mikz:
You can update the base on GitHub.

miguelsorianod:
Yes it's the first thing I did but the branches differ in lots of commits (more than 100 files changed)

mikz:
You can rebase & force push.

PR #190, opened 2019-07-09
2.6 stable updateapicast template26

miguelsorianod:
This PR:
 * Updates apicast template image to use the existing quay community image
 * Removes unused quay-auth imagepullsecret(now it gathers the image from a public quay registry)


miguelsorianod:
I know I've written "ER1" for the quay image.
In this PR lets just use the same image for apicast that is being currently used for apicast in amp.yml

In a future PR we need to update all images (including in the apicast.yml template) to use the new quay images once we have them

PR #203, opened 2019-08-06
Operator APIManager reconcile/upgrade

miguelsorianod:


eguzki:
Commit history should be cleaned @miguelsorianod 

After, it is ready for review

PR #204, opened 2019-08-12
Specific CR field reconciliation pattern implementation approach

eguzki:
New reconciliation approach based on k8s native object fields controlled directly by CR fields.
- [x] Backend reconciliation
- [x] ampimages reconciliation
- [x] AMP Service account reconciliation
- [x] apicast reconciliation
- [x] Redis backend reconciliation
- [x] Redis system reconciliation
- [x] Postgresql system reconciliation
- [x] MySql system reconciliation
- [x] Memcached reconciliation
- [x] Zync reconciliation
- [x] System reconciliation


eguzki:
@miguelsorianod ready for review

pkg/3scale/amp/operator/ampimages_reconciler.go
```diff
@@ -75,48 +77,72 @@ func (r *AMPImagesReconciler) ampImages() (*component.AmpImages, error) {
...
 	if err != nil {
 		return err
 	}
-
-	return r.imagestreamReconciler.Reconcile(desiredImageStream)
+	r.Logger().Info(fmt.Sprintf("%s reconciled", ObjectInfo(desiredImageStream)))
+	return nil
 }
 
-func (r *AMPImagesReconciler) reconcileServiceAccount(desiredServiceAccount *v1.ServiceAccount) error {
```
miguelsorianod: Typo "reconcileServiceAccount"

pkg/3scale/amp/operator/ampimages_reconciler.go
```diff
@@ -75,48 +77,72 @@ func (r *AMPImagesReconciler) ampImages() (*component.AmpImages, error) {
...
+	if err != nil {
+		return err
+	}
+	r.Logger().Info(fmt.Sprintf("%s reconciled", ObjectInfo(desiredImageStream)))
+	return nil
 }
 
 func (r *AMPImagesReconciler) reconcileDeploymentsServiceAccount(desiredServiceAccount *v1.ServiceAccount) error {
-	return r.reconcileServiceAccount(desiredServiceAccount)
+	reconciler := NewServiceAccountBaseReconciler(r.BaseAPIManagerLogicReconciler, NewCreateOnlyServiceAccoountReconciler())
```
miguelsorianod: Typo "NewCreateOnlyServiceAccoountReconciler"  -> Accoount -> Account

pkg/3scale/amp/operator/imagestream_reconciler.go
```diff
@@ -4,137 +4,105 @@ import (
...
-		if updatedTagRef, ok := updatedImageStreamTagReferenceMap[desiredTagRef.Name]; ok {
-			desiredTagRef.Generation = updatedTagRef.Generation
+func NewImageStreamGenericReconciler() *ImageStreamGenericReconciler {
+	return &ImageStreamGenericReconciler{}
+}
 
-			if desiredTagRef.ReferencePolicy.Type == "" {
-				desiredTagRef.ReferencePolicy.Type = updatedTagRef.ReferencePolicy.Type
+func (r *ImageStreamGenericReconciler) IsUpdateNeeded(desired, existing *imagev1.ImageStream) bool {
+	// merging approach will be implemented
```
miguelsorianod: Is there a specific motivation for this in front on just defining what's in the "desired" ImageStream?

pkg/3scale/amp/operator/apicast_reconciler.go
```diff
@@ -63,48 +134,52 @@ func (r *ApicastReconciler) apicast() (*component.Apicast, error) {
...
-
-	return r.deploymentConfigReconciler.Reconcile(desiredDeploymentConfig)
+	r.Logger().Info(fmt.Sprintf("%s reconciled", ObjectInfo(desiredDeploymentConfig)))
+	return nil
 }
 
-func (r *ApicastReconciler) reconcileConfigMap(desiredConfigMap *v1.ConfigMap) error {
-	err := r.InitializeAsAPIManagerObject(desiredConfigMap)
+func (r *ApicastReconciler) reconcileProductionDeploymentConfig(desiredDeploymentConfig *appsv1.DeploymentConfig) error {
+	reconciler := NewDeploymentConfigBaseReconciler(r.BaseAPIManagerLogicReconciler, NewApicastStagingDCReconciler(r.BaseAPIManagerLogicReconciler))
```
miguelsorianod: Is NewApicastStagingDCReconciler correct in the case of the reconcileProductionDeploymentConfig ? Shouldn't it be NewApicastProductionDCReconciler ?
In case they share the same behaviour we could name it NewApicastDCReconciler. In case you don't like that because they might differ in the future I'd create another type and make both call the same function and if they differ in the future act accordingly

miguelsorianod:
I think we should reconcile the APIManager OwnerReference object in case it is not set for all objects that are managed by the APIManager.
For example, if the user precreates a configmap/secret because he wants to set some values different than the default ones the operator should add an owner reference to it.

pkg/3scale/amp/operator/deployment_config_reconciler.go
```diff
@@ -4,228 +4,108 @@ import (
...
-		changed = true
-	}
-
-	// Reconcile PodTemplateSpec labels
-	if r.ObjectMetaMerger.EnsureLabels(updated.Spec.Template.ObjectMeta.Labels, desired.Spec.Template.ObjectMeta.Labels) {
-		changed = true
+	//
+	// Check container resource requirements
+	//
+	if len(desired.Spec.Template.Spec.Containers) != 1 {
```
miguelsorianod: Is this check really needed?
It should never happen. If it's not 1 it means that the source code itself of the definition of the DeploymentConfig is wrong.
Containers is not the number of "replicas" defined but how many container definitions there are in the DeploymentConfig

pkg/3scale/amp/operator/imagestream_reconciler.go
eguzki: IMO, both approaches are valid for imagestreams (not for all k8s resources). Merging approach would leave existing tags while updating approach would delete existing tags not desired.

If there is no clear reason to implement "updating" approach, I would leave it as it is now, i.e. merging approach

pkg/3scale/amp/operator/apicast_reconciler.go
eguzki: I will change NewApicastDCReconciler and when there are differences, make it explicit.

pkg/3scale/amp/operator/deployment_config_reconciler.go
```diff
@@ -4,228 +4,108 @@ import (
...
+		existing.Spec.Template.Spec.Containers = desired.Spec.Template.Spec.Containers
+		update = true
 	}
 
-	r.ensurePodContainers(updated.Spec.Template.Spec.InitContainers, desired.Spec.Template.Spec.InitContainers)
-	if !reflect.DeepEqual(updated.Spec.Template.Spec.InitContainers, desired.Spec.Template.Spec.InitContainers) {
-		fmt.Println(cmp.Diff(updated.Spec.Template.Spec.InitContainers, desired.Spec.Template.Spec.InitContainers, cmpopts.IgnoreUnexported(resource.Quantity{})))
-		updated.Spec.Template.Spec.InitContainers = desired.Spec.Template.Spec.InitContainers
-		changed = true
+	if !reflect.DeepEqual(existing.Spec.Template.Spec.Containers[0].Resources, desired.Spec.Template.Spec.Containers[0].Resources) {
```
miguelsorianod: This comparison causes false positives due to reflect ends comparing unexported resources and the way resourceRequirements are internally stored sometimes differs the way that we set it.
To solve that problem in one of my original commits I implemented a solution for that that you can reuse to fix this false positives. The method where I implemented was called "ensureResourceRequirements"

pkg/3scale/amp/operator/deployment_config_reconciler.go
```diff
@@ -4,228 +4,108 @@ import (
...
-	for idx := range updated {
-		container := &updated[idx]
-		updatedContainerMap[container.Name] = container
-	}
-
-	for idx := range desired {
-		desiredContainer := &desired[idx]
-		if updatedContainer, ok := updatedContainerMap[desiredContainer.Name]; ok {
-			desiredContainer.Image = updatedContainer.Image
+func DeploymentConfigReconcileReplicas(desired, existing *appsv1.DeploymentConfig, logger logr.Logger) bool {
```
miguelsorianod: About the reconciliation of replicas, we have to take into account that currently:
· When HA option in the APIManager CR is NOT set, the desired replicas are set to 1
· When HA option in the APIManager CR is set, the desired replicas are set to 2, with the exception of DBs that cannot have multiple instances at the same time due to they are not compatible, where we keep it to 1.

If we reconcile replicas with the exact specified number in "desired", the user will not be able to change the number of replicas as of now. I think the solution is not good enough at this moment and we should discuss what are we going to do with this aspect.

pkg/3scale/amp/operator/deployment_config_reconciler.go
eguzki: this is an `assert`. Makes sure about assumptions made, and panics with clear error message instead some `index out of range` confusing error.

pkg/3scale/amp/operator/deployment_config_reconciler.go
eguzki: :+1:

pkg/3scale/amp/operator/deployment_config_reconciler.go
eguzki: HAck implemented to enable manually scaling of replicas.

eguzki:
All changes requested applied, ready for a new review @miguelsorianod

pkg/3scale/amp/operator/imagestream_reconciler.go
miguelsorianod: :+1:

pkg/3scale/amp/operator/apicast_reconciler.go
miguelsorianod: :+1:

pkg/3scale/amp/operator/deployment_config_reconciler.go
```diff
@@ -3,229 +3,111 @@ package operator
...
-		desired.RecreateParams = updated.RecreateParams
-	}
+	// TODO implement right replica reconciliation with replica fields in CRD
 
-	if desired.Type == appsv1.DeploymentStrategyTypeCustom && desired.CustomParams == nil {
-		desired.CustomParams = updated.CustomParams
+	// Currently, to alllow scaling manually, we do no reconcile repolicas.
+	// One exception: Just in case some configuration (like HA conf) sets replica to higher than 1 and existing replica is 1,
+	// then the replica will be reconciled. This is a hack to be removed when replica reconciliation is implemented.
+	if desired.Spec.Replicas > 1 && existing.Spec.Replicas == 1 {
```
miguelsorianod: :+1:

pkg/3scale/amp/operator/secret_reconciler.go
```diff
@@ -3,88 +3,79 @@ package operator
...
 
-// TODO should this be a shared Secret reconcile behaviour or could there be
-// different secrets where we reconcile them in a different way?
-// For example, secrets where we just want to check if they are installed it
-// but we don't want to reconcile them. That would make sense for example
-// for system-seed secret.
-// Also in the case where we wanted to reconcile them but we wanted to do it
-// in a different way
-func (r *SecretReconciler) Reconcile(desiredSecret *v1.Secret) error {
+func (r *SecretBaseReconciler) Reconcile(desiredSecret *v1.Secret) error {
```
miguelsorianod: With the changes we have lost the reconciliation of Secret defaults to allow the users to set specific desired values  for them before deploying APIManager. We should allow that.
Notice that I'm not saying reconciling secrets after APIManager has been deployed, which at this moment doesn't work and didn't work either automatically out of the box due to DeploymentConfigs not reacting to that.

pkg/3scale/amp/operator/configmap_reconciler.go
```diff
@@ -3,77 +3,76 @@ package operator
...
+		BaseAPIManagerLogicReconciler: baseAPIManagerLogicReconciler,
+		reconciler:                    reconciler,
 	}
 }
 
-func (r *ConfigMapReconciler) Reconcile(desiredConfigMap *v1.ConfigMap) error {
-	objectInfo := ObjectInfo(desiredConfigMap)
-	existingConfigMap := &v1.ConfigMap{}
-	err := r.Client().Get(context.TODO(), types.NamespacedName{Name: desiredConfigMap.Name, Namespace: desiredConfigMap.Namespace}, existingConfigMap)
+func (r *ConfigMapBaseReconciler) Reconcile(desiredCM *v1.ConfigMap) error {
```
miguelsorianod: I understand the followed criteria to reconcile ConfigMaps/Secrets has been to reconcile the fields of the ones that are indirectly set through a CR field.
Not for this PR, but we should talk about what are we going to do with the reconciliation of ConfigMaps/Secrets.

miguelsorianod:
Left some more comments.

Please see my https://github.com/3scale/3scale-operator/pull/204#issuecomment-527371456 previous comment too.

miguelsorianod:
I don't see AppLabel CR field being reconciled on the objects. I guess we should reconcile it too?

eguzki:
* Labels and Annotations reconciled for all resource types
* Created `DefaultsOnlySecretReconciler`: Only reconciles when field is missing. 
* Ownerrefrence set for pre-created secrets and configmaps 

miguelsorianod:
Looks good :+1:

PR #205, opened 2019-08-13
Detect and apply upgrade POC

miguelsorianod:


pkg/controller/apimanager/apimanager_controller.go
```diff
@@ -100,11 +104,50 @@ type ReconcileAPIManager struct {
 	operator.BaseControllerReconciler
 }
 
+const (
+	ThreescaleVersionAnnotation = "apps.3scale.net/apimanager-threescale-version"
+	OperatorVersionAnnotation   = "apps.3scale.net/threescale-operator-version"
+)
+
+func (r *ReconcileAPIManager) updateVersionAnnotations(cr *appsv1alpha1.APIManager) error {
+	if cr.Annotations == nil {
```
eguzki: Annotations can be nil?

pkg/controller/apimanager/apimanager_controller.go
```diff
@@ -100,11 +104,50 @@ type ReconcileAPIManager struct {
...
+	cr.Annotations[ThreescaleVersionAnnotation] = product.ThreescaleRelease
+	cr.Annotations[OperatorVersionAnnotation] = version.Version
+	err := r.Client().Update(context.TODO(), cr)
+	return err
+}
+
+func (r *ReconcileAPIManager) upgradeAPIManager(cr *appsv1alpha1.APIManager) (reconcile.Result, error) {
+	// The object to instantiate would change in every release of the operator
+	// that upgrades the threescale version
+	upgrade := Upgrade25_to_26{
```
eguzki: upgrade25_to_26 does not make sense

pkg/controller/apimanager/upgrade_25_to_26.go
```diff
@@ -0,0 +1,57 @@
...
+		Value: "anewvalue",
+	}
+	backendListenerEnvVars := existingDeploymentConfig.Spec.Template.Spec.Containers[0].Env
+	found := false
+	changed := false
+	for idx, val := range backendListenerEnvVars {
+		if val.Name == newEnvVar.Name {
+			found = true
+			if val.Value != newEnvVar.Value {
+				backendListenerEnvVars[idx].Name = newEnvVar.Value
```
eguzki: :eyes:

pkg/controller/apimanager/apimanager_controller.go
miguelsorianod: Yes it seems so.

When you don't have annotations you have the zero value of them. Annotations is a map[string]string. The zero-value of a map in Go is nil.

pkg/controller/apimanager/apimanager_controller.go
miguelsorianod: :+1: yes at the end the struct name will be another one for sure.
This was just only an example to test the mechanism when 2.6 was not released yet.

pkg/controller/apimanager/apimanager_controller.go
eguzki: :+1:

eguzki:
Instead of adding some env var to a deployment config, the PoC would be more useful implementing upgrading 3scale images. We do not have urls, yet, but can be mocked, and surely it will be required when upgrading.

miguelsorianod:
I added the code that upgrades the images assuming it contains a different set of images than the deployed ones. Take into account that the code does not have the decision on whether we should upgrade them or not depending on the CR has the image fields or not, which is being discussed.

I also added code that we might need in the future of removing AMPRelease from some places. This is not 100% confirmed yet but it's another good case to review that we can see how it would be. We can remove it later from the PR if desired too.

pkg/controller/apimanager/upgrade_25_to_26.go
miguelsorianod: Fixed :+1:

pkg/controller/apimanager/upgrade_26_to_27.go
```diff
@@ -0,0 +1,383 @@
...
+	// 	return false, nil
+	// }
+
+	apicastImageStream := &imagev1.ImageStream{}
+	err := u.client.Get(context.TODO(), types.NamespacedName{Name: "amp-apicast", Namespace: u.cr.Namespace}, apicastImageStream)
+	if err != nil {
+		return false, err
+	}
+
+	newImageURL := product.CurrentImageProvider().GetApicastImage()
```
eguzki: Yeah , this is the biggest issue: new image URL is hardcoded or they can be read from CR?

I think that we should read from CR and desired state enforces some images, let it be.

If hardcoded, upgrade will update images to hardcoded ones, but inmediate reconciliacion iteration will update then to the desired state.

We should make very clear that upgrading procedure could potentially fail if images are overridden in CR. 

WDYT??

pkg/controller/apimanager/upgrade_25_to_26.go
```diff
@@ -0,0 +1,57 @@
+package apimanager
```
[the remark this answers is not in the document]
miguelsorianod: Removed

pkg/controller/apimanager/upgrade_26_to_27.go
```diff
@@ -0,0 +1,383 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type Upgrade26_to_27 struct {
+	BaseUpgrade
+}
+
+func (u *Upgrade26_to_27) Upgrade() (reconcile.Result, error) {
+	res, err := u.removeAMPReleaseReferences()
```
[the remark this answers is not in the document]
miguelsorianod: I've reorganized PR commits isolating the changes for this so we can remove the commit before merging the PR

pkg/controller/apimanager/upgrade_26_to_27.go
```diff
@@ -0,0 +1,383 @@
...
+
+func (u *Upgrade26_to_27) removeAMPReleaseFromSystemEnvironmentConfigMap() (reconcile.Result, error) {
+	existingSecret := &v1.Secret{}
+	err := u.client.Get(context.TODO(), types.NamespacedName{Name: "system-environment", Namespace: u.cr.Namespace}, existingSecret)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	keyToRemove := "AMP_RELEASE"
+	for key := range existingSecret.StringData {
```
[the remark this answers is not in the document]
miguelsorianod: :+1: fixed

pkg/controller/apimanager/upgrade_26_to_27.go
```diff
@@ -0,0 +1,383 @@
...
+	err := u.client.Get(context.TODO(), types.NamespacedName{Name: "system-sidekiq", Namespace: u.cr.Namespace}, existingDeploymentConfig)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	envVarToRemove := "AMP_RELEASE"
+	changed := false
+	preHookPod := existingDeploymentConfig.Spec.Strategy.RollingParams.Pre.ExecNewPod
+	preHookPodEnv := preHookPod.Env
+	for idx, val := range preHookPodEnv {
```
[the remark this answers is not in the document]
miguelsorianod: I don't really want to filter an element but search for it and remove that element.
Filtering keeps appending the elements that "pass the filter" whereas in this case I'm just removing an element from an existing array based on a condition https://github.com/golang/go/wiki/SliceTricks#delete.

I prefer to leave it this way as is a valid way to do it too.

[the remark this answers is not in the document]
eguzki: `I don't really want to filter an element but search for it and remove that element.` -> :thinking: 
The method implemented is pretty inefficient; it's doing a lot of extra work, due to the way its using append. 
Proposed solution also has the benefit of being simpler and shorter.

But, in the end, up to you!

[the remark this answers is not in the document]
miguelsorianod: What I meant is that filter runs the entire array looking for the element, and keeps going even when the element has been found to keep adding all the other elements), whereas the deletion just searches for the occurrence of the element and deletes it at that position. It also seems to need garbage-collection code after doing that for what I see in the provided link.

About the efficiency, unless we detect a performance problem due to that I don't see this as much of a problem so I'd avoid premature optimization unless there's a justified reason for it.

I like that it clearly expresses that once you find the element, you delete it and that's it.

pkg/controller/apimanager/upgrade_26_to_27.go
miguelsorianod: I agree with what you say.
I've updated the PR and now in an image has been overriden at the CR level, during the ImageStreams upgrade procedure, it does not apply the corresponding imagestream upgrade logging a warning about it and continuing the upgrade process.
I agree that we should make very clear that the upgrade procedure could fail if images are overriden in the CR and that the users should never set them unless they are in some devtest scenario and knowing that they won't be able to correctly upgrade in that case.

As we talked before, I'm a bit concerned about how we will manage the upgrade procedure testing when doing QE as the procedure until now has always required overriding the images and this would make the upgrade not testable, but lets do it this way for the moment as our intention is that users should not override when upgrading, and lets keep talking with QE on how are we going to manage that specific scenario.

miguelsorianod:
I've also updated the PR with a bugfix that fixes a bug that I detected detecting ImageStream URL and tag changes and some logging additions.

miguelsorianod:
I've removed the upgrade procedure part that took care of removing AMP_RELEASE as we are not sure that we will need this for the moment

miguelsorianod:
I refactored a little bit the upgradeImageStream method and I've added additional logic to reconcile the display-name annotation with the new tag name

eguzki:
> I refactored a little bit the upgradeImageStream method and I've added additional logic to reconcile the display-name annotation with the new tag name

IMO you are re-implementing imagestream reconciliation step, so reconciliation logic is duplicated in a great extent. Tag names, references and even annotations are reconciled, exactly what reconciliation step does. Duplicated.

What I would do is to make an instance of Reconciler type and call reconcile method given desired resource.

eguzki:
Upgrading imagestreams using reconciliation logic to avoid duplication will be implemented later.

PR #212, opened 2019-08-22
Community operator updates

eguzki:
Operator package name must be unique across the entire marketplace

https://docs.google.com/document/d/1ohqX7DJ39fuLIghSsFprgddthw_alz9UDGDrb-b_A84/edit

Changing community released operator package name to *3scale-community-operator*


deploy/olm-catalog/3scale-community-operator/3scale-community-operator.v0.3.0.clusterserviceversion.yaml
```diff
@@ -107,29 +107,29 @@ spec:
...
 
     ### License
-    3scale Operator is licensed under the [Apache 2.0 license](https://github.com/3scale/3scale-operator/blob/master/LICENSE)
+    3scale community Operator is licensed under the [Apache 2.0 license](https://github.com/3scale/3scale-operator/blob/master/LICENSE)
 
   keywords:
     - 3scale
     - API
-  displayName: 3scale Operator
+  displayName: 3scale Community Operator
```
gsaslis: If this name is what is displayed in the operator catalogue (e.g. operatorHub.io, or whatever) I think it would make sense to entirely omit "Operator" here. 

It's clear to me that I'm looking at a list of operators anyway, so I don't think the "Operator" in the name is much more than noise. 

(Just like in docker hub you wouldn't see "postgres docker image" but just "postgres" when looking at the list of container images)

eguzki: Agree, and Community will be removed as well because it is explicit from provider type. RH's Operatorhub displays `community` watermark when provider type is community

deploy/olm-catalog/3scale-community-operator/3scale-community-operator.v0.3.0.clusterserviceversion.yaml
```diff
@@ -107,29 +107,29 @@ spec:
...
 
     ### License
-    3scale Operator is licensed under the [Apache 2.0 license](https://github.com/3scale/3scale-operator/blob/master/LICENSE)
+    3scale community Operator is licensed under the [Apache 2.0 license](https://github.com/3scale/3scale-operator/blob/master/LICENSE)
 
   keywords:
     - 3scale
     - API
-  displayName: 3scale Operator
+  displayName: 3scale
```
miguelsorianod: 3scale community here?

deploy/olm-catalog/3scale-community-operator/3scale-community-operator.v0.3.0.clusterserviceversion.yaml
```diff
@@ -257,7 +257,7 @@ spec:
           - tenants
           verbs:
           - '*'
-        serviceAccountName: 3scale-operator
+        serviceAccountName: 3scale-community-operator
```
miguelsorianod: I think this should be "3scale-opertor"
This is the serviceaccount of the 3scale-operator, which has not changed

deploy/olm-catalog/3scale-community-operator/3scale-community-operator.v0.3.0.clusterserviceversion.yaml
eguzki: See discussion [here](https://github.com/3scale/3scale-operator/pull/212#discussion_r317090653)

What do you think? I think leaving 3scale only makes more sense.

deploy/olm-catalog/3scale-community-operator/3scale-community-operator.v0.3.0.clusterserviceversion.yaml
eguzki: productized operator and community operator already share CRD's. ServiceAccount also should be shared? Why not different one, so they could coexist more independently?? Deployment name has also changed.

deploy/olm-catalog/3scale-community-operator/3scale-community-operator.v0.3.0.clusterserviceversion.yaml
thomasmaas: I’d say it should also say community here so you know which one you’re using. Why leaving 3scale only makes more sense?

eguzki: from previous [discussion](https://github.com/3scale/3scale-operator/pull/212#pullrequestreview-278945408):

`Community will be removed as well because it is explicit from provider type. RH's Operatorhub displays community watermark when provider type is community`

deploy/olm-catalog/3scale-community-operator/3scale-community-operator.v0.3.0.clusterserviceversion.yaml
miguelsorianod: Both operators won't be able to coexist in the same project anyway (as the APIManager resources are always the same), so if you thought it in that way I don't see any advantage of doing that. The same applies with the Deployment name.
If the reason is "just to make it more explicit at the resource level names" I'm not strongly against it. The consequence is we'll have to maintain those differences between branches.

eguzki: will change that then, less differences, the better

deploy/olm-catalog/3scale-community-operator/3scale-community-operator.v0.3.0.clusterserviceversion.yaml
```diff
@@ -107,25 +107,25 @@ spec:
...
-    If you encounter any issues while using 3scale operator, you can create an issue on our [website](https://github.com/3scale/3scale-operator) for bugs, enhancements, or other requests.
+    If you encounter any issues while using 3scale community operator, you can create an issue on our [website](https://github.com/3scale/3scale-operator) for bugs, enhancements, or other requests.
 
     ### Contributing
     You can contribute by:
 
-    * Raising any issues you find using 3scale Operator
+    * Raising any issues you find using 3scale community Operator
     * Fixing issues by opening [Pull Requests](https://github.com/3scale/3scale-operator/pulls)
     * Improving [documentation](https://github.com/3scale/3scale-operator/blob/v0.3.0/doc/user-guide.md)
```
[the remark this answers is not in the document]
eguzki: This PR is intended to have 3scale community operator ready to be released. The tagged version will be `v0.3.0`, and the link will become available eventually.

Documentation in redhat.com is meant for productized version. Community release documentation is GH itself, we do not have yet specific doc site for that.

PR #214, opened 2019-08-29
fix zync option reading from operator

eguzki:
One time is enough when reading options from secret


eguzki:

miguelsorianod:
Approved, but we should force a new push so CI is reexecuted again, as currently is not working due to a dependency is not available from its server

miguelsorianod:
Rebase from master please

PR #232, opened 2019-09-17
upgrade procedure respects CR image fields

eguzki:

PR #234, opened 2019-09-20
fix upgrade imagestreams

eguzki:
There was an issue testing upgrading procedure. We are upgrading using operator version but upgrading imagestreams is made using 3scale version. That issue caused endless loop upgrading because upgrading procedure always requeued. The upgrading implementation is not idempotent and if current imagestream spec is correct, it still tries to update and requeue.

There was also some duplication of code/logic. Management of annotations is duplicated in upgrading procedure and at component level. Also the list of imagestreams to be upgraded and logic related to "highavailability". Besides the implementation required "tagToRemove" hardcoded which is not `fromVersion` taken from Upgrading basic structure. 

There were two options for upgrading imagestreams. One is implement upgrade logic calling reconciliation logic and the other is do nothing in upgrading procedure. Let reconciliation phase to _upgrade_ the imagestreams. This last way of upgrading only works for some components, like  imagestreams, but might not be valid for others. 

The proposed solution is upgrade in the upgrade phase, but calling reconciliation logic.


eguzki:
This PR invalidates #232 as reconciliation logic respects CR image fields

pkg/controller/apimanager/upgrade_26_to_27.go
```diff
@@ -16,366 +10,18 @@ type Upgrade26_to_27 struct {
...
-	}
-	if changed {
-		u.logger.Info("Zync Database ImageStream upgraded")
-		anImageStreamChanged = true
-	}
-
-	return reconcile.Result{Requeue: anImageStreamChanged}, nil
+func (u *Upgrade26_to_27) upgradeAMPImageStreams() (reconcile.Result, error) {
+	// implement upgrade procedure by reconcile procedure
+	baseReconciler := operator.NewBaseReconciler(u.client, u.apiClientReader, u.scheme, u.logger)
```
miguelsorianod: Lets use client and not apiClientReader. We should avoid using apiclientreader as much as possible

pkg/controller/apimanager/upgrade_26_to_27.go
```diff
@@ -16,366 +10,18 @@ type Upgrade26_to_27 struct {
...
+	baseLogicReconciler := operator.NewBaseLogicReconciler(baseReconciler)
+	reconciler := operator.NewAMPImagesReconciler(operator.NewBaseAPIManagerLogicReconciler(baseLogicReconciler, u.cr))
```
miguelsorianod: This is not enough.

There are ImageStreams that are not reconciled with the AMPImagesReconciler:

I think are DBs in general:
 * Memcached
 * System PostgreSQL or System MySQL depending on what has been configured
 * Backend Redis
 * System Redis

Also for System Postgresql/Mysql and System and Backend redis they should be upgraded only if HA is not set

eguzki: Upgrading DB images is much more than upgrading imagestreams.

This first "upgrade" implementation should only be upgrading AMP images

miguelsorianod: I understand

miguelsorianod:
I'm ok with the change. Just some comments:

> There was an issue testing upgrading procedure. We are upgrading using operator version but upgrading imagestreams is made using 3scale version. That issue caused endless loop upgrading because upgrading procedure always requeued. The upgrading implementation is not idempotent and if current imagestream spec is correct, it still tries to update and requeue.

If it was not idempotent then there was a bug. The upgrade procedure should be idempotent. I don't see that as a reason to change the implementation approach per se.

> Also the list of imagestreams to be upgraded and logic related to "highavailability"

From this I can see the a justification for this change.

> Besides the implementation required "tagToRemove" hardcoded which is not fromVersion taken from Upgrading basic structure.

The tagToRemove does not necessarily has to be the "fromVersion". tagToRemove is a floating tag, that contains major.minor but not major.minor.bugfix so they don't always have the same values. (For example upgrading 2.6.1 to 2.6.2 in both cases the tag would be 2.6).

PR #235, opened 2019-09-20
[WIP] Add api as product cfgmap entry

miguelsorianod:
Adds new api_as_product rolling update in 'system' ConfigMap, in rolling_updates.yml key.
Also includes upgrade procedure for it.

### DIFF
--- a/pkg/controller/apimanager/upgrade_26_to_27.go
+++ b/pkg/controller/apimanager/upgrade_26_to_27.go
@@ -1,7 +1,15 @@
 package apimanager
 
 import (
+	"context"
+	"fmt"
+
+	"gopkg.in/yaml.v2"
+
 	"github.com/3scale/3scale-operator/pkg/3scale/amp/operator"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
@@ -15,6 +23,11 @@ func (u *Upgrade26_to_27) Upgrade() (reconcile.Result, error) {
 		return res, err
 	}
 
+	res, err = u.upgradeSystemRollingUpdatesCfgMap()
+	if res.Requeue || err != nil {
+		return res, err
+	}
+
 	return reconcile.Result{}, nil
 }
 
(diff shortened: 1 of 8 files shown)


pkg/controller/apimanager/upgrade_26_to_27.go
```diff
@@ -379,3 +385,29 @@ func (u *Upgrade26_to_27) upgradeImageStreams() (reconcile.Result, error) {
...
+	}
+
+	var rollingUpdatesTxt string
+	rollingUpdatesTxt, ok := systemCfgMap.Data[cfgMapEntry]
+	if !ok {
+		return reconcile.Result{}, fmt.Errorf("Expected %s key in %s ConfigMap not found", cfgMapEntry, cfgMapName)
+	}
+
+	if !strings.Contains(rollingUpdatesTxt, expectedValue) {
+		u.logger.Info("%s ConfigMap %s entry updated with '%s' appended", cfgMapName, cfgMapEntry, expectedValue)
```
eguzki: structured logging [logr](https://github.com/go-logr/logr) does not work like that. It compiles, but when you run it, you get `odd number of arguments passed as key-value pairs for logging`

pkg/controller/apimanager/upgrade_26_to_27.go
```diff
@@ -379,3 +385,29 @@ func (u *Upgrade26_to_27) upgradeImageStreams() (reconcile.Result, error) {
...
+		return reconcile.Result{}, err
+	}
+
+	var rollingUpdatesTxt string
+	rollingUpdatesTxt, ok := systemCfgMap.Data[cfgMapEntry]
+	if !ok {
+		return reconcile.Result{}, fmt.Errorf("Expected %s key in %s ConfigMap not found", cfgMapEntry, cfgMapName)
+	}
+
+	if !strings.Contains(rollingUpdatesTxt, expectedValue) {
```
eguzki: should we parse yaml?? what about if other key or value contains `expectedValue`??

pkg/controller/apimanager/upgrade_26_to_27.go
```diff
@@ -379,3 +385,29 @@ func (u *Upgrade26_to_27) upgradeImageStreams() (reconcile.Result, error) {
...
+		u.logger.Info("%s ConfigMap %s entry updated with '%s' appended", cfgMapName, cfgMapEntry, expectedValue)
+		rollingUpdatesTxt = rollingUpdatesTxt + "\n" + "  " + expectedValue + "\n"
```
eguzki: you are modifying a copy

I am missing something?

Besides, if we need to modify serialized json/yaml structured content, we should deserialize, modify and serialize again

pkg/controller/apimanager/upgrade_26_to_27.go
miguelsorianod: Expected value is key + value already so if there's a match we have what we need I guess.
Probably it would be a good idea to parse YAML yes.

pkg/controller/apimanager/upgrade_26_to_27.go
miguelsorianod: Totally true.
I was thinking on fmt.Printf when running it. I'll add fmt.Printf then inside the Info if that's ok

pkg/controller/apimanager/upgrade_26_to_27.go
miguelsorianod: Completely true I'll change it

miguelsorianod:
This PR is no longer valid as the behaviour has changed as of https://github.com/3scale/3scale-operator/pull/237 PR

PR #236, opened 2019-09-23
operand image urls as deployment envvars

eguzki:
The operator (not templates) will have the following image resolution procedure:
* If image found in CR, use that image
* If image found in environment variable, use that image
* Else, use default

This is only for apicast, apisonator, porta and zync images. It could be interesting extending functionality to 
* Backend Redis image
* System redis image
* System mysql image
* System postgresql image
* system memcached image
* zync postgresql image

This change allows setting images in CSV manifests


miguelsorianod:
Just to clarify and so it's reflected here, the motivation of this PR to allow setting the images through the operator's Deployment (and thus CSV manifests) via environment variables is because our internal pipelines allow us to perform manipulation of those environment variables during the productization process, helping ease the productization and testing of those images.

Should we add the system memcached environment variable too? Memcached is productized by us and thus potentially there might be a release where we have a brew image for it. Maybe we can leave it for when the situation arises. What do you think? This would impact the upgrade procedure too.

For this release we should check the URL paths between the final image URLs and our pipeline ones because if the productization pipeline does not change them and they differ the images would not be on the expected URLs

eguzki:
system memcached image also available by environment variable, as it can be result of internal productization

eguzki:
changed internal names from `porta` to `system`.

pkg/3scale/amp/template/adapters/images.go
```diff
@@ -15,39 +14,37 @@ func NewImagesAdapter() Adapter {
...
 
 func (i *ImagesAdapter) Parameters() []templatev1.Parameter {
-	imageProvider := product.CurrentImageProvider()
-
 	return []templatev1.Parameter{
 		templatev1.Parameter{
 			Name:     "AMP_BACKEND_IMAGE",
 			Required: true,
-			Value:    imageProvider.GetBackendImage(),
+			Value:    component.ApisonatorImageURL(),
```
miguelsorianod: The same applies here. References to Apisonator should also be renamed to Backend (envvars, methods, ...)

eguzki:
This change does not make sense at all, this is upstream and should use upstream names like Apisonator and Porta. Productized names should not be used. 

Being said that, changed internal names from `apisonator` to `backend`

miguelsorianod:
> This change does not make sense at all, this is upstream and should use upstream names like Apisonator and Porta. Productized names should not be used.
> 
> Being said that, changed internal names from `apisonator` to `backend`
> 
> ready for review @miguelsorianod

I understand your concern. When operator was started it was for "product" versions only. None of the images were from upstream code so a decision was made to use non-upstream names for references to those components.
You would have the same concern now if we were to choose upstream names, but for the productized versions, as they do not use upstream components and it would not make sense "either". Changing it to upstream names then would also mean changing the CR field names, which would in turn make the CRDs different (incompatible change) and also the names not be appropriate in the productized versions. Those are the reasons why I asked to change it to non-upstream names.

PR #240, opened 2019-09-27
Add routes checking in e2e tests

miguelsorianod:
This PR adds code in the e2e test that checks that all routes that should be created by default in an APIManager deployment are indeed created

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,13 +33,13 @@ commands:
     steps:
       - restore_cache:
           keys:
-            - v2-3scaleoperator-vendor-{{ arch }}-{{ checksum "go.sum" }}
+            - v3-3scaleoperator-vendor-{{ arch }}-{{ checksum "go.sum" }}
       - run:
           name: Install operator dependencies
           command: |
             make vendor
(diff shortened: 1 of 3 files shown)


miguelsorianod:
Ready for review @eguzki

.circleci/config.yml
```diff
@@ -33,13 +33,13 @@ commands:
     steps:
       - restore_cache:
           keys:
-            - v2-3scaleoperator-vendor-{{ arch }}-{{ checksum "go.sum" }}
+            - v3-3scaleoperator-vendor-{{ arch }}-{{ checksum "go.sum" }}
```
eguzki: why?

miguelsorianod: We are caching vendor directory, however we are now using more packages in a "module"(or repo) that we were not using before. Due to the vendor directory was cached the new files belonging to this repository were not used.
It is the case of '  github.com/openshift/client-go v3.9.0+incompatible ' in go.mod. We still need to use this, but now we use an additional package there.
By increasing the version number we force CircleCI to redownload the dependencies and are now included.

PR #241, opened 2019-09-30
Update CircleCI configuration elements to expected configuration values

miguelsorianod:
Apply some changes in CircleCI config to be compliant with current CircleCI configuration version

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,7 +106,7 @@ commands:
 
             oc get events | egrep ' Failed ' || :
             oc get events -o json | jq '[.items[] | select(.reason == "Failed") | debug ] | length == 0'  --exit-status
-          no_output_timeout: 1800
+          no_output_timeout: 30m
 
   deploy-3scale-eval-from-template:
     steps:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -232,7 +231,9 @@ jobs:
             - images/3scaleoperator-image.tar
 
   tag-operator-image-master:
-    machine: true
+    machine:
```
miguelsorianod: Obtained expected / available values from here:
https://circleci.com/docs/2.0/configuration-reference/#machine

It seems that the old default was circleci/classic:latest so I set the image to that

.circleci/config.yml
```diff
@@ -218,8 +218,7 @@ jobs:
     working_directory: /go/src/github.com/3scale/3scale-operator
     steps:
       - setup_remote_docker:
-          reusable: true
-          exclusive: false
```
miguelsorianod: https://circleci.com/docs/2.0/docker-layer-caching/#remote-docker-environment

exclusive is not used anymore (and it is true always now), and reusable has been renamed to docker_layer_caching

PR #243, opened 2019-09-30
do not cache vendor folder

eguzki:
`vendor` folder should not be cached.

Go modules cache directory should only be cached, i.e. `/go/pkg` folder. 

Same go.sum (same modules and releases) may lead to different `vendor` folder, because vendor only contains used packages, not used modules. `/go/pkg` folder contains used modules and can be identified uniquely by `go.sum`. Does not happen the same with `vendor` folder.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,15 +33,16 @@ commands:
     steps:
       - restore_cache:
           keys:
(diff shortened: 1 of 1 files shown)


miguelsorianod:
Shouldn't we increase the cached key version to make the changes effective?

eguzki:

PR #248, opened 2019-10-07
storage and database defaults refactor

eguzki:
Refactor system storage and database defaults and reconciliation.

CR will be updated with defaults accordingly

```
system:
    appSpec:
      replicas: 1
    database:
      mysql:
        image: centos/mysql-57-centos7
    fileStorage:
      persistentVolumeClaim: {}
    sidekiqSpec:
      replicas: 1

```


eguzki:
Closing as image defaults cannot be set in CR. If image are set in defaults, upgrading procedure will not be able to update images as they are in CR.

PR #251, opened 2019-10-11
limit number of channels to two

eguzki:


eguzki:
waiting for a response on https://github.com/operator-framework/operator-lifecycle-manager/issues/1072

If no answer before few days, we will merge

PR #253, opened 2019-10-16
Add basic unit test for APIManager types package

miguelsorianod:
This PR prepares project to execute unit tests via Makefile and integrated into CircleCI and adds basic unit tests for APIManager types

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -319,6 +319,18 @@ jobs:
       - oc-status
       - push-3scale-images-to-quay
 
+  run-unit-tests:
+    docker:
+      - image: circleci/golang:1.12.5
+    working_directory: /go/src/github.com/3scale/3scale-operator
+    steps:
+      - checkout
+      - run: make vendor
+      - run:
+          name: Run unit tests in pkg folder
+          command: |
+            make unit
+
   run-operator-e2e-test:
     machine:
       image: circleci/classic:latest
@@ -392,6 +404,7 @@ workflows:
     jobs:
       - license-check
       - test-crds
+      - run-unit-tests
       - verify-manifest
       - install-operator
       - run-operator-e2e-test:
--- a/pkg/apis/apps/v1alpha1/apimanager_types.go
+++ b/pkg/apis/apps/v1alpha1/apimanager_types.go
@@ -20,6 +20,20 @@ const (
 	OperatorVersionAnnotation   = "apps.3scale.net/threescale-operator-version"
 )
 
+const (
+	defaultAppLabel                    = "3scale-api-management"
+	defaultTenantName                  = "3scale"
+	defaultImageStreamImportInsecure   = false
+	defaultResourceRequirementsEnabled = true
+)
+
+const (
+	defaultApicastManagementAPI = "status"
+	defaultApicastOpenSSLVerify = false
+	defaultApicastResponseCodes = true
+	defaultApicastRegistryURL   = "http://apicast-staging:8090/policies"
+)
+
 // APIManagerSpec defines the desired state of APIManager
 // +k8s:openapi-gen=true
 type APIManagerSpec struct {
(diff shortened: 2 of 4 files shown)


.circleci/config.yml
```diff
@@ -392,6 +404,7 @@ workflows:
     jobs:
       - license-check
       - test-crds
+      - run-unit-tests
```
miguelsorianod: This job has been added in the "operator" workflow although the unit tests could execute tests that are related to templates too. Might it be better to create another workflow called "common" where execute them?

pkg/apis/apps/v1alpha1/apimanager_types.go
```diff
@@ -323,29 +337,29 @@ func (apimanager *APIManager) setApicastSpecDefaults() bool {
 	changed := false
 	spec := &apimanager.Spec
 
-	defaultApicastManagementAPI := "status"
-	defaultApicastOpenSSLVerify := false
-	defaultApicastResponseCodes := true
-	defaultApicastRegistryURL := "http://apicast-staging:8090/policies"
+	tmpDefaultApicastManagementAPI := defaultApicastManagementAPI
```
miguelsorianod: We create a temporary copy of the constant because below in code we need the address of the variable and in Golang you cannot take it from a constant.

It's not very ideal due to there's duplication by creating the temporary variables here and in the test file.
Another option that I've thought of is creating unexported methods (starting with lowercase) that return a pointer to string and do not define them as constants. The unexported methods would be called from both the code and the tests. What do you think?

pkg/apis/apps/v1alpha1/apimanager_types_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+
+	"github.com/3scale/3scale-operator/pkg/3scale/amp/product"
+	"github.com/3scale/3scale-operator/version"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestSetDefaults(t *testing.T) {
```
miguelsorianod: The organization of the test file I've thought is the following one:
 * You have one test function (uppercase) per exported method of the file you want to test (in this case apimanager_types.go). In this function you have several subtests that you add with t.Run for each of them to test the different situations in that method that you want to check.
Another alternative would be to create one exported test function per situation you want to test per method you want to test. The downside of this is that you might end up with duplicated code and you would have more exported methods.

What do you think?

pkg/apis/apps/v1alpha1/apimanager_types_test.go
```diff
@@ -0,0 +1,113 @@
...
+			},
+		},
+	}
+
+	t.Run("Basic defaults are correct", testBasicAPIManagerDefaults(&inputAPIManager, &expectedAPIManager))
+}
+
+func testBasicAPIManagerDefaults(input, expected *APIManager) func(t *testing.T) {
+	return func(t *testing.T) {
+		tmpInput := input.DeepCopy()
```
miguelsorianod: Here we need to copy the input because the SetDefault methods modifies it

miguelsorianod:
Also check the logs for the unit tests to see how it would look

.circleci/config.yml
```diff
@@ -421,6 +434,7 @@ workflows:
             - generator
   nightly:
     jobs:
+      - run-unit-tests
```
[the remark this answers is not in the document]
eguzki: When PR is merged, unittests are executed for master branch. Unittests depend exclusively on source code from this repo. Nightly (integration) tests usually involve components not included in source code of this repo, so they could fail even when master branch of this repo did not change.

pkg/apis/apps/v1alpha1/apimanager_types.go
eguzki: I think it is ok the way it is now.

pkg/apis/apps/v1alpha1/apimanager_types_test.go
eguzki: If you define the input inside, it would not be shared among other subtests.

I would do the `DeepCopy` outside subtest. This way high level test ensures all subtests get expected input data structure and not modified by some subtest

.circleci/config.yml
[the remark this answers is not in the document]
miguelsorianod: :+1: I added it so we had a "double-check" and explicit output about them on nightly too but I agree there's no really a need as unit tests shouldn't fail if they have passed on master as they should only depend on code from the repo.
I've removed it.

pkg/apis/apps/v1alpha1/apimanager_types_test.go
miguelsorianod: makes sense :+1: I'll change it

miguelsorianod: Changed

miguelsorianod:
Ready for re-review @eguzki

PR #254, opened 2019-10-16
Add APIManager controller basic unit tests

miguelsorianod:


pkg/controller/apimanager/apimanager_controller_test.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		t.Fatalf("Unable to add Image scheme: (%v)", err)
+	}
+	routev1.AddToScheme(s)
+	if err != nil {
+		t.Fatalf("Unable to add Route scheme: (%v)", err)
+	}
+
+	// Create a fake client to mock API calls.
+	cl := fake.NewFakeClient(objs...)
```
miguelsorianod: The fake client basically instantly records the object as created when it finds a Create request. This means the update of the "status" section will always succeed(there will not be any requeue in that case)

pkg/controller/apimanager/apimanager_controller_test.go
```diff
@@ -0,0 +1,110 @@
...
+		t.Error("reconcile did not finish end of reconciliation as expected. APIManager should have been reconciled at this point")
+	}
+
+	finalAPIManager := &appsv1alpha1.APIManager{}
+	err = r.Client().Get(context.TODO(), req.NamespacedName, finalAPIManager)
+	if err != nil {
+		t.Fatalf("get APIManager: (%v)", err)
+	}
+
+	backendListenerExistingReplicas := finalAPIManager.Spec.Backend.ListenerSpec.Replicas
```
miguelsorianod: Not exactly what to test/at what level so I checked arbitrarily that the replicas where the expected ones (1, being set by default).
Is that good enough?

pkg/controller/apimanager/apimanager_controller_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+	res, err = r.Reconcile(req)
+	if err != nil {
+		t.Fatalf("reconcile: (%v)", err)
+	}
+
+	if res.Requeue {
+		t.Error("reconcile did not finish end of reconciliation as expected. APIManager should have been reconciled at this point")
+	}
+
```
miguelsorianod: If you look up at until this point you will see that I check that the first reconcile call returns a requeue, as we expect to some defaults to be set with the provided apimanager), and that I check that the second call does not return a requeue as it is expected all objects are created successfully.

miguelsorianod:
There was a new dependency pending to be added into decisions file. Added a new commit that regenerates licenses and adds the new dependency to the dependency_decisions file.
Ready to re-review @eguzki

eguzki:
go.mod and go.sum do not change at all? Does not make sense

miguelsorianod:
I regenerated go.sum with go mod tidy.
It can be reviewed again @eguzki

PR #260, opened 2019-10-23
Upgrade System's AMP_RELEASE on operator upgrade

miguelsorianod:
Until now, for every release AMP_RELEASE environment variable was updated to the new 3scale version as this variable was used to display the version in system's UI.
We lacked this part on the operator's upgrade mechanism. In templates it's already documented on the upgrade documentation.
System team is working on changing the behaviour on how the version is displayed and in fact currently they are not really relying on this value and might not be needed but the new mechanism seems it not being applied either at this moment, and they might not change it/fix it for this release so we are adding this change so they can resort to this in case they need to revert back to the old behavior.

### DIFF
--- a/pkg/controller/apimanager/upgrade_26_to_27.go
+++ b/pkg/controller/apimanager/upgrade_26_to_27.go
@@ -6,7 +6,9 @@ import (
 	"reflect"
 
 	"github.com/3scale/3scale-operator/pkg/3scale/amp/operator"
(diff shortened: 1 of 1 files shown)


miguelsorianod:
Changing the ConfigMap entry does not trigger an update of the DeploymentConfigs as Kubernetes is not aware of environment variable changes when reference a ConfigMap/Secret, but the change will be recognized as the DeploymentConfigs will be redeployed when the image is changed (as desired)

PR #263, opened 2019-10-24
system-app pre hook command update: MASTER_ACCESS_TOKEN as env var

eguzki:


miguelsorianod:
I looked at this some time before and I seem to remember MASTER_ACCESS_TOKEN might not be an environment variable but a parameter of the command (not sure whether of the rake task, bundle or exec...).

We should make sure of what that exactly is. Also, if it ends up being an environment variable, does it reach the desired command with bundle exec in the middle? would affect the other commands? We should get some way to verify it from System team.

Also clarify that this does not "fix" the pre-hook pod command. It should be working without the changes in this PR although I think it would be great if we could move it to an env var gathered from a secret if it ends up being an env var.

Finally, if this change is made the templates upgrade documentation should be updated to:
 *  Add the environment variable into the pre-hook pod
 * Change the pre-hook pod command modification content

eguzki:
MASTER_ACCESS_TOKEN is read as ENV variable. 

Rake does provide another way of setting environment variables:

```
task :first do
  puts "VAR in `first` is #{ENV['VAR'].inspect}"
end
```

```
$ rake first VAR=b
VAR in `first` is "b"
```

http://jacobswanner.com/development/2017/rake-task-arguments/

Now all rake tasks will have access to MASTER_ACCESS_TOKEN as ENV var, but should not be an issue.

miguelsorianod:
That's good news :+1: . If you are confident that the other new rake tasks having the new variable defined do not change their behaviour I'm ok with it :+1: . Keep in mind that even though the CI tests pass (pods are deployed) the behaviour might be inadvertently changed.

Because it's a change that affects existing installations and requires changes to non-reconcilable fields an upgrade should be applied for the operator. We can implement the code into the current release upgrade code (implement it in current upgrade code) or into the next one (create new upgrade code for the new release). Upgrade should be implemented and verify that it works together with this changes.

eguzki:
review @miguelsorianod

PR #265, opened 2019-10-30
Operator unittest

eguzki:
Code coverage (`pkg` folder) in master branch:  **36.7%**
Current code coverage (`pkg` folder) :  **60.9%**

- [x] upgrade
- [x] ampimages
- [x] apicast
- [x] backend
- [x] zync
- [x] secrets
- [x] configmaps
- [x] images
- [x] imagestreams
- [x] mysql
- [x] s3
- [x] service_account
- [x] HA
- [x] route
- [x] service
- [x] PVC
- [x] postgresql
- [x] postgresql image
- [x] redis
- [x] role
- [x] role binding
- [x] sytem
- [x] DC
- [x] memcached
- [x] mysql image

### DIFF
--- a/pkg/3scale/amp/operator/persistent_volume_claim_reconciler_test.go
+++ b/pkg/3scale/amp/operator/persistent_volume_claim_reconciler_test.go
@@ -1,0 +1,224 @@
+package operator
+
+import (
+	"context"
+	"testing"
+
+	appsv1alpha1 "github.com/3scale/3scale-operator/pkg/apis/apps/v1alpha1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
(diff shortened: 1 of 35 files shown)


eguzki:

pkg/3scale/amp/operator/persistent_volume_claim_reconciler_test.go
```diff
@@ -0,0 +1,224 @@
...
+		t.Fatalf("reconciled owner reference is not apimanager, expected: %s, got: %s", name, reconciled.GetOwnerReferences()[0].Name)
+	}
+}
+
+type myCustomPVCReconciler struct {
+}
+
+func (r *myCustomPVCReconciler) IsUpdateNeeded(desired, existing *v1.PersistentVolumeClaim) bool {
+	tmp := "myClass"
+	existing.Spec.StorageClassName = &tmp
```
miguelsorianod: Although it's true that we are reconciling the value of StorageClass, PVCs do not support changing their StorageClassName value once the PVC is created and bound.

Trying to edit the field results i nthe following message:
```
# * spec: Forbidden: is immutable after creation except resources.requests for bound claims
```

eguzki: For PVC we are only using CreateOnly reconcilers, so we are no reconciling StorageClass (if I am not wrong).

In this test, the purpose was to check that PVC reconciler updates if reconciler type is not CreateOnly.

PR #268, opened 2019-11-05
simple amp build process

eguzki:


.circleci/config.yml
```diff
@@ -163,20 +164,7 @@ commands:
...
             oc new-app -f pkg/3scale/amp/manual-templates/amp/build.yml --allow-missing-imagestream-tags
-            set -x
-            oc cancel-build $(oc get bc --output=name)
-            for bc in $(oc get bc --output=name); do oc start-build $bc --wait --follow & done
-            echo "List of build image processes:"
-            for job in $(jobs -p); do
-              ps aux | grep $job
-            done
-            echo "Waiting for the build processes to finish"
-            for job in $(jobs -p); do
```
miguelsorianod: The reason this code was added was because builds are started in background. The reason they are started in background is to allow parallelism of the builds, as some of them can take some time.
Code to keep track of the progress of the background jobs was added to check for the exit status of them.

The new code serializes the execution of the builds. I'd prefer to keep the parallelism of jobs building.

eguzki: This image building happens only at nightly, parallelism is not required. It does not have to be fast.

The building code is simpler. No need to cancel jobs, starts all in parallel and then wait for all are finised.

In addition, you get all the logs in the CI build, easier to debug

Buildconfig triggers are not necessary, we do no use build configs in deployments. We do not need to keep imagestreams up to date with changes with git repos.

miguelsorianod: > The building code is simpler. No need to cancel jobs, starts all in parallel and then wait for all are finised.

I'm ok with the BuildConfigs not having triggers :+1: . That's a different change than the parallelization/serialization of builds.

> The building code is simpler. No need to cancel jobs, starts all in parallel and then wait for all are finised.

If I'm not mistaken with the new code the builds are not started in parallel, it goes one by one.

> In addition, you get all the logs in the CI build, easier to debug

Were some logs not shown in the "old" way? That'd be a good point.

Keep in mind that we sometimes execute it manually so it might be interesting to be fast and we are not the only ones using the BuildConfigs (QE does for example)

eguzki: > I'm ok with the BuildConfigs not having triggers +1 . That's a different change than the parallelization/serialization of builds.
Not entirely. Triggers make all builds start immediately when you create the buildconfigs. Then you need to cancel them and start them again manually in background for parallel building. To keep simple building process avoiding cancel step is nice to have.

> If I'm not mistaken with the new code the builds are not started in parallel, it goes one by one.
I mean that there is no need to
* cancel jobs
* starts all in parallel
* wait for all are finised (sync logic)

> Were some logs not shown in the "old" way?
You are right, you do not get more logs. However, when you have parallelized bulds, logs are coming from several processes in a "random" way and hard to relate log lines. With this change, builds are started one after another and watch its logs until it completes or fails. Ordered logs.

> Keep in mind that we sometimes execute it manually so it might be interesting to be fast and we are not the only ones using the BuildConfigs (QE does for example)
This code is circleCI code. This code is not shared with any other building process. The buildconfig file is shared but it has anything to do with the discussion. Scripts for manual building can implement parallelization to be faster, I agree. I have my own

PR #273, opened 2019-11-06
2.6 stable prod backport fix databaseless postgresql deployment

eguzki:
Fix system database postgresql DC being created when spec.highAvailability.enabled

### DIFF
--- a/pkg/3scale/amp/component/highavailability.go
+++ b/pkg/3scale/amp/component/highavailability.go
@@ -19,9 +19,10 @@ const (
 )
 
 var highlyAvailableExternalDatabases = map[string]bool{
-	"backend-redis": true,
-	"system-redis":  true,
-	"system-mysql":  true,
+	"backend-redis":     true,
+	"system-redis":      true,
+	"system-mysql":      true,
+	"system-postgresql": true,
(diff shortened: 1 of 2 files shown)


miguelsorianod:
When deploying System with PostgreSQL the following additional elements are created:
system-postgresql Service
postgresql-data PersistentVolumeClaim
system-postgresql ImageStream

The Service and the ImageStream are correctly removed already because the variable highlyAvailableExternalDatabases is reused when checking for the removal of them.
We should add code to remove the postgresql-data PersistentVolumeClaim when it exists.

eguzki:
But when `spec.highAvailability` is enabled, with this PR, PVC's are no longer created, nor services or imagestream.

Shouldn't be better to remove post process logic?

miguelsorianod:
Hi,

It would be good to remove some of the postprocess logic yes, but in order to not change behavior in 2.6 itself we should still maintain the postprocess step that increases the replicas number of some DCs and maybe the UpdateDatabasesURLS method (not sure of this latest one)

That or think of a refactor of how to set external DBs, including having it working for templates too.

eguzki:

PR #275, opened 2019-11-06
storage and database defaults refactor

eguzki:
* Fix storageClassName property location
* CR databases defaults cannot be `mysql`. When external databases are configured, this leads to inconsistent information. Changed defaults set for the CR. Mysql is still the *default* database if none configured
* CR storage defaults removed. PVC is still the *default* storage if none configured
* Upgrade procedure to remove databases and storage defaults

### DIFF
--- a/pkg/3scale/amp/operator/system_reconciler.go
+++ b/pkg/3scale/amp/operator/system_reconciler.go
@@ -114,16 +114,19 @@ func NewSystemReconciler(baseAPIManagerLogicReconciler BaseAPIManagerLogicReconc
 }
 
 func (r *SystemReconciler) reconcileFileStorage(system *component.System) error {
-	if r.apiManager.Spec.System != nil && r.apiManager.Spec.System.FileStorageSpec != nil {
-		if r.apiManager.Spec.System.FileStorageSpec.PVC != nil {
-			return r.reconcileSharedStorage(system.SharedStorage())
-		} else if r.apiManager.Spec.System.FileStorageSpec.S3 != nil {
-			return r.reconcileS3AWSSecret(system.S3AWSSecret())
-		} else {
-			return fmt.Errorf("No FileStorage spec specified. FileStorage is mandatory")
-		}
+	if r.apiManager.Spec.System.FileStorageSpec.PVC != nil && r.apiManager.Spec.System.FileStorageSpec.S3 != nil {
+		return fmt.Errorf("Only one system FileStorage can be chosen at the same time")
 	}
-	return nil
+
+	var err error
+	if r.apiManager.Spec.System.FileStorageSpec.PVC != nil {
+		err = r.reconcileSharedStorage(system.SharedStorage())
+	} else if r.apiManager.Spec.System.FileStorageSpec.S3 != nil {
+		err = r.reconcileS3AWSSecret(system.S3AWSSecret())
+	} else {
+		err = fmt.Errorf("No FileStorage spec specified. FileStorage is mandatory")
+	}
+	return err
 }
 
 func (r *SystemReconciler) reconcileS3AWSSecret(desiredSecret *v1.Secret) error {
--- a/pkg/3scale/amp/operator/system_reconciler.go
+++ b/pkg/3scale/amp/operator/system_reconciler.go
@@ -114,19 +114,13 @@ func NewSystemReconciler(baseAPIManagerLogicReconciler BaseAPIManagerLogicReconc
 }
 
 func (r *SystemReconciler) reconcileFileStorage(system *component.System) error {
-	if r.apiManager.Spec.System.FileStorageSpec.PVC != nil && r.apiManager.Spec.System.FileStorageSpec.S3 != nil {
-		return fmt.Errorf("Only one system FileStorage can be chosen at the same time")
+	if r.apiManager.Spec.System != nil &&
+		r.apiManager.Spec.System.FileStorageSpec != nil &&
+		r.apiManager.Spec.System.FileStorageSpec.S3 != nil {
+		return r.reconcileS3AWSSecret(system.S3AWSSecret())
 	}
 
-	var err error
-	if r.apiManager.Spec.System.FileStorageSpec.PVC != nil {
-		err = r.reconcileSharedStorage(system.SharedStorage())
-	} else if r.apiManager.Spec.System.FileStorageSpec.S3 != nil {
-		err = r.reconcileS3AWSSecret(system.S3AWSSecret())
-	} else {
-		err = fmt.Errorf("No FileStorage spec specified. FileStorage is mandatory")
-	}
-	return err
+	return r.reconcileSharedStorage(system.SharedStorage())
 }
 
 func (r *SystemReconciler) reconcileS3AWSSecret(desiredSecret *v1.Secret) error {
--- a/pkg/3scale/amp/operator/system_reconciler.go
+++ b/pkg/3scale/amp/operator/system_reconciler.go
@@ -114,8 +114,7 @@ func NewSystemReconciler(baseAPIManagerLogicReconciler BaseAPIManagerLogicReconc
 }
 
 func (r *SystemReconciler) reconcileFileStorage(system *component.System) error {
-	if r.apiManager.Spec.System != nil &&
-		r.apiManager.Spec.System.FileStorageSpec != nil &&
+	if r.apiManager.Spec.System.FileStorageSpec != nil &&
 		r.apiManager.Spec.System.FileStorageSpec.S3 != nil {
 		return r.reconcileS3AWSSecret(system.S3AWSSecret())
 	}
--- a/pkg/3scale/amp/operator/upgrade.go
+++ b/pkg/3scale/amp/operator/upgrade.go
@@ -281,15 +281,12 @@ func (u *UpgradeApiManager) upgradeSystemRedisImageStream() (reconcile.Result, e
 }
 
 func (u *UpgradeApiManager) upgradeSystemDatabaseImageStream() (reconcile.Result, error) {
-	if u.Cr.Spec.System.DatabaseSpec.MySQL != nil {
-		return u.upgradeSystemMySQLImageStream()
-	}
-
 	if u.Cr.Spec.System.DatabaseSpec.PostgreSQL != nil {
 		return u.upgradeSystemPostgreSQLImageStream()
+	} else {
+		// default is MySQL
+		return u.upgradeSystemMySQLImageStream()
 	}
-
-	return reconcile.Result{}, fmt.Errorf("System database is not set")
 }
 
 func (u *UpgradeApiManager) upgradeSystemMySQLImageStream() (reconcile.Result, error) {
--- a/pkg/3scale/amp/operator/upgrade.go
+++ b/pkg/3scale/amp/operator/upgrade.go
@@ -29,7 +29,12 @@ type UpgradeApiManager struct {
 }
 
 func (u *UpgradeApiManager) Upgrade() (reconcile.Result, error) {
-	res, err := u.upgradeSystemAppPreHookPodEnv()
+	res, err := u.upgradeAPIManagerCR()
+	if res.Requeue || err != nil {
+		return res, err
+	}
+
+	res, err = u.upgradeSystemAppPreHookPodEnv()
 	if res.Requeue || err != nil {
 		return res, err
 	}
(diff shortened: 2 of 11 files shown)


pkg/apis/apps/v1alpha1/apimanager_types.go
```diff
@@ -513,59 +504,36 @@ func (apimanager *APIManager) setSystemSpecDefaults() (bool, error) {
...
 
-	if systemSpec.FileStorageSpec.PVC != nil && systemSpec.FileStorageSpec.S3 != nil {
-		return changed, fmt.Errorf("Only one FileStorage can be chosen at the same time")
-	}
-
-	if systemSpec.FileStorageSpec.PVC == nil && systemSpec.FileStorageSpec.S3 == nil {
-		systemSpec.FileStorageSpec.PVC = defaultFileStorageSpec.PVC
+	// No default values for S3
+	// PVC
+	if systemSpec.FileStorageSpec.PVC == nil {
```
miguelsorianod: If I understand correctly, this makes the  systemSpec.FileStorageSpec.PVC to be always defined.
This means that the resources related to PVC are always created now, which is not what we desire. We want to create them only when PVC is selected, not when S3 is selected.
What defines whether one of the storage options is used or not is whether it is nil or not.

eguzki: good catch!, fixed

eguzki:
all comments fixed, ready for review @miguelsorianod

pkg/apis/apps/v1alpha1/apimanager_types.go
```diff
@@ -513,59 +504,36 @@ func (apimanager *APIManager) setSystemSpecDefaults() (bool, error) {
...
-	}
-
 	if systemSpec.DatabaseSpec == nil {
-		systemSpec.DatabaseSpec = defaultDatabaseSpec
+		systemSpec.DatabaseSpec = &SystemDatabaseSpec{}
 		changed = true
-		return changed, nil
 	}
 
-	if systemSpec.DatabaseSpec.MySQL != nil && systemSpec.DatabaseSpec.PostgreSQL != nil {
```
miguelsorianod: What's the motivation of moving CR validation to reconciliation?
To me conceptually it makes more sense to validate the schema before applying the reconciliation logic. I think ideally the reconciliation should take a valid CR / Internal object

miguelsorianod:
> CR defaults cannot be mysql. When external databases are configured, this leads to inconsistent information. Changed defaults set for the CR. Mysql is still the default database if none configured

Why do you say it leads to "inconsistent information" ? Is it because the user can see that section with mysql: {} set? At the end, you are still considering it set with this changes with an implicit default. Maybe we should only set it when highAvailability is NOT enabled and consider the database field completely optional and when none of it is set it means no DB should be created? Maybe instead of having a highAvailability field we should make each of the DBs be able to specify the source being "external" or "embedded" or something like that? I'm just thinking here what options do we have. It just seems there's something wrong when we explicitly set defaults for all other "union types" we have but not for this one.

pkg/apis/apps/v1alpha1/apimanager_types.go
eguzki: Syntactic validation and at some extent semantic validation is done by CRD openapi validation. Most of the higher level semantic validation is done by option builders, in the reconciliation loop. Hence, moving validation to the reconciliation loop in a distributed way, i.e. each components validates each own part.

I also like having a single entrypoint at the beginning of the reconciliation loop to do all validation checks for high level semantic issues. 

Definitely, validation at a method meant to set defaults is not conceptually correct.

eguzki:
Currently on master branch (changes of this PR not included)
CR with external databases enabled:

```yaml
kind: APIManager
metadata:
  name: example-apimanager
spec:
  wildcardDomain: lvh.me
  highAvailability:
    enabled: true
```
After applying defaults:
```yaml
kind: APIManager
spec:
  ...
  highAvailability:
    enabled: true
  system:
    database:
      mysql: {}
   ...
```
And there is no mysql database deployed. That is what I mean with "inconsistent information".

With the changes in this PR, this inconsistency is removed.

Conceptually speaking, I dislike setting defaults to the CR. It is a limitation. Specifically for future upgrades, as once defaults are set, the operator can not differentiate if it is a default or set by the customer. Defaults values exist in code, i.e. may change from release to release. If we want to give some information to the customer about what options have been used to deploy 3scale, we should use `status` field of the CR for that, because it was designed for that.

pkg/apis/apps/v1alpha1/apimanager_types.go
miguelsorianod: I don't like the idea of validating on the reconciliation loop. I'd prefer knowing that when the reconciliation logic is applied we start from valid contents.

I thought of having a validate method separate from the setDefaults too @eguzki in apimanager_types.go and I'd like to go that way but I found the following dilemma, which is why I left validation on the defaults method:

What/when is considered a CR "valid"? Should the validation be performed before or after setting the defaults?

You can have an "invalid" CR that "becomes valid" after defaults are set for it. For example, the replicas fields are optional but we set them as we need a value for them defined. Same happens for the selected storage type.

If you "validate" before setting the defaults you might detect that the CR is invalid but it is not really invalid, it's just that the defaults have not been set for it yet (replicas or storage case explained before).
If you "validate" after setting the defaults then you might try to set defaults from fields that are not initialized and that you expect them to be (a really invalid object). Shouldn't defaults be applied only when we know that we have a valid object?

What do you think?

miguelsorianod:
>  If we want to give some information to the customer about what options have been used to deploy 3scale, we should use `status` field of the CR for that, because it was designed for that.

That's not what I understand from the api conventions document in the defaulting and late initialization sections:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#defaulting
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#late-initialization

pkg/apis/apps/v1alpha1/apimanager_types.go
eguzki: As I see:
Validation should be performed **before** applying defaults. Applying defaults always **adds** info, never update or delete.

> You can have an "invalid" CR that "becomes valid" after defaults are set for it

I disagree. CR without defaults is also valid, as defaults are always optional fields. As I see, invalid state is when, for instance, required attribute does not exist, or when two excluding fields exist (like pvc and s3). As I see, default concept lives in the code. If user does not set replicas, the object still have the `replica` field with some value set by the code. That is being optional with a default value. Hence, replica will always exist internally, even when the user does not set explicitly. So, it will be valid even when replica is not set.

As I said, the operator should validate before applying defaults. It could happen that after validating and after setting defaults the CR could be invalid. But in that case, this is a bug we should fix.

I do not see any good reason to apply defaults to the CR. We could "serialize" the internal CR object in status if we want to report that info to the customer.

eguzki:
Late initialization and defaulting conventions are not rules. Think about imagestreams, actual image sha's are shown in status and may change in time.

For replicas, I think it is a good convention. For other attributes I think it is a limitation for upgrades

eguzki:
For attributes existing in the CR that are not reconcilable, definitely setting defaults is not a good convention IMHO. They should not even exist in the CRD, but If they exist, they should not be "defaulted" in the CR. 
IMHO

eguzki:
needs review @miguelsorianod

pkg/apis/apps/v1alpha1/apimanager_types.go
miguelsorianod: > I do not see any good reason to apply defaults to the CR. We could "serialize" the internal CR object in status if we want to report that info to the customer.

I think that's not what status should be used for. Status is the "actual" status whereas "spec" is the desired status.
I don't think it's a bad idea to apply defaults to the CR. As said in the API conventions https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#defaulting incorportaing the default values into the Spec ensures that the Spec depicts the full desired state so that it is easier for the system to determine how to achieve the state, and for the user to know what to anticipate.

pkg/3scale/amp/operator/system_reconciler.go
```diff
@@ -114,16 +114,19 @@ func NewSystemReconciler(baseAPIManagerLogicReconciler BaseAPIManagerLogicReconc
...
 func (r *SystemReconciler) reconcileFileStorage(system *component.System) error {
-	if r.apiManager.Spec.System != nil && r.apiManager.Spec.System.FileStorageSpec != nil {
-		if r.apiManager.Spec.System.FileStorageSpec.PVC != nil {
-			return r.reconcileSharedStorage(system.SharedStorage())
-		} else if r.apiManager.Spec.System.FileStorageSpec.S3 != nil {
-			return r.reconcileS3AWSSecret(system.S3AWSSecret())
-		} else {
-			return fmt.Errorf("No FileStorage spec specified. FileStorage is mandatory")
-		}
+	if r.apiManager.Spec.System.FileStorageSpec.PVC != nil && r.apiManager.Spec.System.FileStorageSpec.S3 != nil {
```
miguelsorianod: I still think we should really remove validation from the reconcilers if possible. As I see reconcilers should already receive consistent state of the CR and ensuring that should not be responsability of the reconcilers as I commented in more detail in https://github.com/3scale/3scale-operator/pull/275/files#r347341179

miguelsorianod:
> For attributes existing in the CR that are not reconcilable, definitely setting defaults is not a good convention IMHO. They should not even exist in the CRD, but If they exist, they should not be "defaulted" in the CR.
> IMHO

I left a new comment in https://github.com/3scale/3scale-operator/pull/275/files#r357006175 about this

pkg/3scale/amp/operator/system_reconciler.go
eguzki: Definitely doing validations in a function meant to set defaults does not fit in any best patterns guide

Validations should be done in a single place at the beginning, or by each component reconciler, like it is being doing already.

It is ok for you if I remove validation fix and leave it 'ugly' but still doing other items?
* Fix storageClassName property location
* CR defaults cannot be mysql. When external databases are configured, this leads to inconsistent information. Changed defaults set for the CR. Mysql is still the default database if none configured
* ~~CR validations taken to reconciliation logic and removed from setting defaults logic~~

eguzki:
all comments answered,

waiting for your input @miguelsorianod

pkg/3scale/amp/operator/system_reconciler.go
miguelsorianod: * Ok to fix storageClassName property location :+1: .
* Ok to remove validation "fix" and leave it "ugly". I'd prefer not mixing reconcile logic and mixing defaults logic. As I commented to me conceptually a reconciler should start having an already validated/consistent CR

About "CR defaults cannot be mysql", several comments:

* If we do this, shouldn't we check then when validating that when highAvailibility is defined to true the mysql field must be null?

* I still have one concern related to my previous comment https://github.com/3scale/3scale-operator/pull/275#issuecomment-554454571. If we are going to remove the explicit setting of any default for the database field in the system section I'd say we should do the same for the fileStorage section. It just seems there's something wrong when we explicitly set defaults for all other "union types" we have but not for this one.

* What happens with existing customers deployments and their fileStorage(if we decide to change that too) and database sections where the defaults are ALREADY set? From the point of view of this changes existing clients might already have an inconsistent view of the CR. Should we have migration code to "unset" them if they are set to defaults? If we decide to not migrate/modify anything for existing deployments could this lead us to some problem in the future?

cc @eguzki to see this feedback

eguzki: > If we do this, shouldn't we check then when validating that when highAvailibility is defined to true the mysql field must be null?

We could do that. Should we? Is there any use case where the customer wants the following scenario?
* redis databases managed externally
* system relational databases be managed internally

If that would be the case, I would express the desired state like this:

```
spec:                                               
  wildcardDomain: apps.example.com                
  highAvailability:                                 
    enabled: true           
  system:
    database:
      postgresql: {}
```
This desired state is not supported currently. That is why that validation you mention should be done. But if the use case above would be valid, the validation you request would not make sense. 

I will implement that validation. 

> I still have one concern related to my previous comment #275 (comment). If we are going to remove the explicit setting of any default for the database field in the system section I'd say we should do the same for the fileStorage section. It just seems there's something wrong when we explicitly set defaults for all other "union types" we have but not for this one.

I think it is a good idea. I will implement it.

> What happens with existing customers deployments and their fileStorage(if we decide to change that too) and database sections where the defaults are ALREADY set? From the point of view of this changes existing clients might already have an inconsistent view of the CR. Should we have migration code to "unset" them if they are set to defaults? If we decide to not migrate/modify anything for existing deployments could this lead us to some problem in the future?

For customers on previous 3scale releases and `.spec.highAvailability.enabled` set to `true`, the CR looks like this:

```
spec:
  ...
  highAvailability:
    enabled: true
  system:
    database:
      mysql: {}
```
I think the CR should be "cleaned" in the upgrade procedure. I will implement it.

eguzki:
all comments answered,

waiting for your input @miguelsorianod

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -29,7 +29,12 @@ type UpgradeApiManager struct {
 }
 
 func (u *UpgradeApiManager) Upgrade() (reconcile.Result, error) {
-	res, err := u.upgradeSystemAppPreHookPodEnv()
+	res, err := u.upgradeApimangerCR()
```
miguelsorianod: typo on the method name

miguelsorianod:
I've renamed the appearances of "upgradeApimanger.*" in the upgrade.go file to "upgradeAPIManager" to fix the typos

PR #280, opened 2019-11-15
Move system smtp to secret

miguelsorianod:
This PR changes the deployment of the SMTP configuration from a ConfigMap to a Secret.
The reason of this is because some of the configuration contains credentials.

### DIFF
--- a/doc/apimanager-reference.md
+++ b/doc/apimanager-reference.md
@@ -326,3 +326,15 @@ existing secret names.
 | --- | --- | --- |
 | AWS_ACCESS_KEY_ID | AWS Access Key ID to use in S3 Storage for System's file storage | N/A |
 | AWS_SECRET_ACCESS_KEY | AWS Access Key Secret to use in S3 Storage for System's file storage | N/A |
+
+#### system-smtp
+
+| **Field** | **Description** | **Default value** |
+| --- | --- | --- |
+| address | Address (hostname or IP) of the remote mail server to use. If set to a value different than `""` System will use the mail server to send mails related to events that happen in the API management solution |  `""` |
+| port | Port of the remote mail server to use | `""` |
+| domain | In case the mail server requires a HELO domain | `""` |
+| authentication | In case the mail server requires authentication, set this setting to the authentication type here. `plain` to send the password in the clear, `login` to send password Base64 encoded, or `cram_md5` to combine a Challenge/Response mechanism based on the HMAC-MD5 algorithm | `""` |
+| username | In case the mail server requires authentication and the authentication type requires it | `""` |
+| password | In case the mail server requires authentication and the authentication type requires it | `""` |
+| openssl.verify.mode | When using TLS, you can set how OpenSSL checks the certificate. This is really useful if you need to validate a self-signed and/or a wildcard certificate. You can use the name of an OpenSSL verify constant: `none` or `peer` | `""` |
--- a/pkg/3scale/amp/auto-generated-templates/amp/amp-eval-s3.yml
+++ b/pkg/3scale/amp/auto-generated-templates/amp/amp-eval-s3.yml
@@ -1551,7 +1551,15 @@ objects:
       threescale_component: system
     name: system
 - apiVersion: v1
-  data:
+  kind: Secret
+  metadata:
+    creationTimestamp: null
+    labels:
+      app: ${APP_LABEL}
+      threescale_component: system
+      threescale_component_element: smtp
+    name: system-smtp
+  stringData:
     address: ""
     authentication: ""
     domain: ""
@@ -1559,14 +1567,6 @@ objects:
     password: ""
     port: ""
     username: ""
-  kind: ConfigMap
-  metadata:
-    creationTimestamp: null
-    labels:
-      app: ${APP_LABEL}
-      threescale_component: system
-      threescale_component_element: smtp
-    name: smtp
 - apiVersion: v1
   data:
     AMP_RELEASE: ${AMP_RELEASE}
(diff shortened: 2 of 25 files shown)


miguelsorianod:
Pending to implement the upgrade procedure but the current PR state should have the needed code in templates and operator to deploy SMTP configuration with a secret instead of a configmap.
@eguzki you can review this part for the moment

miguelsorianod:
Updated PR to include upgrade process to migrate from the 'smtp' ConfigMap to the 'system-smtp' Secret.
In the future the order/coordination in how the upgrade is performed might have to be changed to have consistent changes

eguzki:
looking good

miguelsorianod:
Currently the upgrade procedure should work because there isn't an upgrade step at this moment that requires "atomicity" of the changes.
How do you want to proceed? We can merge this and work on a solution when the need arises/handle it in a separate PR or specify what we should solve in this PR before merging it.

eguzki:
missing smtp secret documentation

miguelsorianod:
Added system-smtp secret documentation.
Ready for re-review @eguzki

miguelsorianod:
I've noticed the APIManager controller unit test related to the upgrade fails. The reason for that is that the upgrade code expects the "system" ConfigMap to exist (see migrateSystemSMTPData method in upgrade.go) but it does not exist anymore due to it has been removed in this PR being replaced by a secret.

Should we make the UpgradeApiManager implement an interface and use an interface instead from the apimanager_controller being the upgrader a field of the ReconcileAPIManager struct? and mock it in the test so the upgrade method returns an arbitrary result? What do you think?

eguzki:
Why not just create the smtp configmap for the upgrade test? you can even test the config map has been replaced by secret.

miguelsorianod:
> Why not just create the smtp configmap for the upgrade test? you can even test the config map has been replaced by secret.

I thought about that but this means that in the next release the test will break/need to be changed so they would be prone to errors. Also the number of calls to the Reconcile method in the test will vary depending on the number of elements to upgrade.

In any case I've implemented it so you can see how the addition of this new functionality has affected the test in the apimanager_controller_test (in a new commit). I've also modified the upgrade.go implementation when creating the system-smtp secret because it was being created without the controllerreference and now it sets the secret data using Data instead of StringData field due to when mocking the API calls the StringData seems that is not automatically converted to Data when creating the object and this caused tests to fail too. You can re-review that file too.

eguzki:
In the next release, all the upgrade tests do not make sense at all and have to be "removed", as it has to be done with the upgrade code.

doc/apimanager-reference.md
```diff
@@ -326,3 +326,15 @@ existing secret names.
 | --- | --- | --- |
 | AWS_ACCESS_KEY_ID | AWS Access Key ID to use in S3 Storage for System's file storage | N/A |
 | AWS_SECRET_ACCESS_KEY | AWS Access Key Secret to use in S3 Storage for System's file storage | N/A |
+
+#### system-smtp
```
eguzki: Recently some user complained about the lack of documentation about SMTP configuration. Do you think it is enough to document a table with available fields in the apimanager reference?

thomasmaas: What are the alternatives?

doc/apimanager-reference.md
```diff
@@ -326,3 +326,15 @@ existing secret names.
...
+
+| **Field** | **Description** | **Default value** |
+| --- | --- | --- |
+| address | Address (hostname or IP) of the remote mail server to use. If set to a value different than `""` System will use the mail server to send mails related to events that happen in the API management solution |  `""` |
+| username | In case the mail server requires authentication, set the username using this setting | `""` |
```
thomasmaas:
```suggestion
| username | In case the mail server requires authentication | `""` |
```

doc/apimanager-reference.md
```diff
@@ -326,3 +326,15 @@ existing secret names.
...
+| password | In case the mail server requires authentication, set the password using this setting | `""` |
```
thomasmaas:
```suggestion
| password | In case the mail server requires authentication. | `""` |
```

doc/apimanager-reference.md
```diff
@@ -326,3 +326,15 @@ existing secret names.
...
+| domain | If a HELO domain is needed to be specified set it using this setting | `""` |
```
thomasmaas:
```suggestion
| domain | In case the mail server requires a HELO domain. | `""` |
```

doc/apimanager-reference.md
```diff
@@ -326,3 +326,15 @@ existing secret names.
...
+| port | Port of the remote mail server to use | `""` |
+| authentication | In case the mail server requires authentication, set this setting to the authentication type here. `plain` to send the password in the clear, `login` to send password Base64 encoded, or `cram_md5` to combine a Challenge/Response mechanism to exchange information and a cryptographic Message Digest 5 algorithm to hash important information | `""` |
```
thomasmaas:
```suggestion
| authentication | Specify the authentication type in case the mail server requires authentication. Use `plain` to send the password in the clear, `login` to send password Base64 encoded, or `cram_md5` to combine a Challenge/Response mechanism to exchange information and a cryptographic Message Digest 5 algorithm to hash important information | `""` |
```

thomasmaas: But "or `cram_md5` to combine a Challenge/Response mechanism to exchange information and a cryptographic Message Digest 5 algorithm to hash important information" is very hard to read, can we simplify?

doc/apimanager-reference.md
```diff
@@ -326,3 +326,15 @@ existing secret names.
...
+| openssl.verify.mode | When using TLS, you can set how OpenSSL checks the certificate. This is really useful if you need to validate a self-signed and/or a wildcard certificate. You can use the name of an OpenSSL verify constant: `none` or `peer` | `""` |
```
thomasmaas:
```suggestion
| openssl.verify.mode | When using TLS, use this field to specify how OpenSSL checks the certificate. This is useful to validate a self-signed and/or a wildcard certificate. Use the name of an OpenSSL verify constant: `none` or `peer` | `""` |
```

thomasmaas: Screenshot 2019-12-18 18 42 19

wouldn't it make sense to re-order and rename the fields to:
- address
- port
- helo.domain
- auth.type
- auth.username
- auth.password
- openssl.verify.mode

doc/apimanager-reference.md
thomasmaas: would you use bother username and password fields when picking `cram_md5`?

doc/apimanager-reference.md
miguelsorianod: Ok to the order.
You cannot rename the fields as they are the names that System looks.

doc/apimanager-reference.md

doc/apimanager-reference.md

doc/apimanager-reference.md

doc/apimanager-reference.md
miguelsorianod: That description is the one from the official rubydocs documentation. Anyway, I've changed it to '`cram_md5` to combine a Challenge/Response mechanism based on the HMAC-MD5 algorithm '

If you don't need to set username and password then you don't have to in that case.
I've updated the username and password fields to
"In case the mail server requires authentication and the authentication type requires it"
instead of just
"In case the mail server requires authentication"

miguelsorianod:
system-smtp documentation table updated

PR #281, opened 2019-11-18
remove s3 secret reconciliation

eguzki:
* remove s3 secret reconciliation. Not required at all, all fields are not reconciliable and required, not optional.
* add validation of CR when s3 is configured


miguelsorianod:
Let's see how Jira issue https://issues.jboss.org/browse/THREESCALE-3868 related to allowing configurability of S3 api-compatible software evolves, as part of the initial proposal was adding new optional fields. Other proposals have been submitted there.

eguzki:
IMO this PR fixes something that is wrong now. Anything to be decided with some future S3 features does not have anything to do with fixing something broken now.

miguelsorianod:
As we talked with the new approach of using a secret mounted as a volume the s3 reconciliation in this code is not performed anymore so the changes in the reconciliation of fields / this secret name does not exist anymore.

I've looked at the code and I don't think this applies anymore so I'd say we close it. Maybe the "getTestSecret" in the ha_test file would be useful for other stuff?

What do you think @eguzki. Shall we close the issue?

PR #283, opened 2019-11-19
remove system database reconciliation on external databases mode

eguzki:


miguelsorianod:
If we want to remove reconciliation of those Secrets the change looks good :+1: . Why we don't want to allow reconciliation of those fields?

pkg/3scale/amp/operator/system_reconciler.go
```diff
@@ -232,31 +232,9 @@ func (r *SystemReconciler) Reconcile() (reconcile.Result, error) {
...
 	}
 
-	// TODO rethink where to create the system-database secret
-	if r.apiManager.Spec.HighAvailability != nil && r.apiManager.Spec.HighAvailability.Enabled {
-		ha, err := r.highAvailability()
-		if err != nil {
-			return reconcile.Result{}, err
-		}
-
-		err = r.reconcileDatabaseHASecret(ha.SystemDatabaseSecret())
```
miguelsorianod: We could remove the reconcileDatabaseHASecret method itself then too

eguzki: Templates need to create the secret with parameters passed to the template. Method cannot be removed.

eguzki:
>  Why we don't want to allow reconciliation of those fields?

* reconciliation of database url is not supported
* There are no optional attributes, so reconciliation is not needed to set defaults.

pkg/3scale/amp/operator/system_reconciler.go
miguelsorianod: But the reconcileDatabaseHASecret is an unexported method of the "operator" package (pkg/3scale/amp/operator), it is not used by the templates.

eguzki: sure, I was referring to `ha.SystemDatabaseSecret()`

eguzki:
ready to review @miguelsorianod

PR #286, opened 2019-11-22
Options provider refactor

eguzki:
Refactor objectives:

* Options providers output can be verified and tested
* Required and optional attributes redefined, as they were wrong
* Delegate validation logic to https://godoc.org/github.com/go-playground/validator Less code to maintain. Options builders are gone. Easier to add validations in a declarative way. Less error prone.
* Defaults refactored to be in a single location, available to template and operator option providers.
* Reading from secrets to fill options encapsulated to be reusable from any option provider.

- [x] Backend option provider
- [x] AMPImages option provider
- [x] Apicast option provider
- [x] Memcached option provider
- [x] System option provider
- [x] System MySQL option provider
- [x] System PostgreSQL option provider
- [x] Redis option provider
- [x] System MySQL Image option provider
- [x] System PostgreSQL Image option provider
- [x] s3
- [x] HA
- [x] Zync option provider
- [x] Add option provider tests


pkg/3scale/amp/component/backend_options.go
```diff
@@ -0,0 +1,98 @@
...
+	StorageURL                   string `validate:"required"`
+	QueuesURL                    string `validate:"required"`
+	StorageSentinelHosts         string
+	StorageSentinelRole          string
+	QueuesSentinelHosts          string
+	QueuesSentinelRole           string
+	ListenerResourceRequirements v1.ResourceRequirements `validate:"-"`
+	WorkerResourceRequirements   v1.ResourceRequirements `validate:"-"`
+	CronResourceRequirements     v1.ResourceRequirements `validate:"-"`
+	ListenerReplicas             int32
```
miguelsorianod: The replicas fields are not marked as "required". This means that validating a BackendOptions where the replicas fields have not been provided would be valid. When replicas field are not provided they use the zero-value. In this case 0, which would deploy 3scale with replicas set to 0. I've seen you set those fields to 1 intentionally from the BackendOptionsProvider in both templates and operator. Is the idea giving the responsibility of creating a "coherent" "default" 3scale deployment to the operator/templates option providers implementations? (for example make sure that a 3scale with 1 replica is deployed when using defaults)

I don't have a clear idea on what's the criteria on what should be "required" and what not. What criteria will we follow to decide whether a field is "required" or not in the validator context?
To try to clarify what it was being done before, I think the criteria I tried to follow previously was that if the value was the same between templates and operator then I considered the option "non-required"(I'm talking of the old concept of required/non-required, nothing to do with the validator in this PR) as a default for it was provided in the component itself.

pkg/3scale/amp/component/backend_options.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/go-playground/validator/v10"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+type BackendOptions struct {
+	ServiceEndpoint              string `validate:"required"`
+	RouteEndpoint                string `validate:"required"`
+	StorageURL                   string `validate:"required"`
+	QueuesURL                    string `validate:"required"`
```
miguelsorianod: Regarding the "required" struct tag field of the Validator:

When having a field as "required" in the new validator context, if I understand correctly, specifying the "required" tag it makes sure that the set value for that option is NOT the zero-value.
How would we handle a required value, that we want to be sure is specified in the struct but at the same time we want to be able to allow the zero-value for it?  Kubernetes tells you to use pointers when using optional values. To distinguish the zero-value from a value not being introduced. Would we use a pointer too in that case?

pkg/3scale/amp/component/backend_options.go
```diff
@@ -0,0 +1,98 @@
...
+		},
+	}
+}
+
+func DefaultSystemBackendUsername() string {
+	return "3scale_api_user"
+}
+
+func DefaultSystemBackendPassword() string {
+	return oprand.String(8)
```
miguelsorianod: Could this potentially be "dangerous" for the stability of an existing deployment?
Each time that function is called would return a different password.
I think it should not be a problem right now because in the Operator options provider code is just used to be set as a default that is just used when the field does not already exist in the secret, but anyway I'd like to comment about this.

pkg/3scale/amp/component/backend_options.go
```diff
@@ -0,0 +1,98 @@
...
+			v1.ResourceMemory: resource.MustParse("80Mi"),
+		},
+		Requests: v1.ResourceList{
+			v1.ResourceCPU:    resource.MustParse("50m"),
+			v1.ResourceMemory: resource.MustParse("40Mi"),
+		},
+	}
+}
+
+func DefaultSystemBackendUsername() string {
```
miguelsorianod: Why there are no defaults defined for the "queuesSentinelRole"? I understand in the template options provider we are not setting it and we are taking advantage of the zero-value of the struct. Why is it different? Why is not a default* method in the backend component?

pkg/3scale/amp/component/backend_options.go
```diff
@@ -0,0 +1,98 @@
...
+	RouteEndpoint                string `validate:"required"`
+	StorageURL                   string `validate:"required"`
+	QueuesURL                    string `validate:"required"`
+	StorageSentinelHosts         string
+	StorageSentinelRole          string
+	QueuesSentinelHosts          string
+	QueuesSentinelRole           string
+	ListenerResourceRequirements v1.ResourceRequirements `validate:"-"`
+	WorkerResourceRequirements   v1.ResourceRequirements `validate:"-"`
+	CronResourceRequirements     v1.ResourceRequirements `validate:"-"`
```
miguelsorianod: Why are this fields not required? I guess this is related to the question I made about what's the general criteria of considering some option required/non-required and what does that means for us and the validator and where the responsabilities lie

pkg/3scale/amp/operator/backend_options_provider.go
```diff
@@ -0,0 +1,144 @@
...
+
+	err := o.setSecretBasedOptions()
+	if err != nil {
+		return nil, err
+	}
+
+	o.setResourceRequirementsOptions()
+	o.setReplicas()
+
+	err = o.backendOptions.Validate()
```
miguelsorianod: * Might it happen that if we modify an existing "valid" BackendOptions the struct becomes an "invalid" struct from the validator point of view?

pkg/3scale/amp/operator/backend_options_provider.go
```diff
@@ -0,0 +1,144 @@
...
+	return o.backendOptions, err
+}
+
+func (o *OperatorBackendOptionsProvider) setSecretBasedOptions() error {
```
miguelsorianod: Do you think it might be a good idea to modularize the setSecretBasedOptions method? Like separating the generation of cases into different methods so we group related sets of options to make them easily localizable too. For example have a method for the options related to system-backend secret, other for the backend-api secret etc...

It seems large and a little bit hard to find compared with when we had functions

pkg/3scale/amp/operator/backend_options_provider.go
```diff
@@ -0,0 +1,144 @@
...
+		},
+		{
+			&o.backendOptions.QueuesSentinelRole,
+			component.BackendSecretBackendRedisSecretName,
+			component.BackendSecretBackendRedisQueuesSentinelRoleFieldName,
+			"",
+		},
+	}
+
+	for _, option := range cases {
```
miguelsorianod: In setSecretBasedOptions how expected mandatory fields are going to be managed?
The code only seems to manage fields that have defaults.
I understand this is still pending implementation right and has not been done because it's not been the case here yet that we have an option in that situation.

pkg/3scale/amp/operator/backend_options_provider.go
```diff
@@ -0,0 +1,144 @@
...
+		val, err := o.secretSource.FieldValue(option.secretName, option.secretField, option.defValue)
```
miguelsorianod: In setSecretBasedOptions, for a secret that has N fields that secret is "getted" N times.
Before it was only one time. It might be good enough as I think operator-sdk does some sort of caching but if the operator-sdk that the caching has does not work as we expect this means that it makes lot more API calls than before.

pkg/3scale/amp/operator/option_provider_utils.go
```diff
@@ -0,0 +1,61 @@
...
+
+func (s *SecretSource) RequiredFieldValue(secretName, fieldName string) (*string, error) {
+	return s.fieldReader(secretName, fieldName, false, true, "")
+}
+
+func (s *SecretSource) RequiredFieldValueFromRequiredSecret(secretName, fieldName string) (*string, error) {
+	return s.fieldReader(secretName, fieldName, true, true, "")
+}
+
+func (s *SecretSource) fieldReader(secretName, fieldName string, secretRequired, fieldRequired bool, def string) (*string, error) {
```
miguelsorianod: We take advantage in the fieldReader method of the fact that a non-existing secret behaves as an existing one.
Thinking about it, with how setSecretBasedOptions is implemented right now, it could happen that the secret might not exist at all and we would not get any errors when providing the options. Could this behavior cause us any problem?

pkg/3scale/amp/component/backend_options.go
```diff
@@ -0,0 +1,98 @@
+package component
+
+import (
+	oprand "github.com/3scale/3scale-operator/pkg/crypto/rand"
+	"github.com/go-playground/validator/v10"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+type BackendOptions struct {
```
miguelsorianod: What validation and what defaults would we set for fields in a component that were previously pointers because it allows us to identify not being it set vs being it set with its zero-value?
See for example in the System component the fields *S3FileStorageOptions and *PVCFileStorageOptions. Only one should be set and we need to distinguish between not having provided the option vs having it provided it with the default value of the value it points to.

miguelsorianod:
I like the approach :+1: . I left some comments so we have an idea on how to proceed/clarify some doubts I have with certain things / what implications they have.

pkg/3scale/amp/operator/backend_options_provider.go
```diff
@@ -0,0 +1,144 @@
...
+	err = o.backendOptions.Validate()
+	return o.backendOptions, err
+}
+
+func (o *OperatorBackendOptionsProvider) setSecretBasedOptions() error {
+	cases := []struct {
+		field       *string
+		secretName  string
+		secretField string
+		defValue    string
```
miguelsorianod: What if the field has no default value because it is required? setting it as "" might not be a solution because there are some fields where the default value is really a "" and would not mean, 'not having a default' (distinguish vs default value and value not provided)

pkg/3scale/amp/component/backend_options.go
eguzki: The owner of the default value should be the `operator` package (or template package). It is high level information. That could be different for the operator controller and the template adapter. That does not mean that the value could not be defined in `component` package to be used by template adapters or operator controller.

Be careful with the concept of "required". Regarding the backend options, all fields are required for backend option user. For example [here](https://github.com/3scale/3scale-operator/blob/ce0b0a87de1253c478f17f70ec0f288ecc72056d/pkg/3scale/amp/component/backend.go#L252):

`Resources: *backend.Options.listenerResourceRequirements`

`listenerResourceRequirements` even if it is a pointer, it is required by the `component` package, as it is being dereferenced. Otherwise you would get Invalid memory address because of nil pointer dereference. It is correct that the value in the CRD is a pointer, because that way you differentiate between not being set and zero value. But for options being used in component, it is different.

For the options sent to the component package for building objects, generally all attributes should be required, otherwise makes no sense to pass it. Generally, there may be edge cases or more complex cases like in system component which is more complex.

Thus, required for the validator means it cannot have zero value. Back to the listener `replica`, we allow it to be 0, because that is the value set by the customer in the CR. Thus, we cannot enforce it to be not zero.

pkg/3scale/amp/component/backend_options.go
eguzki: see comment above

pkg/3scale/amp/component/backend_options.go
eguzki: I agree, can be changed

pkg/3scale/amp/component/backend_options.go
eguzki: Yep, there could be amethod like this:
```
func DefaultBackendSecretBackendRedisStorageSentinelRole() string {
    return ""
}
```

Then, in backend_options_provider, 

instead of 

```
  {                                                                         
      &o.backendOptions.QueuesSentinelHosts,                                
      component.BackendSecretBackendRedisSecretName,                        
      component.BackendSecretBackendRedisQueuesSentinelHostsFieldName,      
      "",                                                                   
  },                                                                        
```

it would be

```
  {                                                                         
      &o.backendOptions.QueuesSentinelHosts,                                
      component.BackendSecretBackendRedisSecretName,                        
      component.BackendSecretBackendRedisQueuesSentinelHostsFieldName,      
      component.DefaultBackendSecretBackendRedisStorageSentinelRole(),                                                                   
  },                                                                        
```

I agree it is more robust than relying on zero value of string

pkg/3scale/amp/component/backend_options.go
eguzki: Yes, above comments my answer. 

Let me add that `validate:"-"` means skip validation and zero value is a valid value.

pkg/3scale/amp/operator/backend_options_provider.go
eguzki: Yes

With this implementation, options returned by the OptionProvider are enforced to be valid. What the receiver does with those options does not concern to the OptionProvider.

pkg/3scale/amp/operator/backend_options_provider.go
eguzki: Yes, can be an improvement.

pkg/3scale/amp/operator/backend_options_provider.go
eguzki: `SecretSource` struct has the following public methods available:

```
func (s *SecretSource) FieldValue(secretName, fieldName string, def string) (*string, error)
func (s *SecretSource) FieldValueFromRequiredSecret(secretName, fieldName string, def string) (*string, error)
func (s *SecretSource) RequiredFieldValue(secretName, fieldName string) (*string, error)
func (s *SecretSource) RequiredFieldValueFromRequiredSecret(secretName, fieldName string) (*string, error)
```

I think all cases are covered. For the backend options specific case, the secret existence is not mandatory and any field existence is mandatory. Thus `FieldValue` method is being used.

pkg/3scale/amp/operator/backend_options_provider.go
eguzki: `SecretSource` struct should implement caching. There is a `// TODO implement caching?? ` comment there to start discussion. I think is should implement caching. Very simple implementation as is it assumed that a new `SecretSource` object will be instantiated for each option provider and for each reconciliation loop. So no need to implement size limit, expiration or any eviction action.

pkg/3scale/amp/operator/option_provider_utils.go
eguzki: For the backend options provider specific case, the `setSecretBasedOptions` does not require the existence of the secret nor the existence of any of the fields, so using `FieldValue` is appropriate. Other option providers might have different requirements and should be using `FieldValueFromRequiredSecret`, `RequiredFieldValue` or `RequiredFieldValueFromRequiredSecret` methods

pkg/3scale/amp/operator/backend_options_provider.go
eguzki: If options received by the OptionProvider are modified for any reasons, you can always call `Validate()` method to check if they are still valid. `Validate()` method belongs to the options struct, not to the OptionProvider.

pkg/3scale/amp/component/backend_options.go
eguzki: [Validator library](https://godoc.org/github.com/go-playground/validator) is powerful enough for these cases. I made sure it could handle this kind of situations. 

There are struct level validations to validate that either `S3FileStorageOptions` or `PVCFileStorageOptions` fields have been set. See this [example](https://github.com/go-playground/validator/blob/master/_examples/struct-level/main.go#L99)

It also allows conditional validation. So nested struct will be validated when it exists and validation will be omitted when it is nil. For instance:

```
type Address struct {                       
    Street string `validate:"required"`     
    City   string `validate:"required"`   
} 
type User struct { 
    Address        *Address `validate:"omitempty"`
}
```

I think it will be enough to have good validation for the `S3FileStorageOptions` and `PVCFileStorageOptions` fields

pkg/3scale/amp/operator/backend_options_provider.go
eguzki: This is implementation for the `backend_options_provider`. Does not apply. As explained earlier,  other option providers may require existence of a fields, so default values would not make sense at all and other public methods of `SecretSource` will be needed, not having any argument for the default value.

eguzki:
All comments have been answered.

waiting for your input @miguelsorianod

pkg/3scale/amp/operator/backend_options_provider.go
miguelsorianod: :+1: . I thought you were trying to make the setSecretBasedOptions valid for all scenarios but for what I've understood for other comments that is specific code for the backend_options_provider and thje structure of the setSecretBasedOptions for other components might change a lot (using different structs, different methods of the SecretSource class, ...) :+1:

pkg/3scale/amp/operator/backend_options_provider.go
miguelsorianod: :+1:

pkg/3scale/amp/component/backend_options.go
miguelsorianod: :+1: It might be better being consistent with what we do

pkg/3scale/amp/component/backend_options.go
miguelsorianod: To what would you change it?
From what I understand from other comments, the responsibility of instantiating the component itself(not the options) with the correct values for the options (being consistent between them and no erroneus default values) is of the options providers.
That's why I wrote that it should not be a problem right now because it's just used as the default, but at the same time is somewhat risky if we have a bug in the options providers code

pkg/3scale/amp/operator/backend_options_provider.go
miguelsorianod: I think we should implement it too or change it to get the secret at the beginning for all fields.

Leaving it as it is right now I think it might be dangerous. Keep in mind that right now for each get we might receive different resourceVersions of the object so in the same method we could receive "inconsistent" states of the fields among them

pkg/3scale/amp/component/backend_options.go
miguelsorianod: Ok, if I understand correctly they will still be pointers. They could not be required though because the zero-value of a pointer is nil

pkg/3scale/amp/component/backend_options.go
miguelsorianod: :+1:

miguelsorianod:
@eguzki I left some feedback based on your answers.
I think more work can be done for the other components as the doubts I had have been answered,  but I think we need to decide on what to do with the SecretSource and caching/possible inconsistencies that might make the data we get inconsistent as it is dangerous for the deployments.

pkg/3scale/amp/component/backend_options.go
eguzki: The behavior did not change respect to the current code base. On each reconciliation loop if a certain secret does not have some optional field, a entirely new default value will be generated. It should not be an issue because it is expected to be setting defaults only once.

So, how can be improved? having in mind that the operator only manages one 3scale instance, we could use process level (golang level) constants initialized at bootstrap time. This solutions would not work if the operator could manage several 3scale instances.

pkg/3scale/amp/operator/backend_options_provider.go
eguzki: Ok, I will implement a simple caching solutions. Caching life-cycle will be limited to the reconciliation loop and each options provider will have its own caching instance.

pkg/3scale/amp/component/backend_options.go
miguelsorianod: :+1: yes that's why I said it should not be a problem right now because it's just used as the default, same as before. Both ways are risky.
I don't think it is a priority to change this compared to other things. I leave it up to you. To me changin it is not really a must at this moment.

pkg/3scale/amp/component/backend_options.go
eguzki: Yes, optional fields, optional for component package, not for the CR, will be pointers.

pkg/3scale/amp/operator/backend_options_provider.go
miguelsorianod: :+1: I think it makes sense at this moment

pkg/3scale/amp/operator/backend_options_provider.go
eguzki: tried to simplify but result is not good enough. Improvement for future PR's

eguzki:
secret source caching implemented

eguzki:

pkg/3scale/amp/template/adapters/backend.go
```diff
@@ -34,11 +36,22 @@ func (b *Backend) Objects() ([]common.KubernetesObject, error) {
...
-	bob.TenantName("${TENANT_NAME}")
-	bob.WildcardDomain("${WILDCARD_DOMAIN}")
-	return bob.Build()
+	bo := component.NewBackendOptions()
+	bo.AppLabel = "${APP_LABEL}"
+	bo.SystemBackendUsername = "${SYSTEM_BACKEND_USERNAME}"
+	bo.SystemBackendPassword = "${SYSTEM_BACKEND_PASSWORD}"
+	bo.TenantName = "${TENANT_NAME}"
+	bo.WildcardDomain = "${WILDCARD_DOMAIN}"
+	bo.RouteEndpoint = fmt.Sprintf("https://backend-%s.%s", "${TENANT_NAME}", "${WILDCARD_DOMAIN}")
```
miguelsorianod: See related review comment in the options backend options provider about making this a default function

pkg/3scale/amp/template/adapters/system-mysql.go
```diff
@@ -58,12 +58,16 @@ func (m *SystemMysqlAdapter) Objects() ([]common.KubernetesObject, error) {
...
-	mob.RootPassword("${SYSTEM_DATABASE_ROOT_PASSWORD}")
-	mob.DatabaseURL("mysql2://root:" + "${SYSTEM_DATABASE_ROOT_PASSWORD}" + "@system-mysql/" + "${SYSTEM_DATABASE}")
-	return mob.Build()
+	mo := component.NewSystemMysqlOptions()
+	mo.AppLabel = "${APP_LABEL}"
+	mo.DatabaseName = "${SYSTEM_DATABASE}"
+	mo.User = "${SYSTEM_DATABASE_USER}"
+	mo.Password = "${SYSTEM_DATABASE_PASSWORD}"
+	mo.RootPassword = "${SYSTEM_DATABASE_ROOT_PASSWORD}"
+	mo.DatabaseURL = "mysql2://root:" + "${SYSTEM_DATABASE_ROOT_PASSWORD}" + "@system-mysql/" + "${SYSTEM_DATABASE}"
```
miguelsorianod: Can the DefaultSystemMysqlDatabaseURL method be reused or would it not be equivalent?

pkg/3scale/amp/template/adapters/system-postgresql.go
```diff
@@ -50,11 +50,15 @@ func (r *SystemPostgreSQLAdapter) Objects() ([]common.KubernetesObject, error) {
...
-	mob.User("${SYSTEM_DATABASE_USER}")
-	mob.Password("${SYSTEM_DATABASE_PASSWORD}")
-	mob.DatabaseURL("postgresql://${SYSTEM_DATABASE_USER}:" + "${SYSTEM_DATABASE_PASSWORD}" + "@system-postgresql/" + "${SYSTEM_DATABASE}")
-	return mob.Build()
+	o := component.NewSystemPostgreSQLOptions()
+	o.AppLabel = "${APP_LABEL}"
+	o.DatabaseName = "${SYSTEM_DATABASE}"
+	o.User = "${SYSTEM_DATABASE_USER}"
+	o.Password = "${SYSTEM_DATABASE_PASSWORD}"
+	o.DatabaseURL = "postgresql://${SYSTEM_DATABASE_USER}:" + "${SYSTEM_DATABASE_PASSWORD}" + "@system-postgresql/" + "${SYSTEM_DATABASE}"
```
miguelsorianod: Can the DefaultSystemPostgresqlDatabaseURL method be used or would it not be equivalent?

pkg/3scale/amp/component/system_options.go
```diff
@@ -0,0 +1,293 @@
...
+	RedisNamespace                         *string `validate:"required"`
+	MessageBusRedisURL                     *string `validate:"required"`
+	MessageBusRedisSentinelHosts           *string `validate:"required"`
+	MessageBusRedisSentinelRole            *string `validate:"required"`
+	MessageBusRedisNamespace               *string `validate:"required"`
+	ApicastSystemMasterProxyConfigEndpoint string  `validate:"required"`
+	ApicastSystemMasterBaseURL             string  `validate:"required"`
+	AdminEmail                             *string `validate:"required"`
+
+	AppMasterContainerResourceRequirements    *v1.ResourceRequirements `validate:"required"`
```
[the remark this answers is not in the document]
eguzki: Sure, I will review all resource requirements.

pkg/3scale/amp/component/ampimages_options.go
```diff
@@ -0,0 +1,27 @@
...
+type AmpImagesOptions struct {
+	AppLabel                    string `validate:"required"`
+	AmpRelease                  string `validate:"required"`
+	ApicastImage                string `validate:"required"`
+	BackendImage                string `validate:"required"`
+	SystemImage                 string `validate:"required"`
+	ZyncImage                   string `validate:"required"`
+	ZyncDatabasePostgreSQLImage string `validate:"required"`
+	SystemMemcachedImage        string `validate:"required"`
+	InsecureImportPolicy        bool
```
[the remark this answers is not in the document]
eguzki: Sure, I will review all flags for import policy

pkg/3scale/amp/operator/backend_options_provider.go
```diff
@@ -0,0 +1,144 @@
...
+			&o.backendOptions.ServiceEndpoint,
+			component.BackendSecretBackendListenerSecretName,
+			component.BackendSecretBackendListenerServiceEndpointFieldName,
+			component.DefaultBackendServiceEndpoint(),
+		},
+		{
+			&o.backendOptions.RouteEndpoint,
+			component.BackendSecretBackendListenerSecretName,
+			component.BackendSecretBackendListenerRouteEndpointFieldName,
+			fmt.Sprintf("https://backend-%s.%s", *o.apimanager.Spec.TenantName, o.apimanager.Spec.WildcardDomain),
```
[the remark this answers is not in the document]
eguzki: :+1:

pkg/3scale/amp/component/system.go
```diff
@@ -414,19 +414,19 @@ func (system *System) RedisSecret() *v1.Secret {
...
-			SystemSecretSystemRedisURLFieldName:                *system.Options.redisURL,
-			SystemSecretSystemRedisSentinelHosts:               *system.Options.redisSentinelHosts,
-			SystemSecretSystemRedisSentinelRole:                *system.Options.redisSentinelRole,
-			SystemSecretSystemRedisMessageBusRedisURLFieldName: *system.Options.messageBusRedisURL,
-			SystemSecretSystemRedisMessageBusSentinelHosts:     *system.Options.messageBusRedisSentinelHosts,
-			SystemSecretSystemRedisMessageBusSentinelRole:      *system.Options.messageBusRedisSentinelRole,
-			SystemSecretSystemRedisNamespace:                   *system.Options.redisNamespace,
-			SystemSecretSystemRedisMessageBusRedisNamespace:    *system.Options.messageBusRedisNamespace,
+			SystemSecretSystemRedisURLFieldName:                system.Options.RedisURL,
+			SystemSecretSystemRedisSentinelHosts:               *system.Options.RedisSentinelHosts,
```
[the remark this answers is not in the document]
eguzki: :+1:

pkg/3scale/amp/component/system.go
```diff
@@ -461,20 +461,20 @@ func (system *System) SeedSecret() *v1.Secret {
...
-			SystemSecretSystemSeedAdminEmailFieldName:        *system.Options.adminEmail,
-			SystemSecretSystemSeedTenantNameFieldName:        system.Options.tenantName,
+			SystemSecretSystemSeedMasterDomainFieldName:      system.Options.MasterName,
+			SystemSecretSystemSeedMasterAccessTokenFieldName: system.Options.MasterAccessToken,
+			SystemSecretSystemSeedMasterUserFieldName:        system.Options.MasterUsername,
+			SystemSecretSystemSeedMasterPasswordFieldName:    system.Options.MasterPassword,
+			SystemSecretSystemSeedAdminAccessTokenFieldName:  system.Options.AdminAccessToken,
+			SystemSecretSystemSeedAdminUserFieldName:         system.Options.AdminUsername,
+			SystemSecretSystemSeedAdminPasswordFieldName:     system.Options.AdminPassword,
+			SystemSecretSystemSeedAdminEmailFieldName:        *system.Options.AdminEmail,
```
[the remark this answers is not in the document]
eguzki: AdminEmail allows empty string. 

So to verify that some value has been set, it is a pointer. Then the validator checks that value has been set. Any value is valid, even empty string. The alternative approach is to have it as value type and remove validation.  Generally, to be consistent, the first approach has been implemented. The alternative approach is also valid IMO.

pkg/3scale/amp/operator/highavailability_options_provider.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return nil, err
+	}
+	err = h.setSystemDatabaseOptions()
+	if err != nil {
+		return nil, err
+	}
+
+	// not required for operator, but required for templates
+	h.options.AppLabel = "-"
```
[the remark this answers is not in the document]
eguzki: Ok,

This irrelevant value setting code is a clear warning note that the code does not belong to the operator package, but to the templates package.

pkg/3scale/amp/operator/system_postgresql_options_provider.go
```diff
@@ -0,0 +1,122 @@
...
+	s.options.Password = val
+
+	val, err = s.secretSource.FieldValue(
+		component.SystemSecretSystemDatabaseSecretName,
+		component.SystemSecretSystemDatabaseURLFieldName,
+		component.DefaultSystemPostgresqlDatabaseURL(s.options.User, s.options.Password, component.DefaultSystemPostgresqlDatabaseName()))
+	if err != nil {
+		return err
+	}
+	s.options.DatabaseURL = val
```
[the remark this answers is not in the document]
eguzki: :+1:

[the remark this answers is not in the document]
eguzki: not really, because the default value for the `DatabaseURL` field is dependent on `User` and `Password` fields

pkg/3scale/amp/template/adapters/backend.go
eguzki: done

pkg/3scale/amp/template/adapters/system-mysql.go
eguzki: :+1:

pkg/3scale/amp/template/adapters/system-postgresql.go
eguzki: :+1:

pkg/3scale/amp/component/system_options.go
```diff
@@ -0,0 +1,293 @@
...
+	AppDeveloperContainerResourceRequirements *v1.ResourceRequirements `validate:"required"`
+	SidekiqContainerResourceRequirements      *v1.ResourceRequirements `validate:"required"`
+	SphinxContainerResourceRequirements       *v1.ResourceRequirements `validate:"required"`
+
+	S3FileStorageOptions  *S3FileStorageOptions  `validate:"required_without=PvcFileStorageOptions"`
+	PvcFileStorageOptions *PVCFileStorageOptions `validate:"required_without=S3FileStorageOptions"`
+
+	AppReplicas     *int32 `validate:"required"`
+	SidekiqReplicas *int32 `validate:"required"`
+
```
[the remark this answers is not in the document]
eguzki: Made consistent

pkg/3scale/amp/component/system_options.go
```diff
@@ -0,0 +1,293 @@
...
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+type S3FileStorageOptions struct {
+	ConfigurationSecretName string `validate:"required"`
+}
+
+type SystemSMTPSecretOptions struct {
+	Address           *string `validate:"required"`
```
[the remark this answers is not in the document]
eguzki: As commented above, there are two approached. 

Implemented approach for other fields where zero value is valid is to check value has been set, even if it is empty string. 

Nothing prevent us from implementing with the alternative approach you are suggesting. But I do not see a good reason to have different approaches. There could be uses cases where it is justified. This case does not seem so IMO.

pkg/3scale/amp/component/highavailability_options.go
```diff
@@ -0,0 +1,29 @@
+package component
+
+import "github.com/go-playground/validator/v10"
+
+type HighAvailabilityOptions struct {
+	AppLabel                            string `validate:"required"`
+	BackendRedisQueuesEndpoint          string `validate:"required"`
+	BackendRedisQueuesSentinelHosts     string `validate:"required"`
```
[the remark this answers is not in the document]
eguzki: The "arbitrary settings" are the symptoms of a bad smell, not the cause. 

IMO The fields of the options struct should be checked against not being set or whatever validation are needed. Removing validation because for templates they are not needed does not seems to me a good way to go. 

We should refactor the templates usage of HighAvailability option and you can have the best of both worlds: validation and not having to set "arbitrary" values. wdyt?

eguzki:
All feedback answered, @miguelsorianod

pkg/3scale/amp/operator/system_postgresql_options_provider.go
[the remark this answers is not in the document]
miguelsorianod: :+1:

pkg/3scale/amp/component/system.go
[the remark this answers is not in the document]
miguelsorianod: Understood :+1:

pkg/3scale/amp/component/system_options.go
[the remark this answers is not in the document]
miguelsorianod: :+1:

pkg/3scale/amp/operator/highavailability_options_provider.go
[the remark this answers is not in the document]
miguelsorianod: :+1:

pkg/helper/secretutils.go
```diff
@@ -74,3 +77,89 @@ func MergeSecretData(from, to map[string][]byte) map[string][]byte {
...
+
+func (s *SecretSource) FieldValue(secretName, fieldName string, def string) (string, error) {
+	return s.fieldReader(secretName, fieldName, false, false, def)
+}
+
+func (s *SecretSource) FieldValueFromRequiredSecret(secretName, fieldName string, def string) (string, error) {
+	return s.fieldReader(secretName, fieldName, true, false, def)
+}
+
+func (s *SecretSource) RequiredFieldValue(secretName, fieldName string) (string, error) {
```
miguelsorianod: Does this method has any use case?
In case a field of a secret is required, shouldn't the secret be required too? If a field is required (there's no default value) and the secret does not exist wouldn't we always get an error?

pkg/3scale/amp/component/system_options.go
```diff
@@ -0,0 +1,293 @@
...
+
+func DefaultSystemMessageBusRedisSentinelRole() string {
+	return ""
+}
+
+func DefaultSystemRedisNamespace() string {
+	return ""
+}
+
+func DefaultSystemMessgeBusRedisNamespace() string {
```
miguelsorianod: Typo on this function name. It should be `MessageBus` instead of `MessgeBus`

pkg/helper/secretutils.go
eguzki: Yes, 

The field requirement implies secret existence.

I will remove the method.

PR #287, opened 2019-11-28
Update community operator 0.4.0 CSV fields

miguelsorianod:


miguelsorianod:
Ready for review @eguzki . I've applied those changes and some more that I've detected doing a diff.
There's also the decision on whether we should set maturity to alpha (in prod is set like that) or stable (right now in this branch)

deploy/olm-catalog/3scale-operator/3scale-community-operator.v0.4.0.clusterserviceversion.yaml
```diff
@@ -164,6 +163,7 @@ spec:
                 - name: ZYNC_IMAGE
                   value: "quay.io/3scale/3scale27:zync-3scale-2.7.0-GA"
                 image: quay.io/3scale/3scale-operator:v0.4.0
+                imagePullPolicy: Always
```
eguzki: why? 

This tag is not floating tag and default pull policy `IfNotPresent` is good enough

miguelsorianod: This was to match what's in the 2.7-stable-prod branch

eguzki: If the tag is not floating, I would go for default pull policy

miguelsorianod: :+1: done

miguelsorianod:
>  There's also the decision on whether we should set maturity to alpha (in prod is set like that) or stable (right now in this branch).

I set it to stable and I'll update the master branch to have it in stable too.
I'll also remove imagePullPolicy:always from master branch

PR #290, opened 2019-12-03
metrics endpoint include 3scale operator and product version

eguzki:
Update the metrics endpoint in the operator to include operator version and 3scale product version:
```
threescale_version_info{operator_version="0.4.0",version="2.7"} 0
```


miguelsorianod:
For what I've read using the global registry is not very recommendable and they recommend creating your own registry over that because the global registry is global state and for some other reasons:
https://github.com/operator-framework/operator-sdk/issues/1426

At the same time operator-sdk though it seemed/seems the operator-sdk used/uses the global registry too so I think what we have now works good enough :+1: . However if you want to take time to look into that up to you.

PR #292, opened 2019-12-04
Change imagestream tagging structure

miguelsorianod:
Currently DeploymentConfigs use "latest" ImageStream tag as the image change trigger.
We update that ImageStream tag to point to every 3scale new release version. This means that when upgrading 3scale version, upgrading the ImageStream tag latest pointing to a new 3scale version the DeploymentConfigs that use that tag are automatically updated.

This makes performing several changes to a DeploymentConfig together with an image change in a single update not possible, due to  the change to the DC is triggered by a change in the ImageStream.
This PR removes the usage of the "latest" tag from the DCs and uses version-specific tag in the DeploymentConfig image change triggers. In this way we can control the moment of the image change at will at DC level

Implement for:

- [x] APIcast
- [x] Backend
- [x] System
- [x] System MySQL
- [x] System PostgreSQL
- [x] System Memcached
- [x] System Redis
- [x] Backend Redis
- [x] Zync
- [x] Zync PostgreSQL

Also:
- [x] Implement the upgrade procedure for all the previous changes. EDIT: Discarded unit testing of it as it involves too much setup (creating stub objects for all existing DCs, all existing ImageStreams, and all needed secrets to be able to instantiate the component objects that use them when creating the Option objects) for a content that will be deprecated on next release or even in future PRs.

### DIFF
--- a/pkg/3scale/amp/auto-generated-templates/amp/amp-eval.yml
+++ b/pkg/3scale/amp/auto-generated-templates/amp/amp-eval.yml
@@ -26,16 +26,6 @@ objects:
       local: false
     tags:
     - annotations:
-        openshift.io/display-name: amp-backend (latest)
-      from:
-        kind: ImageStreamTag
-        name: ${AMP_RELEASE}
-      generation: null
-      importPolicy: {}
-      name: latest
-      referencePolicy:
-        type: ""
-    - annotations:
         openshift.io/display-name: amp-backend ${AMP_RELEASE}
       from:
         kind: DockerImage
@@ -62,16 +52,6 @@ objects:
     lookupPolicy:
       local: false
     tags:
-    - annotations:
-        openshift.io/display-name: AMP Zync (latest)
-      from:
-        kind: ImageStreamTag
-        name: ${AMP_RELEASE}
-      generation: null
-      importPolicy: {}
-      name: latest
-      referencePolicy:
-        type: ""
     - annotations:
         openshift.io/display-name: AMP Zync ${AMP_RELEASE}
       from:
@@ -100,16 +80,6 @@ objects:
       local: false
     tags:
     - annotations:
-        openshift.io/display-name: AMP APIcast (latest)
-      from:
-        kind: ImageStreamTag
-        name: ${AMP_RELEASE}
-      generation: null
-      importPolicy: {}
-      name: latest
-      referencePolicy:
-        type: ""
-    - annotations:
         openshift.io/display-name: AMP APIcast ${AMP_RELEASE}
       from:
         kind: DockerImage
@@ -136,16 +106,6 @@ objects:
     lookupPolicy:
       local: false
     tags:
-    - annotations:
-        openshift.io/display-name: AMP System (latest)
-      from:
-        kind: ImageStreamTag
-        name: ${AMP_RELEASE}
-      generation: null
-      importPolicy: {}
-      name: latest
-      referencePolicy:
-        type: ""
     - annotations:
         openshift.io/display-name: AMP system ${AMP_RELEASE}
       from:
@@ -174,16 +134,6 @@ objects:
       local: false
     tags:
     - annotations:
-        openshift.io/display-name: Zync PostgreSQL (latest)
-      from:
-        kind: ImageStreamTag
-        name: ${AMP_RELEASE}
-      generation: null
-      importPolicy: {}
-      name: latest
-      referencePolicy:
-        type: ""
-    - annotations:
         openshift.io/display-name: Zync ${AMP_RELEASE} PostgreSQL
       from:
         kind: DockerImage
@@ -210,16 +160,6 @@ objects:
     lookupPolicy:
       local: false
     tags:
-    - annotations:
-        openshift.io/display-name: System Memcached (latest)
-      from:
-        kind: ImageStreamTag
-        name: ${AMP_RELEASE}
-      generation: null
-      importPolicy: {}
-      name: latest
-      referencePolicy:
-        type: ""
     - annotations:
         openshift.io/display-name: System ${AMP_RELEASE} Memcached
       from:
@@ -254,16 +194,6 @@ objects:
     lookupPolicy:
       local: false
     tags:
-    - annotations:
-        openshift.io/display-name: System MySQL (latest)
-      from:
-        kind: ImageStreamTag
-        name: ${AMP_RELEASE}
-      generation: null
-      importPolicy: {}
-      name: latest
-      referencePolicy:
-        type: ""
     - annotations:
         openshift.io/display-name: System ${AMP_RELEASE} MySQL
       from:
@@ -353,7 +283,7 @@ objects:
         - backend-redis
         from:
           kind: ImageStreamTag
-          name: backend-redis:latest
+          name: backend-redis:${AMP_RELEASE}
       type: ImageChange
   status:
     availableReplicas: 0
@@ -467,16 +397,6 @@ objects:
       local: false
     tags:
     - annotations:
-        openshift.io/display-name: Backend Redis (latest)
-      from:
-        kind: ImageStreamTag
-        name: ${AMP_RELEASE}
-      generation: null
-      importPolicy: {}
-      name: latest
-      referencePolicy:
-        type: ""
-    - annotations:
         openshift.io/display-name: Backend ${AMP_RELEASE} Redis
       from:
         kind: DockerImage
@@ -566,7 +486,7 @@ objects:
         - system-redis
         from:
           kind: ImageStreamTag
-          name: system-redis:latest
+          name: system-redis:${AMP_RELEASE}
       type: ImageChange
   status:
     availableReplicas: 0
@@ -624,16 +544,6 @@ objects:
     lookupPolicy:
       local: false
     tags:
-    - annotations:
-        openshift.io/display-name: System Redis (latest)
-      from:
-        kind: ImageStreamTag
-        name: ${AMP_RELEASE}
-      generation: null
-      importPolicy: {}
-      name: latest
-      referencePolicy:
-        type: ""
     - annotations:
         openshift.io/display-name: System ${AMP_RELEASE} Redis
       from:
@@ -780,7 +690,7 @@ objects:
         - backend-cron
         from:
           kind: ImageStreamTag
-          name: amp-backend:latest
+          name: amp-backend:${AMP_RELEASE}
       type: ImageChange
   status:
     availableReplicas: 0
@@ -906,7 +816,7 @@ objects:
         - backend-listener
         from:
           kind: ImageStreamTag
-          name: amp-backend:latest
+          name: amp-backend:${AMP_RELEASE}
       type: ImageChange
   status:
     availableReplicas: 0
@@ -1099,7 +1009,7 @@ objects:
         - backend-worker
         from:
           kind: ImageStreamTag
-          name: amp-backend:latest
+          name: amp-backend:${AMP_RELEASE}
       type: ImageChange
   status:
     availableReplicas: 0
@@ -1253,7 +1163,7 @@ objects:
         - system-mysql
         from:
           kind: ImageStreamTag
-          name: system-mysql:latest
+          name: system-mysql:${AMP_RELEASE}
       type: ImageChange
   status:
     availableReplicas: 0
@@ -1407,7 +1317,7 @@ objects:
         - memcache
         from:
           kind: ImageStreamTag
-          name: system-memcached:latest
+          name: system-memcached:${AMP_RELEASE}
       type: ImageChange
   status:
     availableReplicas: 0
@@ -2807,7 +2717,7 @@ objects:
         - system-master
         from:
           kind: ImageStreamTag
-          name: amp-system:latest
+          name: amp-system:${AMP_RELEASE}
       type: ImageChange
   status:
     availableReplicas: 0
@@ -3194,7 +3104,7 @@ objects:
         - system-sidekiq
         from:
           kind: ImageStreamTag
-          name: amp-system:latest
+          name: amp-system:${AMP_RELEASE}
       type: ImageChange
   status:
     availableReplicas: 0
@@ -3337,7 +3247,7 @@ objects:
         - system-sphinx
         from:
           kind: ImageStreamTag
-          name: amp-system:latest
+          name: amp-system:${AMP_RELEASE}
       type: ImageChange
   status:
     availableReplicas: 0
@@ -3614,7 +3524,7 @@ objects:
         - zync
         from:
           kind: ImageStreamTag
-          name: amp-zync:latest
+          name: amp-zync:${AMP_RELEASE}
       type: ImageChange
   status:
     availableReplicas: 0
@@ -3720,7 +3630,7 @@ objects:
         - que
         from:
           kind: ImageStreamTag
-          name: amp-zync:latest
+          name: amp-zync:${AMP_RELEASE}
       type: ImageChange
   status:
     availableReplicas: 0
@@ -3803,7 +3713,7 @@ objects:
         - postgresql
         from:
           kind: ImageStreamTag
-          name: zync-database-postgresql:latest
+          name: zync-database-postgresql:${AMP_RELEASE}
       type: ImageChange
   status:
     availableReplicas: 0
@@ -3966,7 +3876,7 @@ objects:
         - apicast-staging
         from:
           kind: ImageStreamTag
-          name: amp-apicast:latest
+          name: amp-apicast:${AMP_RELEASE}
       type: ImageChange
   status:
     availableReplicas: 0
@@ -4091,7 +4001,7 @@ objects:
         - apicast-production
         from:
           kind: ImageStreamTag
-          name: amp-apicast:latest
+          name: amp-apicast:${AMP_RELEASE}
       type: ImageChange
   status:
     availableReplicas: 0
--- a/pkg/3scale/amp/operator/apicast_reconciler.go
+++ b/pkg/3scale/amp/operator/apicast_reconciler.go
@@ -2,6 +2,7 @@ package operator
 
 import (
 	"github.com/3scale/3scale-operator/pkg/3scale/amp/component"
+	appsv1alpha1 "github.com/3scale/3scale-operator/pkg/apis/apps/v1alpha1"
 	appsv1 "github.com/openshift/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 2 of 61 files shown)


miguelsorianod:
Some important information:
* We need to take care of the upgrade with the templates too. I've been looking at it and what we need to do there is:
  * Add the new ImageStream version tag (2.8) for every component
  * For every DeploymentConfig in our deployment (in the following specified order)
    * Add the new ImageChange trigger with. For example in the case of APIcast production:
      ```
      oc set triggers dc/apicast-production --from-image=amp-apicast:<new-tag> --containers=system-master-svc,apicast-production
      ```
    * Remove the "latest" ImageChange trigger. For example in the case of APIcast production:
      ```
      oc set triggers dc/apicast-production --from-image=amp-apicast:<old-tag> --containers=system-master-svc,apicast-production --remove
      ```
      Notice that temporarily the DC ImageStream triggers field apparently can have two or more ImageChange triggers at the same time from different tags.
      Doing tests, I've seen that changing/adding/removing a trigger itself in a DC does NOT produce a redeploy of the DC/Pods in it by itself. Changing the trigger only redeploys the DC/pods when changing the trigger changes the image itself. This means that If for example you change from trigger "2.7" to "latest" or viceversa and the images that those imagestreamtags are equal that would not produce a redeployment of the DC
* For the Databases, the upgrade would maybe provoke the DBs to be recreated in case the DB tag has a new sha. Even if the URL does not change it could happen that there's a new image for that tag as DB image tags are floating for minor versions
* 
* In CircleCI config we have the following code:
   ```
            oc image mirror $(for component in apicast zync ; do
              echo 172.30.1.1:5000/$project/amp-$component:latest=quay.io/3scale/$component:nightly
            done) 172.30.1.1:5000/$project/amp-backend:latest=quay.io/3scale/apisonator:nightly \
            172.30.1.1:5000/$project/amp-system:latest=quay.io/3scale/porta:nightly --insecure
   ```
with removal of "latest" tag this can no longer be done. What do we do with it? I thought of using the version but in master there's no "specific" version :thinking:

pkg/3scale/amp/component/apicast_options_builder.go
```diff
@@ -9,12 +9,14 @@ import (
...
-	tenantName     string
-	wildcardDomain string
+	appLabel           string
+	managementAPI      string
+	openSSLVerify      string
+	responseCodes      string
+	tenantName         string
+	wildcardDomain     string
+	productionImageTag string
+	stagingImageTag    string
```
miguelsorianod: Decided to provide the tag value as an option of the component, having one option per DeploymentConfig (configurable at DeploymentConfig level)

pkg/3scale/amp/operator/deployment_config_reconciler.go
```diff
@@ -127,3 +128,46 @@ func NewCreateOnlyDCReconciler() *CreateOnlyDCReconciler {
...
 }
+
+func DeploymentConfigReconcileDeploymentTriggerImageChangeParamsName(desired, existing *appsv1.DeploymentConfig, logger logr.Logger) bool {
+	update := false
+	findDeploymentTriggerOnImageChange := func(triggerPolicies []appsv1.DeploymentTriggerPolicy) (int, error) {
+		result := -1
+		for i := range triggerPolicies {
+			if triggerPolicies[i].Type == appsv1.DeploymentTriggerOnImageChange {
+				if result != -1 {
+					return -1, fmt.Errorf("found more than one imageChangeParams Deployment trigger policy")
```
miguelsorianod: You can have more than one ImageChange deploymentconfig trigger at the same time but I'm going to assume this should not happen in our DeploymentConfigs

pkg/3scale/amp/operator/deployment_config_reconciler.go
```diff
@@ -127,3 +128,46 @@ func NewCreateOnlyDCReconciler() *CreateOnlyDCReconciler {
...
+		if result == -1 {
+			return -1, fmt.Errorf("no imageChangeParams deployment trigger policy found")
+		}
+
+		return result, nil
+	}
+
+	desiredDeploymentTriggerImageChangePos, err := findDeploymentTriggerOnImageChange(desired.Spec.Triggers)
+	if err != nil {
+		panic(fmt.Sprintf("unexpected: '%s' in DeploymentConfig '%s'", err, desired.Name))
```
miguelsorianod: In case the ImageStream imagechange trigger has not been found or more than one imagecange trigger has been found we panic and finish the execution

pkg/3scale/amp/operator/apicast_reconciler.go
```diff
@@ -51,6 +51,9 @@ func (r *ApicastStagingDCReconciler) IsUpdateNeeded(desired, existing *appsv1.De
 	tmpUpdate = DeploymentConfigReconcileContainerResources(desired, existing, r.Logger())
 	update = update || tmpUpdate
 
+	tmpUpdate = DeploymentConfigReconcileDeploymentTriggerImageChangeParamsName(desired, existing, r.Logger())
```
miguelsorianod: We reconcile the image change trigger, specifically the name of it. I'm unsure on whether we should reconcile this. What do you think, should we reconcile ImageChange trigger names in the DCs? It's not directly related to a CR parameter but at the same time makes sense to make sure it has the correct version set. The same situation more or less happens with what we are reconciling on our ImageStreams.

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -50,9 +50,43 @@ func (u *UpgradeApiManager) Upgrade() (reconcile.Result, error) {
...
+}
+
+func (u *UpgradeApiManager) upgradeAPIcastDeploymentConfigs() (reconcile.Result, error) {
+	baseReconciler := NewBaseReconciler(u.Client, u.ApiClientReader, u.Scheme, u.Logger)
+	baseLogicReconciler := NewBaseLogicReconciler(baseReconciler)
+	reconciler := NewApicastReconciler(NewBaseAPIManagerLogicReconciler(baseLogicReconciler, u.Cr))
+	// TODO we are only interested on the DCs reconciliation, specifically on the ImageChange trigger
+	// so maybe is not a good idea to implement upgrade procedure by reconcile procedure??
+	// the Reconcile method will try to reconcile everything that is reconciled in the Apicast reconciler.
+	// This is, DeploymentConfigs(and not only the ImageChange trigger), ConfigMaps, Secrets, ...
```
miguelsorianod: See this comment. Should we just have a specific upgrade procedure changing just what we want instead?

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -50,9 +50,43 @@ func (u *UpgradeApiManager) Upgrade() (reconcile.Result, error) {
...
+	if res.Requeue || err != nil {
+		return res, err
+	}
+
+	// TODO because we are deprecating the use of the "latest" ImageStreamTag
+	// that tag is not necessary anymore. At this moment the upgrade does NOT take
+	// care of removing that ImageStreamTag although it is not used anymore.
+	// Should we take care of doing that?. Keep in mind that this would probably
+	// require some kind of coordination / wait to make sure we do not remove the
+	// tag until the DeploymentConfigs are successfully using the new tag
```
miguelsorianod: See this comment

miguelsorianod:
@eguzki for the moment I've implemented the changes for APIcast DeploymentConfigs. Let's review this and if it's ok / when it's decided how are we going to proceed I'll implement it for the other components.

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -50,9 +50,43 @@ func (u *UpgradeApiManager) Upgrade() (reconcile.Result, error) {
 		}
 	}
 
+	// TODO should there be some kind of coordination/wait between the ImageStream
+	// tags creation and the trigger update/swap ??
```
miguelsorianod: See this comment

eguzki:
This PR conflict with https://github.com/3scale/3scale-operator/pull/286, and since it is older it should be resolved first

pkg/3scale/amp/component/apicast_options_builder.go
eguzki: :+1:

pkg/3scale/amp/auto-generated-templates/amp/amp-eval.yml
```diff
@@ -3970,7 +3960,7 @@ objects:
         - apicast-staging
         from:
           kind: ImageStreamTag
-          name: amp-apicast:latest
+          name: amp-apicast:${AMP_RELEASE}
```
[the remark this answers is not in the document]
miguelsorianod: Hi,

I think we still need to keep the ImageChange trigger.

A ConfigChange trigger defined in a DeploymentConfig (https://docs.openshift.com/container-platform/3.11/dev_guide/deployments/basic_deployment_operations.html#config-change-trigger) results in a new replication controller whenever changes are detected in the PodTemplate of the deployment configuration.

A ImageChange trigger defined in a DeploymentConfig (https://docs.openshift.com/container-platform/3.11/dev_guide/deployments/basic_deployment_operations.html#image-change-trigger) results in a new replication controller whenever the content of an image stream tag changes (when a new version of the image is pushed).

Between 3scale releases (2.7, 2.8, ...) no ImageChange trigger would be needed in case we decide to directly set the image in the 'image' field of the DeploymentConfig and not in the ImageChangeTrigger field.
However, take into account the patch releases, where the AMP_RELEASE does NOT change. The only way to incorporate those changes then is through the refreshment of the ImageStream, and the way to update the DeploymentConfig automatically when an ImageStream changes is by using an ImageChange trigger.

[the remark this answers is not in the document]
miguelsorianod: I've made some additional tests:

If I change a configuration field of the PodTemplateSection of the DeploymentConfig, a redeployment is performed due to the DeploymentConfig Config change trigger. The "generation" metadata field of the DC is incremented. Here the status can be seen:
```
status:
  availableReplicas: 1
  conditions:
  - lastTransitionTime: "2020-02-25T10:42:19Z"
    lastUpdateTime: "2020-02-25T10:42:19Z"
    message: Deployment config has minimum availability.
    status: "True"
    type: Available
  - lastTransitionTime: "2020-02-25T11:13:06Z"
    lastUpdateTime: "2020-02-25T11:13:08Z"
    message: replication controller "backend-worker-7" successfully rolled out
    reason: NewReplicationControllerAvailable
    status: "True"
    type: Progressing
  details:
    causes:
    - type: ConfigChange
    message: config change
  latestVersion: 7
  observedGeneration: 9
  readyReplicas: 1
  replicas: 1
  unavailableReplicas: 0
  updatedReplicas: 1
```

If I change the imageChangeTrigger referenced image in the DeploymentConfig, a redeployment is performed due to the DeploymentConfig Image change trigger. The "generation" metadata field of the DC is incremented. Here the status can be seen:
```
status:
  availableReplicas: 1
  conditions:
  - lastTransitionTime: "2020-02-25T10:42:19Z"
    lastUpdateTime: "2020-02-25T10:42:19Z"
    message: Deployment config has minimum availability.
    status: "True"
    type: Available
  - lastTransitionTime: "2020-02-25T11:07:06Z"
    lastUpdateTime: "2020-02-25T11:07:09Z"
    message: replication controller "backend-worker-6" successfully rolled out
    reason: NewReplicationControllerAvailable
    status: "True"
    type: Progressing
  details:
    causes:
    - imageTrigger:
        from:
          kind: DockerImage
          name: quay.io/miguelsorianod/apisonator-test@sha256:83160cea70919e5ec71fa41277a580722683d9275c3b61382cca3c63eafeae7e
      type: ImageChange
    message: image change
  latestVersion: 6
  observedGeneration: 8
  readyReplicas: 1
  replicas: 1
  unavailableReplicas: 0
  updatedReplicas: 1
```

When an edit is performed on the DC modifying a PodTemplateSpec field and  the referenced image in the ImageChangeTrigger, both being modified at the same time (with oc edit) it has been observed that the generation field has been incremented by two, but I've only been able to see the redeployment of one pod (for example if the original pod was backend-worker-9-* the new one that I observe is backend-worker-10). Same happens with the associated ReplicationController. In the events of the DC I also see just one event of DeploymentCreated that indicates that a new replicationcontroller has been created. Additionally in the status section of the DC I see that in the status field it shows an ImageChange has happened:
```
status:
  availableReplicas: 1
  conditions:
  - lastTransitionTime: "2020-02-25T10:42:19Z"
    lastUpdateTime: "2020-02-25T10:42:19Z"
    message: Deployment config has minimum availability.
    status: "True"
    type: Available
  - lastTransitionTime: "2020-02-25T11:35:37Z"
    lastUpdateTime: "2020-02-25T11:35:40Z"
    message: replication controller "backend-worker-10" successfully rolled out
    reason: NewReplicationControllerAvailable
    status: "True"
    type: Progressing
  details:
    causes:
    - imageTrigger:
        from:
          kind: DockerImage
          name: quay.io/miguelsorianod/apisonator-test@sha256:83160cea70919e5ec71fa41277a580722683d9275c3b61382cca3c63eafeae7e
      type: ImageChange
    message: image change
  latestVersion: 10
  observedGeneration: 14
  readyReplicas: 1
  replicas: 1
  unavailableReplicas: 0
  updatedReplicas: 1
```

It does not show that a ConfigChange has happened, but the pod has the change applied to the PodTemplateSpec too.

pkg/3scale/amp/component/apicast_options_builder.go
```diff
@@ -9,12 +9,14 @@ import (
...
-	responseCodes  string
-	tenantName     string
-	wildcardDomain string
+	appLabel           string
+	managementAPI      string
+	openSSLVerify      string
+	responseCodes      string
+	tenantName         string
+	wildcardDomain     string
+	productionImageTag string
```
eguzki: why `productionImageTag` and `stagingImageTag`. Is there any use case where they will be different?

why not use `ampRelease` property like in other option structures

pkg/3scale/amp/operator/apicast_reconciler.go
eguzki: IMO we should not reconcilliate imageChangeTrigger. We do not reconcilliate either `ampRelease` (which has the same value) for system deployments. 

We reconcilliate imagestreams because images are exposed in the CR.

pkg/3scale/amp/operator/deployment_config_reconciler.go
eguzki: already commented

pkg/3scale/amp/operator/deployment_config_reconciler.go
eguzki: already commented

pkg/3scale/amp/operator/upgrade.go
eguzki: Good question. I would expect that OCP takes care of all going behind the scenes and, thus, we do not need to wait. 

If imagestream new tags are not available when we change image trigger, I expect the DC would fail to "evaluate" the new image and will retry eventually.

Thus, I would not implement wait, and keep an eye open just in case we find issues.

pkg/3scale/amp/operator/upgrade.go
eguzki: I would implement removing latest tags from existing imagestreams.

eguzki: Let's try to have upgraded deployments as close as we can to new deployments. Otherwise, divergences might be cause unexpected events.

pkg/3scale/amp/operator/upgrade.go
eguzki: I kind of agree, reconciliation logic can be used when useful. For this case, I would only update required field.

pkg/3scale/amp/operator/upgrade.go
miguelsorianod: :+1:

pkg/3scale/amp/component/apicast_options_builder.go
miguelsorianod: The name/idea of "ampRelease" does not convince me in this cases. image tags do not necessarily have to have the same value as ampRelease so I wanted to diverge from that concept. I prefer to just provide the image tag that we want.
Both production and apicast images are the same. I think it's not a bad idea to be able to set them separately and the implementation. It's added flexibility for almost negligible implementation cost.

pkg/3scale/amp/operator/apicast_reconciler.go
miguelsorianod: So we leave the reconciliation of the imageChangeTrigger to upgrade-only right? Because we have to change it when a new 3scale release (non-cve) appears

pkg/3scale/amp/operator/upgrade.go
miguelsorianod: :+1:

pkg/3scale/amp/component/apicast_options_builder.go
eguzki: There is no use case for having different tags (image references) for staging and production. They should be the same tag always as production gateway is expected to be the same image as staging, but different configuration.

pkg/3scale/amp/operator/apicast_reconciler.go
eguzki: Yes, we need to upgrade, but not reconcilliate. I do not see any reason to reconcilliate that part of the deployment config.

miguelsorianod:
I've updated the upgrade procedure to perform:
1. Update the ImageStream to add the new version ImageStreamTag
2. Update the DeploymentConfig ImageChangeTrigger to point to the new version ImageStreamTag
3. Delete the "latest" and "2.8" tag from the ImageStream.

This has been done for APIcast DeploymentConfigs.
I'd like a review of this @eguzki, and if we like the approach I will implement it for the other elements.

A note:
I've placed all the changes in the upgrade inside the "UpgradeImages" method, but it will probably change when/if other needs arise. I've thought of moving them into the main "Upgrade" method but it seems to me the final organization might be totally dependent on the needs. For example, we could group by "logical change" or by k8s element, but I see us taking one approach or another totally dependent on the needs that arise.

eguzki:
Looking good. 

Go ahead and implement for other components

miguelsorianod:
Ready for review @eguzki . I've manually tried the upgrade with the current implementation and it seems it works ok

miguelsorianod:
PR updated to not remove the previous version-specific tag. We only remove the "latest" tag now.

The motivation of this is that we preserve the old version-specific tags so it is available what versions has 3scale gone through and we keep the implementation/management of upgrades simpler.

PR #293, opened 2019-12-04
include coverage from integration tests

eguzki:


eguzki:
Integration tests coverage does not contribute a lot. My guess because operator is run in some other goroutine or something like this. Check integration.cov file and operator lines are not there, just test client code lines.

eguzki:
tried several e2e testing modes, and none work (operator sdk 0.8, golang 12.5)

Even [not recommended running go test directly without operator-sdk](https://github.com/operator-framework/operator-sdk/blob/master/doc/test-framework/writing-e2e-tests.md#running-go-test-directly-not-recommended) does not work. `integration.cov` only includes client code lines.

PR #294, opened 2019-12-04
testing coverage report on any package

eguzki:
Testing coverage is reduced from ~60% to 40%. However, the total line numbers included in the report is increased from 3k to 10k. So, numbers are more accurate.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ e2e: e2e-clean e2e-setup e2e-run
 
 $(PROJECT_PATH)/_output/unit.cov:
 	mkdir -p "$(PROJECT_PATH)/_output"
-	go test ./pkg/... -v -covermode=count -test.coverprofile="$(PROJECT_PATH)/_output/unit.cov"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -69,7 +69,7 @@ e2e: e2e-clean e2e-setup e2e-run
 
 $(PROJECT_PATH)/_output/unit.cov:
 	mkdir -p "$(PROJECT_PATH)/_output"
-	go test ./pkg/... -v -covermode=count -test.coverprofile="$(PROJECT_PATH)/_output/unit.cov"
+	go test ./pkg/... -v -tags=unit -covermode=count -coverprofile $(PROJECT_PATH)/_output/unit.cov -coverpkg ./...
```
miguelsorianod: I understand -tags=unit has been added just in case in the future we want to explicitely disallow one of the test files to be run by writing //+build !unit. If it's not the case please tell me what's the motivation. Thank you.

eguzki: The idea is to separate Go tests with Build Tags. Then you can filter.

PR #295, opened 2019-12-12
[WIP] Add s3 apicompatible support

miguelsorianod:
Add support for S3 API-compatible services.

The final desired state of this PR is:
 * Use a secret to provide S3 configuration and mount it into the system pods that need it, as a volume, instead of using the CR fields. The reason for this is to decouple the contents of the S3 configuration from the operator. The operator just provides the secret to System without taking care of what's the content of the configuration. Decoupling this allows us to not having to change the operator code when new S3 configuration for System is adde
 * Have an upgrade procedure at operator level to migrate from the environment variables based approach to the volume approach
 * Have an upgrade procedure at templates level to migrate from the environment variables based approach to the volume approach


pkg/3scale/amp/template/ampevals3.go
```diff
@@ -10,23 +11,31 @@ func init() {
...
+	template.Annotations["description"] = "3scale API Management main system (Evaluation) with shared file storage in AWS S3."
 }
 
 type AmpEvalS3TemplateFactory struct {
 }
 
 func (f *AmpEvalS3TemplateFactory) Adapters() []adapters.Adapter {
-	ampFactory := NewAmpTemplateFactory()
-	evalAdapter := adapters.NewEvalAdapter()
-	s3Adapter := adapters.NewS3Adapter()
```
miguelsorianod: As you can see, we've removed S3 component completely from templates and operator

pkg/3scale/amp/component/system.go
```diff
@@ -76,19 +76,31 @@ const (
...
+)
+
 func NewSystem(options *SystemOptions) *System {
 	return &System{Options: options}
 }
 
 func (system *System) Objects() []common.KubernetesObject {
-	sharedStorage := system.SharedStorage()
+	objects := []common.KubernetesObject{}
+	if system.fileStorageIsPVC() {
```
miguelsorianod: Now we only return the system's PVC shared storage as an object when the FileStorage PVC option is provided. This Objects method is used by the templates

pkg/3scale/amp/component/system.go
```diff
@@ -269,16 +280,6 @@ func (system *System) buildSystemBaseEnv() []v1.EnvVar {
 	systemBackendInternalAPIPass := envVarFromSecret("CONFIG_INTERNAL_API_PASSWORD", "backend-internal-api", "password")
 	result = append(result, systemBackendInternalAPIUser, systemBackendInternalAPIPass)
 
-	if system.Options.s3FileStorageOptions != nil {
-		result = append(result,
```
miguelsorianod: We remove all the envvars as they are not used anymore.
At this moment we should decide how system will be aware of in what mode is in: S3 or PVC. We notified that to System by setting the "FILE_UPLOAD_STORAGE" environment variable to "s3" but not sure whether this approach is the one we want. Maybe System could just check if the configuration file exists? If the secret exists or the volume is mounted or a directory is created?

pkg/3scale/amp/component/system.go
```diff
@@ -499,9 +490,7 @@ func (system *System) MasterApicastSecret() *v1.Secret {
 
 func (system *System) appPodVolumes() []v1.Volume {
 	res := []v1.Volume{}
-	if system.Options.pvcFileStorageOptions != nil {
-		res = append(res, system.FileStorageVolume())
-	}
+	res = append(res, system.fileStorageVolume())
```
miguelsorianod: We now always add a volume, being a PVC-based volume or a Secret-based volume so I tried to generalize the code at this point to just get the appropriate volume based on the provided options

pkg/3scale/amp/component/system.go
```diff
@@ -767,6 +756,33 @@ func (system *System) FileStorageVolume() v1.Volume {
 	}
 }
 
+func (system *System) s3FileStorageVolume() v1.Volume {
+	return v1.Volume{
+		// TODO here we could use the secret name provided as the volume name but
+		// this means if the secret name changes the value that arrives here would
+		// change too. Do we really want to allow that to happen/reconcile it?
+		Name: SystemFileStorageS3VolumeName,
```
miguelsorianod: See the comment above

pkg/3scale/amp/component/system.go
```diff
@@ -767,6 +756,33 @@ func (system *System) FileStorageVolume() v1.Volume {
...
+	if system.Options.pvcFileStorageOptions != nil {
+		return system.pvcFileStorageVolume()
+	}
+
+	if system.Options.s3FileStorageOptions != nil {
+		return system.s3FileStorageVolume()
+	}
+
+	// TODO should we panic? what would be best approach here?
+	panic("One of the volume types should be defined")
```
miguelsorianod: See comment above

pkg/3scale/amp/component/system.go
```diff
@@ -898,34 +912,65 @@ func (system *System) systemConfigVolumeMount() v1.VolumeMount {
...
 func (system *System) appProviderContainerVolumeMounts() []v1.VolumeMount {
-	return system.appCommonContainerVolumeMounts(false)
+	return system.appCommonContainerVolumeMounts(!system.fileStorageIsPVC())
 }
 
 func (system *System) appDeveloperContainerVolumeMounts() []v1.VolumeMount {
-	// TODO why system-app developer container has the system-config volume set to true? is it really necessary?
+	// TODO why system-app developer container has the system-storage VolumeMount set
+	// to readonly true? is it really necessary?
 	// other containers in the same pod have it to false
```
miguelsorianod: See comment above too.

pkg/3scale/amp/component/system.go
```diff
@@ -898,34 +912,65 @@ func (system *System) systemConfigVolumeMount() v1.VolumeMount {
...
+	}
+	if system.fileStorageVolume().Secret != nil {
+		return system.s3FileStorageVolumeMountPath()
 	}
+
+	// Should we panic?
+	panic("At least one of both file storages types has to be set")
+}
+
+func (system *System) s3FileStorageVolumeMountPath() string {
```
miguelsorianod: At this moment we are mounting the S3 config in a DIFFERENT directory than the system extra-configs volume mount (system-config). By looking at documentation/links we might be able to share the same directory without overlapping/hidding other volumes for what I see here https://blog.sebastian-daschner.com/entries/multiple-kubernetes-volumes-directory via the subpath attribute.. I have not implemented it as a first approach. The main reason for not having done it as of now is because doing that would imply having to change/migrate the configuration of the system-config volume for already existing client installations.

pkg/3scale/amp/component/system_options_builder.go
```diff
@@ -8,11 +8,7 @@ import (
 )
 
 type S3FileStorageOptions struct {
-	AWSAccessKeyId       string
-	AWSSecretAccessKey   string
-	AWSRegion            string
-	AWSBucket            string
-	AWSCredentialsSecret string
+	S3ConfigSecret string
```
miguelsorianod: We expect the secret name as the S3 file storage option now. At this moment we only care about that. We are not performing any kind of validation of the keys of the secret nor the contents of the keys. We just mount the secret(and automatically all of the keys). What are your thoughts on this?

pkg/3scale/amp/component/system_options_builder.go
```diff
@@ -290,6 +281,9 @@ func (s *SystemOptionsBuilder) setRequiredOptions() error {
 		return fmt.Errorf("no wildcard domain has been provided")
 	}
 
+	if s.options.s3FileStorageOptions != nil && s.options.pvcFileStorageOptions != nil {
```
miguelsorianod: Should we have this validation here? Should we assume the user of this class will "use it correctly" and therefore is not needed?

pkg/3scale/amp/operator/system.go
```diff
@@ -223,58 +210,39 @@ func (o *OperatorSystemOptionsProvider) setResourceRequirementsOptions(b *compon
...
+
+	return err
 }
 
-func (o *OperatorSystemOptionsProvider) setAWSSecretOptions(sob *component.SystemOptionsBuilder) error {
-	awsCredentialsSecretName := o.APIManagerSpec.System.FileStorageSpec.S3.AWSCredentials.Name
-	currSecret, err := getSecret(awsCredentialsSecretName, o.Namespace, o.Client)
+func (o *OperatorSystemOptionsProvider) setS3FileStorageOptions(b *component.SystemOptionsBuilder) error {
+	s3ConfigurationSecretName := o.APIManagerSpec.System.FileStorageSpec.S3.ConfigurationSecretRef.Name
+	_, err := getSecret(s3ConfigurationSecretName, o.Namespace, o.Client)
```
miguelsorianod: We make sure the specified secret in the CR exists. We do not perform any other kind of validation at keys level nor key contents level at this moment

pkg/3scale/amp/operator/system_reconciler.go
```diff
@@ -118,7 +118,7 @@ func (r *SystemReconciler) reconcileFileStorage(system *component.System) error
 		if r.apiManager.Spec.System.FileStorageSpec.PVC != nil {
 			return r.reconcileSharedStorage(system.SharedStorage())
 		} else if r.apiManager.Spec.System.FileStorageSpec.S3 != nil {
-			return r.reconcileS3AWSSecret(system.S3AWSSecret())
+			return nil // Do we need to reconcile something?
```
miguelsorianod: See comment

pkg/3scale/amp/template/adapters/system.go
```diff
@@ -7,14 +7,15 @@ import (
 )
 
 type System struct {
+	FileStorageType component.SystemFileStorageType
 }
 
-func NewSystemAdapter() Adapter {
-	return NewAppenderAdapter(&System{})
+func NewSystemAdapter(fileStorageType component.SystemFileStorageType) Adapter {
```
miguelsorianod: To allow configurability of the storage type at System level via the templates we've added a new attribute to the System template-specific struct

pkg/3scale/amp/template/adapters/system.go
```diff
@@ -161,6 +165,52 @@ func (s *System) options() (*component.SystemOptions, error) {
 	sob.BackendSharedSecret("${SYSTEM_BACKEND_SHARED_SECRET}")
 	sob.TenantName("${TENANT_NAME}")
 	sob.WildcardDomain("${WILDCARD_DOMAIN}")
-	sob.StorageClassName(nil)
+
+	s.setFileStorageOptions(&sob)
 	return sob.Build()
 }
+
+func (s *System) setFileStorageOptions(sob *component.SystemOptionsBuilder) {
```
miguelsorianod: In templates we now set some options conditionally depending on the desired storage type for system, which changes depending on the template being used

pkg/3scale/amp/template/adapters/system.go
```diff
@@ -161,6 +165,52 @@ func (s *System) options() (*component.SystemOptions, error) {
...
+				Name:        "RWX_STORAGE_CLASS",
+				Description: "The Storage Class to be used by ReadWriteMany System's PVC",
+				Required:    false,
+				Value:       "null", //TODO this is incorrect. The value that should be used at the end is null and not "null", however template parameters only accept strings.
+			},
+		}
+	case component.SystemFileStorageTypeS3:
+		return []templatev1.Parameter{
+			templatev1.Parameter{
+				Name:        "S3_CONFIGURATION_SECRET_NAME",
```
miguelsorianod: The S3 template now has a new required parameter being the secret name configuration and the old parameters that were the bucket name, region etc... have been removed.
This is an important change on how the s3 template is deployed because now to deploy it it requires the user to pre-create the S3 configuration secret BEFORE deploying the template, whereas before they just filled the S3-specific related values

pkg/3scale/amp/template/adapters/system.go
```diff
@@ -161,6 +165,52 @@ func (s *System) options() (*component.SystemOptions, error) {
...
+				Description: "Name of the secret that contains S3 configuration for System's File Storage. The secret name needs to exist previously.",
+				Required:    true,
+			},
+		}
+	default:
+		panic("s3 or pvc has to be set as system's file storage option")
```
miguelsorianod: Is it OK to panic?

pkg/3scale/amp/template/amp.go
```diff
@@ -29,7 +29,7 @@ func (f *AmpTemplateFactory) Adapters() []adapters.Adapter {
 		adapters.NewBackendAdapter(),
 		adapters.NewMysqlAdapter(),
 		adapters.NewMemcachedAdapter(),
-		adapters.NewSystemAdapter(),
+		adapters.NewSystemAdapter(component.SystemFileStorageTypePVC),
```
miguelsorianod: As you can see, we now create the System adapter with the desired fileStorage type depending on the template

pkg/apis/apps/v1alpha1/apimanager_types.go
```diff
@@ -229,9 +229,14 @@ type SystemPVCSpec struct {
...
 
 type SystemS3Spec struct {
-	AWSBucket      string                  `json:"awsBucket"`
-	AWSRegion      string                  `json:"awsRegion"`
-	AWSCredentials v1.LocalObjectReference `json:"awsCredentialsSecret"`
+	// TODO we now have a problem which is that some fields that were required
+	// when specifying S3 as the File Storage should no longer be required as
+	// the configuration is located in ConfigurationSecretName. What happens
+	// with existing installations?
+	AWSBucket              string                  `json:"awsBucket"`
```
miguelsorianod: This is what concerns me more about all:
We had some required fields in the CR to provide the S3 configuration that we have decided we are not going to use anymore, but they are required. And also, we are adding a new required field which is the configuration secret.
I think this breaks compatibility right? because the existing deploymets do not have the new ConfigurationSecretRef required parameter.

pkg/3scale/amp/component/system.go
```diff
@@ -898,34 +912,65 @@ func (system *System) systemConfigVolumeMount() v1.VolumeMount {
...
+func (system *System) fileStorageVolumeMountPath() string {
+	if system.fileStorageVolume().PersistentVolumeClaim != nil {
+		return system.pvcFileStorageVolumeMountPath()
+	}
+	if system.fileStorageVolume().Secret != nil {
+		return system.s3FileStorageVolumeMountPath()
 	}
+
+	// Should we panic?
+	panic("At least one of both file storages types has to be set")
```
miguelsorianod: See comment above

pkg/3scale/amp/template/adapters/system.go
```diff
@@ -161,6 +165,52 @@ func (s *System) options() (*component.SystemOptions, error) {
...
+		// expansion, which does not exist in Kubernetes. We take care of replacing
+		// the value with the double brace parameter expansion as a postProcess
+		// step after generating the template objects
+		sob.PVCFileStorageOptions(component.PVCFileStorageOptions{})
+	case component.SystemFileStorageTypeS3:
+		sob.S3FileStorageOptions(component.S3FileStorageOptions{
+			S3ConfigSecret: "${S3_CONFIGURATION_SECRET_NAME}",
+		})
+	default:
+		panic("s3 or pvc has to be set as system's file storage option")
```
miguelsorianod: Should we panic?

miguelsorianod:
@eguzki I've implemented a first approach of just the installation part with the new method for both templates and operator. Several assumptions have been made that I'd like to discuss. I left those in comments. Can you review them and the current code?

I've also made several refactors to try to accommodate the configurability of the storage type on the templates and removing the S3 component as it was just doing the creation of the S3 secret (does not exist anymore) and postprocessing some fields and I replaced that postprocessing behavior by configurability at System component level.

Thank you.

pkg/3scale/amp/component/system.go
eguzki: Why don't you add those comments in the code instead of PR's comments

pkg/3scale/amp/component/system.go
eguzki: I do not think this is the correct approach. 

This kind of logic belongs to `operator` or `template` packages. `component` packages should only contain functions to generate OCP objects. 

High level there should be: `func s3AppDeploymentConfig()` and `func pvcAppDeploymentConfig` and make all generalizations (to leverage code reuse) downstream until you get your implementation. In other words, the volume could be created and passed to a function that creates deploymentconfig and uses the param to populate `VolumeMounts`.

deploy/crds/apps_v1alpha1_apimanager_crd.yaml
```diff
@@ -119,16 +119,19 @@ spec:
...
+                          type: string
+                        configurationSecretRef:
                           properties:
                             name:
                               type: string
                           type: object
-                        awsRegion:
-                          type: string
                       required:
                       - awsBucket
```
eguzki: still required?

pkg/apis/apps/v1alpha1/apimanager_types.go
eguzki: One proposal:

* deprecate `spec.filestorage.amazonSimpleStorageService`which is optional
* create new optional field: `spec.filestorage.s3secret` with `v1.LocalObjectReference` type.

Upgrading path:

Gather all the information from `spec.filestorage.amazonSimpleStorageService` and `aws-auth` s3 secret and update `aws-auth` s3 secret with all data. Update CR to look like something like this:

```
spec:
  system:                                            
    fileStorage:                                     
      s3secret:                    
        name: aws-auth                  
```

pkg/3scale/amp/component/system.go
eguzki: As I understood from the meeting with system guys, `FILE_UPLOAD_STORAGE` is still there to notify type of storage

eguzki:
Comments available.

pkg/3scale/amp/component/system.go
miguelsorianod: I have concerns about the proposed approach as you are going the way the same approach as the templates take. Once you start to have multiple combinations you would have to implement each one of those and start to name them differently for each combination, even if you reuse code:

```
AppDeploymentConfigWithS3AsFileStorage
AppDeploymentConfigWithPVCAsFileStorage
AppDeploymentConfigWithS3AsFileStorageAndAnotherSpecificConfig1
AppDeploymentConfigWithS3AsFileStorageAndAnotherSpecificConfig2
AppDeploymentConfigWithPVCAsFileStorageAndAnotherSpecificConfig1
AppDeploymentConfigWithPVCAsFileStorageAndAnotherSpecificConfig2
```

repeat for the N possible combinations among them.

pkg/3scale/amp/component/system.go
miguelsorianod: It is still there. I was saying this to evaluate/discuss whether another approach would be better than using the environment variable

pkg/apis/apps/v1alpha1/apimanager_types.go
miguelsorianod: What you say about deprecating that field and then migrating it to the new configuration secret seems a good solution.
The idea of that new secret is that it HAS to be provided by the user with any name we want. What do we do in the case of the migration that we are the ones creating it? Do we hardcode a specific name and document in the release notes that it is created with that name? For new installations that secret name will not be created as users will use the name  they want.

Regarding the proposed field name, you cannot use numbers as part of the name of the fields because it fails openapi validation related code https://github.com/operator-framework/operator-sdk/issues/1120. That's the reason I chose amazonSimpleStorageService instead of s3 for that section so we'd need to choose another name.

Do you think it might be better to set the s3secret field (required!) inside a new section in fileStorage(amazonsthree for example) instead of being directly a field in the FileStorage section? Keep in mind also that inside the fileStorage section there are also other fields NOT related to S3 so if in the future we need to add more s3 related fields and we directly place the s3secret there we would end up with more than one s3 related field in that section mixed with other filestorage configurability elements, not related to s3, making us not able/really hard to check what deployment method is used for fileStorage. Creating a new specific section for it would allow us to add more optional fields related to s3 in case we need them

Example:
```
spec:
  system:                                            
    fileStorage:                                     
      amazonsthree(this is not going to be the final name probably):                    
        sthreesecret:
          name: aws-auth
```

instead of
```
spec:
  system:                                            
    fileStorage:                                     
      s3secret:                    
        name: aws-auth
```

What do you think?

miguelsorianod:
@eguzki comments answered

pkg/3scale/amp/component/system.go
eguzki: @thomasmaas  who is going to work on this task in system? we would like to know his/her opinion about this

pkg/apis/apps/v1alpha1/apimanager_types.go
eguzki: Ok, it allows some extensibility. But I cannot think of anything being outside the secret.

eguzki:
I am concerned about the increased complexity of the `component.System` module. `component` package implements OCP resource model and should be simple and flexible to be customizable. It is a simple matter of separation of concerns. Currently the `component` package includes too many branches that makes it difficult to extend and reuse. `component` package should be a list of "lego` bricks. Simple, easy to use, easy to create new ones and reuse them. IMO, we are getting far from this model.

Also using a single `SystemOptions` struct to build all system components is making hard to extend it.  It feels like `SystemOptions` should be divided into smaller components and each component should receive only a small part. I guess this would be a refactor out of scope of this PR, however, this PR shows the limitations and concerns I was referring to and IMO justifies a need to refactor.

pkg/3scale/amp/component/system.go
miguelsorianod: After talking with System, mounting the S3 config in a different directory than "system-extra-configs", where currently there are some configurations already, does not seem a good solution due to for every new secret/configmap with keys that we want to mount we would have to do it in a different directory, which would mean System code and Operator/Templates code would have to change every time.

I've explored the option to mount two ConfigMaps/Secrets in the same directory but this does not seem to be possible with how Kubernetes VolumeMounts work and the current Volume options we have. The reasons are:
* You cannot have multiple VolumeMounts in a container with the same "mountPath" attribute. The reason for that is that the mountPath attribute is used as a merge key when multiple volumeMounts have the same mountPath volume. This can be seen here: https://github.com/kubernetes/api/blob/7dc09db16fb8ff2eee16c65dc066c85ab3abb7ce/core/v1/types.go#L2181. So, if you try to create to volumeMounts with the same mountPath value, the two volumeMounts are merged into one.
* It seems there's sort of way to mount several fields of different Secrets/ConfigMaps in the same directory, but it implies changing the SubPath explicitely defining the key name of the desired Secret/ConfigMap to mount. This defeats precisely the purpose of making additions/changes without System/Operator/Templates being aware of it (and it also would imply a migration from templates/operator is needed). An example of this can be found here https://blog.sebastian-daschner.com/entries/multiple-kubernetes-volumes-directory. As you can observe, you have to explicitly write the secret attribute key name to mount in the subPath and in the mountPath attributes.
* We found the existence of "Projected" volumes that might seem interesting for the purpose but it is not enabled by default in Kubernetes 1.11 https://kubernetes.io/docs/concepts/storage/volumes/#projected so we cannot use it, as it's a version we must support (for OpenShift 3.11)

eguzki:
shall we close this @miguelsorianod

miguelsorianod:
Yes I'd say so, as I don't think is going to be prioritized soon.
The branch is still there so we can always open a new one.

Closing it.

PR #299, opened 2020-01-07
Productize 2.8 release

miguelsorianod:
Changes:

~Addition of new CSV version~
~Addition of new OLM channel 2.8-stable and make it default~
~Update 3scale component images in code and YAML files (operator.yaml and new CSV)~
Adapt Memcached readiness probe to memcached productized image
Update Memcached image to use in tests with latest productized one
Add RedHat registry credentials in CircleCI config
Update APIManager controller e2e tests to use RedHat registry secret
Update documentation to provide RedHat registry information
Update documentation to match new images and 3scale release version
Update 3scale and APIcast templates
Update 3scale release version from 2.7 to 2.8
Update 3scale operator version from 0.4.0 to 0.5.0

You will see there are references to 3scale memcached 2.7 (and not 2.8) in CircleCI config file and in the apimanager_controller_test e2e test. That's intentional due to 2.8 registry.redhat.io images are not out until GA and we cannot use the upstream one for the productized branch because the readinessProbe for the system-memcache DC is different (the same thing was done on the 2.7 release, the memcached image used in the 2.6 release was used then in those files).

~Do NOT merge at this moment. Before merging make sure that:~
~* template regeneration is up to date~
~* expected operator image URL in the new CSV is correct (matches brew suffix)~
~* CRD fields~

~Warning: Right now destination branch is "master". Before merging make sure to create a 2.8-stable-prod  branch and change the destination branch in this PR to there.~


miguelsorianod:
When doing a diff of the generated templates between this branch and 2.7-stable-prod with:
git diff 2.7-stable-prod prepare-2.8-stable-prod  --  ./auto-generated-templates/amp/amp.yml

I realized in the pre-hook pod command line the MASTER_ACCESS_TOKEN reference was still there in 2.7 and not referenced via a secret reference, and the secret reference was not created in the environment of the pre-hook pod either, which should have been there (it seems it was an error).

For this new release probably we'll need to change the pre-hook pod command again because system will fix the workaround we made so this would resolve our problem when applying the upgrade on the operator side.
On the templates side we will need to document for this release again how to upgrade the pre-hook pod command and how to add the new master access token secret reference in the pre-hook pod environment

miguelsorianod:
I think we can merge this now

PR #302, opened 2020-01-10
S3 apicompatible via envvars

miguelsorianod:
PR that allows to configure an alternative endpoint for S3 through environment variables. We expose the configurability of the envvars through new optional CR fields in the operator and through new OpenShift template parameters in the templates

This PR is an alternative to the https://github.com/3scale/3scale-operator/pull/295 where there's an inprogress implementation via the usage of configuration file in a secret and mounting the secret.
The reason for this PR is to have an alternative that we know it should work, as currentl #295 feasibility/advantages are not clear for what has been discussed (see that PR for details)

- [x] All configuration related to S3 located in a single secret
Deprecate CRD fields `spec.system.fileStorage.amazonSimpleStorageService.awsBucket` and `spec.system.fileStorage.amazonSimpleStorageService.awsRegion`. Upgrade procedure will move that information to S3 secret.
With the implemetantion in #302 s3 information will be scattered in a) apimanager CR, b) system enviornment config map and c) aws secret. This PR pretends to put all related s3 configuration in a single location, a secret. The idea also takes into account that in short term, s3 configuration will be passed to system mounting volumes from secret. So this idea of a single location of s3 configuration is a step forward into that direction.

- [x] Fix s3 secret reconciliation
- [x] test templates with AWS S3
- [ ] test templates with https://play.minio.io
- [x] test operator with AWS S3
- [ ] test operator with https://play.minio.io
- [x] test upgrade procedure
- [x] update documentation


pkg/3scale/amp/component/s3.go
```diff
@@ -140,6 +147,8 @@ func (s3 *S3) AddS3PostprocessOptionsToSystemEnvironmentCfgMap(objects []common.
 	systemEnvCfgMap.Data["FILE_UPLOAD_STORAGE"] = "s3"
 	systemEnvCfgMap.Data["AWS_BUCKET"] = s3.Options.awsBucket
 	systemEnvCfgMap.Data["AWS_REGION"] = s3.Options.awsRegion
+	systemEnvCfgMap.Data[StorageServiceEndpointScheme] = s3.Options.storageServiceEndpointScheme
```
miguelsorianod: Notice how this is in the method AddS3PostprocessOptionsToSystemEnvironmentCfgMap.
This is a postprocess method of S3 for the templates.
Be aware the configuration of the envvars and setting of the valus in the system-environment ConfigMap in the templates use this method but NOT in the operator (Postprocess methods are specific and needed for the templates)

pkg/3scale/amp/component/system.go
```diff
@@ -333,6 +335,8 @@ func (system *System) AddS3ConfigIntoEnvironmentConfigMap(configMap *v1.ConfigMa
 	configMap.Data["FILE_UPLOAD_STORAGE"] = "s3"
 	configMap.Data["AWS_BUCKET"] = system.Options.s3FileStorageOptions.AWSBucket
 	configMap.Data["AWS_REGION"] = system.Options.s3FileStorageOptions.AWSRegion
+	configMap.Data[StorageServiceEndpointScheme] = system.Options.s3FileStorageOptions.EndpointScheme
```
miguelsorianod: We save the values of the scheme and host in the system-environment ConfigMap.
This is not really needed (could directly be referenced as an envvar) but for consistency with the other existing S3 fields (AWS_BUCKET, AWS_REGION), and to keep track of what's the value in every moment, and to have a centralized source of truth for the value I've decided to do it this way anyway

pkg/3scale/amp/component/system_options_builder.go
```diff
@@ -13,6 +13,8 @@ type S3FileStorageOptions struct {
 	AWSRegion            string
 	AWSBucket            string
 	AWSCredentialsSecret string
+	EndpointScheme       string
+	EndpointHost         string
```
miguelsorianod: TODO:
In dev environments the port is often set to a non-standard port.
Talk with system team on whether we can set the port or not. I've assumed  that we can and the way to provide it is via the EndpointHost field (separated by a colon when you want to specify a port). In that case maybe a better name than EndpointHost could be used (as it optionally  considers that an optional port separated by a colon can be provided in this attribute)

miguelsorianod: I also thought on use a simple Endpoint attribute that included scheme, host and port but System team gave me feedback saying it would require parsing on their side and that they prefer to split it. Talk with them about that if you consider we should unify everything

pkg/3scale/amp/auto-generated-templates/amp/amp-s3.yml
```diff
@@ -4548,3 +4600,9 @@ parameters:
   name: AWS_BUCKET
 - description: AWS Region to use in S3 Storage for assets.
   name: AWS_REGION
+- description: Storage Service URL scheme to use for the S3 API-compatible storage
+    service
+  name: STORAGE_SERVICE_ENDPOINT_SCHEME
+- description: Storage Service URL host and optionally a port separated by a colon
```
miguelsorianod: Change this description to an adequate one if we change the behavior of this field

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -389,3 +394,60 @@ func (u *UpgradeApiManager) upgradeAPIManagerCRDatabaseDefaults() (reconcile.Res
 
 	return reconcile.Result{}, nil
 }
+
+func (u *UpgradeApiManager) upgradeSystemS3() (reconcile.Result, error) {
```
miguelsorianod: With the current implementation, we need to upgrade the system-environment ConfigMap to include the new S3 related fields with their default values (the empty string). We also need to update the environment of the containers that are supposed to use the S3 configuration

pkg/3scale/amp/component/s3.go
```diff
@@ -14,6 +14,11 @@ const (
 	S3SecretAWSSecretAccessKeyFieldName = "AWS_SECRET_ACCESS_KEY"
 )
 
+const (
+	StorageServiceEndpointScheme = "STORAGE_SERVICE_ENDPOINT_SCHEME"
```
guicassolato: System will stick to the convention from other AWS S3 config fields and name these new ones as well `AWS_X`. See https://github.com/3scale/porta/pull/1522.

From System (3scale) perspective, it's AWS S3 API. These values are used to establish a connection to AWS S3 (or a compatible service that follows AWS S3 API). The fields are named after the API, which in this case is a AWS (S3) API.

```suggestion
	StorageServiceEndpointScheme = "AWS_HOSTNAME"
```

pkg/3scale/amp/component/s3.go
```diff
@@ -14,6 +14,11 @@ const (
...
+	StorageServiceEndpointHost   = "STORAGE_SERVICE_ENDPOINT_HOST" // TODO Includes Host and Port. Maybe choose a better name?
```
guicassolato:
```suggestion
	StorageServiceEndpointHost   = "AWS_HOSTNAME" // TODO Includes Host and Port. Maybe choose a better name?
```

See https://github.com/3scale/3scale-operator/pull/302/files?file-filters%5B%5D=.go#r365307907

pkg/3scale/amp/component/s3.go
miguelsorianod: I don't really like the idea of using AWS_* for the environment variables.
Precisely the motivation of this PR is to allow S3 api-compatible services to be used, and those services might not be used in AWS at all.
I think it is confusing for a user seeing AWS_PROTOCOL or AWS_HOSTNAME when they want to configure a S3 api-compatible service (like min.io) that is for example deployed in Azure...
There could be an intermediate solution which is envvars being named AWS_* as you proposed but the config exposed to the user (in the ConfigMap, in the CR fields, in the OpenShift templates) named another way that does not include the AWS prefix...

pkg/3scale/amp/component/s3.go
miguelsorianod: See my comment above

pkg/3scale/amp/component/s3.go
miguelsorianod: Also AWS_PROTOCOL by itself even if only used in AWS (which is not the case anymore) I don't think it's a good name. There are multiple services in AWS, so AWS_PROTOCOL what does it mean? what protocol? the protocol used in s3? in rds? in an EC2 machine? it's confusing too

guicassolato: I understand your concern, @miguelsorianod. However, @3scale/system will stick to the current convention for the time being.

The `AWS_` prefix clearly states the nature of the connection to be established. It's about the API, not about the service provider behind. (In fact, it should even be called `AWS_S3_` IMHO, to make them ever more clear.)

Of course, we cannot make any assumptions about the service provider behind. There might be a proxy, for example, routing requests that we believe to be going to AWS S3, in reality, to a different provider (even when the value of the "hostname" config option says `aws.amazon.com`). From our perspective however, we implemented a connection to an AWS S3-provided service; System uses a library that trusts it is really AWS S3 providing the service (written and maintained by Amazon!) For now, anything other than AWS S3 can still be used, but the risk is on the customer. If we update the library and integration with a pretend-to-be-AWS service breaks, we cannot test that, we cannot track that, we can barely explain that to the customer.

And we have analogous cases, such as `THINKING_SPHINX_*`, `REDIS_*`, `MEMCACHE_*`. Those are all configuration options related to specific products in the market (free or commercial) that we once decided to commit to. One could argue now that we are actually committed to those products' APIs, rather than to the products providers themselves. Yet, we won't rename our env vars, for example, because some player in the market decided to create an alternative to Redis that turns out to copy Redis API and promised to be 100% compatible.

Integration with AWS S3 (or AWS S3 compatible services, if you prefer) is not being introduced with this PR. It's already there, and all other env vars related are names `AWS_*`.

hallelujah: Everybody uses `AWS_*` environment variables, I do not see why we don't do that.
Min.io uses it in its client

https://github.com/minio/minio/blob/46b4dd8e20dbcab872206ff7b90774f243ed5bfd/docs/lifecycle/README.md#2-enable-bucket-lifecycle-configuration

eguzki:
Some tests were carried out using https://play.minio.io as S3 provider. Templates and operator results were the same. As of today, test failed. Since it does not seem to be operator issue, but System issue, we have decided to merge this PR. I will report here my findings.

Tried to deploy with the following configuration:
```
AWS_ACCESS_KEY_ID: Q3AM3UQ867SPQQA43P2F
AWS_BUCKET: testbucket
AWS_HOSTNAME: https://play.min.io
AWS_PROTOCOL: ""
AWS_REGION: us-east-1
AWS_SECRET_ACCESS_KEY: ***
```

Tried several values for `AWS_HOSTNAME` field: `play.min.io`, `https://play.min.io`  with the same error from system-provider pod when trying to upload logo:

```
Exception -- [54a9e2ac-d28c-43ad-8fc2-4f5a131bb59f] [3scale-admin.eguzki-test.apps.devel-ocp-41-ga.3sca.net] [147.161.94.114] {:exception=>{:class=>Aws::S3::Errors::PermanentRedirect, :message=>"The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.", :backtrace=>["/opt/system/vendor/bundle/ruby/2.5.0/gems/aws-sdk-core-2.6.34/lib/seahorse/client/plugins/raise_response_errors.rb:15:in `call'", "/opt/system/vendor/bundle/ruby/2.5.0/gems/aws-sdk-core-2.6.34/lib/aws-sdk-core/plugins/s3_sse_cpk.rb:19:in `call'", "/opt/system/vendor/bundle/ruby/2.5.0/gems/aws-sdk-core-2.6.34/lib/aws-sdk-core/plugins/s3_dualstack.rb:24:in `call'", "/opt/system/vendor/bundle/ruby/2.5.0/gems/aws-sdk-core-2.6.34/lib/aws-sdk-core/plugins/s3_accelerate.rb:34:in `call'"]}, :parameters=>{}}
[54a9e2ac-d28c-43ad-8fc2-4f5a131bb59f] [3scale-admin.eguzki-test.apps.devel-ocp-41-ga.3sca.net] [147.161.94.114] 
Aws::S3::Errors::PermanentRedirect (The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.):
  app/controllers/provider/admin/account/logos_controller.rb:20:in `update_logo'
  app/controllers/provider/admin/account/logos_controller.rb:10:in `update'
  lib/three_scale/middleware/multitenant.rb:116:in `_call'
  lib/three_scale/middleware/multitenant.rb:111:in `call'
  lib/action_dispatch/rails5_ssl.rb:86:in `call'

```
The most meaninfui message is: `The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.`

I tried using the aws-sdk gem directly:

```
require 'aws-sdk'
Aws.config.update(
        endpoint: 'https://play.min.io',
        access_key_id: 'Q3AM3UQ867SPQQA43P2F',
        secret_access_key: 'zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG',
        force_path_style: true,
        region: 'us-east-1'
)
s3 = Aws::S3::Client.new
s3.put_object(
        key: 'testobject',
        body: 'Hello from MinIO!!',
        bucket: 'testbucket',
        content_type: 'text/plain'
)
```
And the snippets uploads correctly the content to `https://play.min.io`

I also made sure there is no bug propagating configuration values. `system-provider` container logs outputs all variables values (security issue?? :)), and the values were correct. There is no bug passing configuration to `system-provider` container.

```
declare -x AWS_ACCESS_KEY_ID="Q3AM3UQ867SPQQA43P2F"
declare -x AWS_BUCKET="testbucket"
declare -x AWS_HOSTNAME="https://play.min.io"
declare -x AWS_PROTOCOL=""
declare -x AWS_REGION="us-east-1"
declare -x AWS_SECRET_ACCESS_KEY="****"
```

Tried several values for `AWS_REGION` field: `us-east-1`, `eu-west-1` with no luch

Then I tried a real aws s3 account and real bucket with correct access key and secret but in hostname I configured https://play.min.io

```
AWS_ACCESS_KEY_ID: MyKey
AWS_BUCKET: enterprise-onprem.3sca.net
AWS_HOSTNAME: https://play.min.io
AWS_PROTOCOL: ""
AWS_REGION: us-east-1
AWS_SECRET_ACCESS_KEY: ***
```
Turns out that the assets are stored in aws S3 despite of AWS_HOSTNAME: https://play.min.io. My conclusion is that AWS_HOSTNAME is not being processed correctly.

miguelsorianod:
I tested 3scale with min.io with templates and operator and seems to work. Updated original comment to check the tested fields.
AWS_HOSTNAME does not accept providing a protocol/scheme on system side. It expects just the hostname. It seems the way to provide the values is:

```
AWS_PROTOCOL: "https"
AWS_HOSTNAME: "play.minio.io"
```

or without AWS_PROTOCOL at all (https is set by default on System)

PR #304, opened 2020-01-15
Configure any s3 provider from s3 secret

eguzki:
- [x] All configuration related to S3 located in a single secret
Deprecate CRD fields `spec.system.fileStorage.amazonSimpleStorageService.awsBucket` and `spec.system.fileStorage.amazonSimpleStorageService.awsRegion`. Upgrade procedure will move that information to S3 secret.
With the implemetantion in #302 s3 information will be scattered in a) apimanager CR, b) system enviornment config map and c) aws secret. This PR pretends to put all related s3 configuration in a single location, a secret. The idea also takes into account that in short term, s3 configuration will be passed to system mounting volumes from secret. So this idea of a single location of s3 configuration is a step forward into that direction.

- [x] Fix s3 secret reconciliation
- [x] test templates with AWS S3
- [ ] test templates with https://play.minio.io
- [x] test operator with AWS S3
- [ ] test operator with https://play.minio.io
- [x] test upgrade procedure
- [x] update documentation


pkg/3scale/amp/auto-generated-templates/amp/amp-eval-s3.yml
```diff
@@ -4499,9 +4499,7 @@ parameters:
   name: AWS_BUCKET
 - description: AWS Region to use in S3 Storage for assets.
   name: AWS_REGION
-- description: Storage Service URL scheme to use for the S3 API-compatible storage
-    service
-  name: STORAGE_SERVICE_ENDPOINT_SCHEME
-- description: Storage Service URL host and optionally a port separated by a colon
-    (:) to use as the S3 API-compatible storage service
-  name: STORAGE_SERVICE_ENDPOINT_HOST
+- description: AWS S3 provider endpoint protocol. HTTP or HTTPS.
```
thomasmaas:
```suggestion
- description: AWS S3 compatible provider endpoint protocol. HTTP or HTTPS.
```

not sure if we should add something like that to all descriptions to make clear it's not just AWS?

eguzki: I will update with your change. 

That file is automatically generated, so should not be edited manually.

PR #308, opened 2020-01-22
Add support for PodDisruptionBudget

wei-lee:
## WHAT

This PR will add support for PodDisruptionBudget to the 3scale operator. See [JIRA](https://issues.redhat.com/browse/THREESCALE-4258).

## HOW

A new feature flag is introduced to the `APIManager` CRD to allow users to enable this feature.  An example can be found [here](https://github.com/wei-lee/3scale-operator/blob/ae4397ca2585fabdc57ce83cef118bfb51357e35/deploy/crds/apps_v1alpha1_apimanager_cr_pdb.yaml).

When enabled, the following PDBs are created:

* apicast-production
* apicast-staging
* backend-cron
* backend-listener
* backend-worker
* system-app
* system-sidekiq
* zync
* zync-que

All PDBs have the `MaxUnavailable` field value set to `1`. This will ensure when the node is being drained, only 1 pod per component is allowed to be evicted.  We do not to update/delete these PDBs even if there is only 1 instance of these components are deployed as it won't violate the rules in PDBs. Also all the components can still be scaled up and down as normal. It will be count against the PDB rules, but PDB won't stop replica controllers from scaling down the pods.

## Verification

To verify:

1. Follow the instructions [here](https://github.com/3scale/3scale-operator/blob/master/doc/development.md#deploy-3scale-operator-manually) to build the operator and deploy it an OpensShift 4 cluster. You can use this image `quay.io/wei_lee/3scale-operator:test`.
2. Create a new APIManager CR using [this example](https://github.com/wei-lee/3scale-operator/blob/ae4397ca2585fabdc57ce83cef118bfb51357e35/deploy/crds/apps_v1alpha1_apimanager_cr_pdb.yaml).
3. Wait for deployments start for all the components.
4. List all the PDBs in the namespace and you should see 9 of them. You can also check the content of them to make sure they have the right spec.
5. Try scale down some of the components and you should still be able to do that.
6. Delete the APIManager CR that is created and make sure all the PDBs are deleted as well.


miguelsorianod:
Hi @wei-lee,

Thank you for the contribution.

I've been trying the PodDisruptionBudget and the behaviour I understand it has is the following one:
* As you commented, when draining a node, this PR allows that, when activated, it ensures that at most one pod at a time will be evicted from the node for each component. It is important to note that if the node to drain has only one pod for a component, the PodDisruptionBudget will comply the rules, as the MaxUnavailable is set to 1. This means that this functionality does not prevent a node being left without a pod of a component. What it does, is it makes the draining of the node for all of its pods components to be moved to another node one at a time. That's the intent of this PR right? and not preventing a node being left without a pod of a component.
* As you also commented, the PodDisruptionBudget stats are taken into account when scaling up/down a DeploymentConfig/Deployment or performing some kind of rolling update (for example when you change the PodTemplate spec and this triggers a new deployment), but they are not enforced in this case, so we should be able to perform changes into the Deployment and scale up/down without problems even if the PodDisruptionBudget stats indicate that it shouldn't be possible to evict.

As I understand, apart from enabling the PodDisruptionBudget field, to really use this functionality the number of replicas of the different components have to be configured with more than one replica. Otherwise downtime avoidance is not prevented.

We are currently reviewing the code and we'll get back to you with feedback about it.

Thank you.

go.mod
```diff
@@ -69,3 +69,5 @@ replace (
 	sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.1.10
 	sigs.k8s.io/controller-tools => sigs.k8s.io/controller-tools v0.1.11-0.20190411181648-9d55346c2bde
 )
+
+go 1.13
```
miguelsorianod: In this PR we should not include the change to make Go 1.13 as a requirement. I understand the new changes don't require Go 1.13.

go.mod
```diff
@@ -30,7 +30,7 @@ require (
 	github.com/openshift/client-go v3.9.0+incompatible
 	github.com/operator-framework/operator-sdk v0.8.1-0.20190517223317-f7f644008098
 	github.com/peterbourgon/diskv v2.0.1+incompatible // indirect
-	github.com/prometheus/client_golang v1.1.0 // indirect
+	github.com/prometheus/client_golang v1.1.0
```
miguelsorianod: Why has this changed?

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
```diff
@@ -82,3 +83,11 @@ func (r *BaseAPIManagerLogicReconciler) updateResource(obj common.KubernetesObje
 	r.Logger().Info(fmt.Sprintf("Updated object %s", ObjectInfo(obj)))
 	return r.Client().Update(context.TODO(), obj) // don't wrap error
 }
+
+func (r *BaseAPIManagerLogicReconciler) reconcilePodDisruptionBudget(desiredPDB *v1beta1.PodDisruptionBudget) error {
+	if r.apiManager.Spec.PodDisruptionBudget.Enabled {
```
miguelsorianod: I've been thinking about this.

I'm still thinking on what would be the appropriate way to expose this in the CR. What structure to have and how configurable/what granularity it should have.

If we assume we will use the implemented approach at this moment I have the following comment:

If you deploy an APIManager with this setting enabled it will deploy and reconcile the PodDisruptionBudgets. After APIManager has been deployed, if the user decides to change the PodDisruptionBudget.Enabled to false the PodDisruptionBudget will not be reconciled anymore but it will still exist, and being used in the specified Deployments with the label.
This means this parameter as it is implemented right now it is a "deploy-only" attribute in the CR.

I think we might add reconciliation code to check that when it is not enabled we ensure the PodDisruptionBudgets do not exist, and in case they do remove them. cc @eguzki what do you think?

pkg/3scale/amp/component/backend.go
```diff
@@ -52,6 +53,10 @@ func (backend *Backend) Objects() []common.KubernetesObject {
 	redisSecret := backend.RedisSecret()
 	listenerSecret := backend.ListenerSecret()
 
+	workerPDB := backend.WorkerPodDisruptionBudget()
+	cronPDB := backend.CronPodDisruptionBudget()
+	listenerPDB := backend.ListenerPodDisruptionBudget()
```
miguelsorianod: I see the PodDisruptionBudget objects have been added into the "Objects" methods of every component that uses them.
The "Objects" methods in the components are used by the templates code to generate the templates.

Just to clarify, in this repository we maintain two methods of deploying a 3scale API Management solution:
* The 3scale-operator, via the APIManager resource
* The OpenShift templates, that are autogenerated from source code (https://github.com/3scale/3scale-operator/blob/master/doc/development.md#building-3scale-templates)

The "Objects" methods are only used by the templates generation code at this moment. Just as a thought for possible refactorings/improvements not related to this PR, this makes me thing that maybe this methods shouldn't belong to the "component" package and probably should belong to the templates-related code only.

With the current implementation, PodDisruptionBudgets are added in all template files. This is a difference in behaviour compared to the operator.

I think we have the following possible approaches:
* Make this functionality an operator functionality only. Not include the PodDisruptionBudgets in any template (not even the HA one). To do this it would be as simple as not including the PodDisruptionBudgets objects in the "Objects" methods
* Make this functionality only available in the HA template. The best solution I can think to this is, in the templates code somehow provide a flag to make the PodDisruptionBudgets instantiated depending on the template being used. As it is done right now, this means that the component code (not template-specific code) might probably have to have a new attribute to do this, which means impact in both templates and operator code. Maybe a better solution in this case would be move the "Objects" methods from the component package to the templates one. This way probably we would not need to change code in the component package.
* Make this functionality available in all templates. As I explained, this is somehow inconsistent between how operators and templates would work, as for example the standard template (non-ha) would have a PodDisruptionBudget but the standard APIManager deployment with operator wouldn't.

What do you think @eguzki?

pkg/apis/apps/v1alpha1/apimanager_types.go
```diff
@@ -473,6 +478,11 @@ func (apimanager *APIManager) setAPIManagerCommonSpecDefaults() bool {
 		changed = true
 	}
 
+	if spec.PodDisruptionBudget == nil {
```
miguelsorianod: For other fields we also followed this approach of initializing the field explicitly when it is not initialized.

This has proved inconvenient for some use cases and currently we tend to favour the approach of not explicitly setting the spec field. The reason for that is that explicitly setting a spec field makes hard to differentiate whether the user has been the one that has set the field or it is a default set by the operator. This is a scenario that we might be interested to check as we have needed for other fields (for example the database field in the System subsection of the CR).

In case we really want to show the state of this we could add it in the "status" section of the CR.

I tend to incline to not explicitly set the default in the spec.

cc @eguzki

pkg/apis/apps/v1alpha1/apimanager_types.go
```diff
@@ -47,7 +47,8 @@ type APIManagerSpec struct {
 	// +optional
 	Zync *ZyncSpec `json:"zync,omitempty"`
 	// +optional
-	HighAvailability *HighAvailabilitySpec `json:"highAvailability,omitempty"`
+	HighAvailability    *HighAvailabilitySpec    `json:"highAvailability,omitempty"`
+	PodDisruptionBudget *PodDisruptionBudgetSpec `json:"podDisruptionBudget,omitempty"`
```
miguelsorianod: I think PodDisruptionBudget should have the // +optional tag comment in this case

wei-lee:
@miguelsorianod 

> Hi @wei-lee,
> 
> Thank you for the contribution.
> 
> I've been trying the PodDisruptionBudget and the behaviour I understand it has is the following one:
> 
> * As you commented, when draining a node, this PR allows that, when activated, it ensures that at most one pod at a time will be evicted from the node for each component. It is important to note that if the node to drain has only one pod for a component, the PodDisruptionBudget will comply the rules, as the MaxUnavailable is set to 1. This means that this functionality does not prevent a node being left without a pod of a component. What it does, is it makes the draining of the node for all of its pods components to be moved to another node one at a time. That's the intent of this PR right? and not preventing a node being left without a pod of a component.

Yes, that's right.

> * As you also commented, the PodDisruptionBudget stats are taken into account when scaling up/down a DeploymentConfig/Deployment or performing some kind of rolling update (for example when you change the PodTemplate spec and this triggers a new deployment), but they are not enforced in this case, so we should be able to perform changes into the Deployment and scale up/down without problems even if the PodDisruptionBudget stats indicate that it shouldn't be possible to evict.

That's right too.

> As I understand, apart from enabling the PodDisruptionBudget field, to really use this functionality the number of replicas of the different components have to be configured with more than one replica. Otherwise downtime avoidance is not prevented.

Yes, right again.

> We are currently reviewing the code and we'll get back to you with feedback about it.
> 
> Thank you.

Thank you for reviewing!

go.mod
wei-lee: I didn't add this myself, I believe this is generated by the `go mod` command. I can remove it manually but I think if you upgrade `golang` to newer versions, this will come back. Do you still want me to manually remove it for now?

go.mod
wei-lee: Not sure. Again I believe the `require` section of `go.mod` file are generated by the `go mod` tool. If I manually revert it it will be added back again.

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
wei-lee: I agree that we could do that, however, I am not sure we should do that, a few reasons:

1. I struggle to find a real world use case where you want to enable PodDisruptionBudgets and then disable it later. Especially with the way that we create those `PodDisruptionBudgets`- they still allow pods to be evicted, just need to be one pod at a time. And scaling up/down pods are not affected at all.
2. It will make code more complicated, for a use case that we don't know if ever going to be needed. On the other hand, we can create docs and explain that if user ever need to remove the PDBs, they can disable it in the CR and manually delete them.

Of course we can consider this again if users come back with good reasons why they need this feature, and then we can implement it. For now, I think we should keep it simple.

pkg/3scale/amp/component/backend.go
wei-lee: I am happy with whatever approach you guys suggested here. Another approach I think could work is to have a separate yaml file just for the pdbs. E.g. `amp-pdbs.yaml`. This way users can decide if they want to create PodDiscruptionBudgets by applying this file.

pkg/apis/apps/v1alpha1/apimanager_types.go
wei-lee: ok, I am happy to remove this.

pkg/apis/apps/v1alpha1/apimanager_types.go
wei-lee: will fix this.

pkg/3scale/amp/component/backend.go
eguzki: We should make operator and templates be as consistent as we can. That means, try to keep the differences when deploying OCP resources using templates and the operator to the minimum.

eguzki: "HA template" is a confusing name. That template should be called "External Databases" template. 

I think we should only allow PDB for deployments recommended for production use, i.e. "External Databases" templates and when "HA" mode is enabled in APImanager CR

wei-lee: @eguzki completely agree. I am happy to remove this from the templates for now if you guys prefer.

eguzki:
Good Job :medal_sports: 

The PR introduces lots of changes. Reviewing will continue

eguzki:
Feel free to ask anything about the code structure. Happy to answer

pkg/3scale/amp/component/backend.go
wei-lee: @eguzki Looks like we both commented at the same time. So looks like you guys prefer to only have PDBs in the `HA` template? If so I am happy to make the change.

eguzki:
No PDB has been created for `system-sphinx` deploymentconfig. Any reason not to do it or just missed it?

wei-lee:
> No PDB has been created for `system-sphinx` deploymentconfig. Any reason not to do it or just missed it?

By the look of [this code], it looks like `system-sphinx` is not scalable. That's why I didn't create the PDB for it as it won't make any differences. Should it be scalable too?

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
```diff
@@ -82,3 +83,11 @@ func (r *BaseAPIManagerLogicReconciler) updateResource(obj common.KubernetesObje
 	r.Logger().Info(fmt.Sprintf("Updated object %s", ObjectInfo(obj)))
 	return r.Client().Update(context.TODO(), obj) // don't wrap error
 }
+
+func (r *BaseAPIManagerLogicReconciler) reconcilePodDisruptionBudget(desiredPDB *v1beta1.PodDisruptionBudget) error {
+	if r.apiManager.Spec.PodDisruptionBudget != nil && r.apiManager.Spec.PodDisruptionBudget.Enabled {
```
eguzki: It would be nice to have in `pkg/apis/apps/v1alpha1/apimanager_types.go`
```
func (apimanager *APIManager) IsPDBEnabled() bool {                              
  return apimanager.Spec.PodDisruptionBudget != nil && apimanager.Spec.PodDisruptionBudget.Enabled  
}                                                                                             
```

Then use it in any place:
```
if r.apiManager.IsPDBEnabled() {
}
```

pkg/controller/apimanager/apimanager_controller.go
```diff
@@ -88,6 +89,11 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	err = c.Watch(&source.Kind{Type: &v1beta1.PodDisruptionBudget{}}, ownerHandler)
```
eguzki: What is the motivation behind watching PDB resources?

eguzki:
> @eguzki By the look of this code, it looks like system-sphinx is not scalable. That's why I didn't create the PDB for it as it won't make any differences. Should it be scalable too?

@wei-lee Fair enough. To be honest we do now even know if it is scalable. My bet is that it is actually scalable. Furthermore, the operator creates a service for it.

For this PR, we are not creating PDB for `system-sphinx`. But we have created and action point to ask system guys about it and include PDB for it in case it is scalable in some other PR.

pkg/3scale/amp/component/backend.go
wei-lee: @eguzki @miguelsorianod Per our discussion, I have updated the code to only generate PDBs for the HA template. Do you mind reviewing it again and see if you guys are happy with the approach I took. Thanks.

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
wei-lee: yeah, good idea. Will do that.

pkg/controller/apimanager/apimanager_controller.go
wei-lee: The idea is that whenever changes are made to the PDBs, it will trigger a reconcile action on the operator so that it will resync them if there are changes made by the user. 

Happy to remove it if you guys think it is unnecessary, as at the moment, pre k8s 1.16, it is not possible to update PDBs once they are created.  However, from k8s 1.16 (openshift 4.3 I believe), it will be possible to update PDBs. Generally speaking though, I think it is a good practice to watch for all the secondary resources that are managed by the operator to ensure they are kept in sync.

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
wei-lee: Resolved

pkg/controller/apimanager/apimanager_controller.go
miguelsorianod: I think it's a good idea to keep the Watch in the secondary resources so I'd keep it. That's why the other secondary resources were added in the code too (to react to changes and keep them in the desired state)

go.mod
miguelsorianod: Please remove it yes. We are not raising the minimum required Go version to 1.13 in this PR

go.mod
miguelsorianod: Does it work without modifying the go.mod file?
For what it looks like, it seems they are just updates to existing modules versions that have been applied when performing go mod commands.
If they are not needed let's not add them in the PR. We can take a look at go.mod/go.sum status of the module versions in general in another PR/issue

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
miguelsorianod: Hi @wei-lee,

Although we understand your reasons we would like to have the ability to be able to change the field:
 * It would be more intuitive and consistent for the user if he wants to change the value. We try to avoid, if possible/reasonable, fields in the CR that are "non-reconcilable"
 * There might be some case the user / we might want to temporarily disable it. If you want to make a maintenance where you want to go faster (don't mind that several pods are down at the same time), testing of functionality, reduce complexity of the deployment if you realize you don't ned it...

The idea would be that in case the flag is not enabled, if the object exists we would delete it. Unless I'm missing something I think that the functionality might be contained in the reconcilePodDisruptionBudget method.

pkg/3scale/amp/component/backend.go
miguelsorianod: Hi,

We are happy with the approach taken :+1: . We think adding configurability to the template adapters is a goo approach. In fact, we followed the same approach for a similar use case one PR that has not been merged yet (https://github.com/3scale/3scale-operator/pull/295).

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
wei-lee: @miguelsorianod fair enough. I will add support for removing the PDBs as you suggested. Thanks.

go.mod
wei-lee: ok

go.mod
wei-lee: I will try, should be possible. It's just my IDE keeps modifying it automatically.

pkg/3scale/amp/component/backend.go
eguzki: HA template implementation looks good to me :)

pkg/controller/apimanager/apimanager_controller.go
eguzki: The operator does not have any action implemented to react upon PDB resource updates. I would remove this watch, or we will have unneeded reconciliation loops.

wdty @miguelsorianod ?

eguzki:
looking soo good, congrats @wei-lee :medal_sports:

pkg/controller/apimanager/apimanager_controller.go
wei-lee: @eguzki I am updating this PR now, there will be logic to update/remove PDBs based on the suggestion from @miguelsorianod. Will ping you once it's ready.

eguzki: Ok, there is a point. I did not get it the first time. If someone updates PDB manually, the operator will update them back to desired state. Fair enough.

Thanx

miguelsorianod: @eguzki, even if we don't make specific fields configurable at CR level we are interested in keeping the desired state of our model with what currently exists in Kubernetes.

You want to react to events that happen in the resource to keep it in sync with the desired state, even if it is not configurable at CR level. It's a very common pattern among controllers.
Let's keep the watch.

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
wei-lee: @miguelsorianod I have updated the PR as suggested. Now the operator will remove the PDBs that are created by the operator if the flag is disabled. You see from [this commit](https://github.com/3scale/3scale-operator/pull/308/commits/06cfd21e2d2f24de2f71f08552a0a87820aecfec).

wei-lee:
@miguelsorianod @eguzki Thanks for the review, really appreciated it. I think I have addressed all your comments now, do you mind taking a look again and let me know if there is anything else you want to change.

Thanks a lot!

wei-lee:
Forgot to mention I have also deployed the 3scale operator to an openshift cluster and it works fine.

pkg/apis/capabilities/v1alpha1/zz_generated.openapi.go
```diff
@@ -42,6 +40,7 @@ func schema_pkg_apis_capabilities_v1alpha1_API(ref common.ReferenceCallback) com
 		Schema: spec.Schema{
 			SchemaProps: spec.SchemaProps{
 				Description: "API is the Schema for the apis API",
+				Type:        []string{"object"},
```
eguzki: there should be no changes in `pkg/apis/capabilities/v1alpha1/zz_generated.openapi.go`

My guess? you are using a diferent version of the operator-sdk. Our requirements state that:
```
operator-sdk version v0.8.0
```

There is WIP PR to upgrade operator-sdk #310 to 0.14.1

pkg/apis/capabilities/v1alpha1/zz_generated.openapi.go
```diff
@@ -1,7 +1,5 @@
 // +build !ignore_autogenerated
```
miguelsorianod: This PR does not have changes in the capabilities CRDs so this file shouldn't be commited.
I highly suspect this changes happen because you must be using a different operator-sdk version than 0.8.0 or a more recent Go version than 1.12.x

pkg/apis/apps/v1alpha1/zz_generated.openapi.go
```diff
@@ -24,6 +22,7 @@ func schema_pkg_apis_apps_v1alpha1_APIManager(ref common.ReferenceCallback) comm
 		Schema: spec.Schema{
 			SchemaProps: spec.SchemaProps{
 				Description: "APIManager is the Schema for the apimanagers API",
+				Type:        []string{"object"},
```
miguelsorianod: See comment https://github.com/3scale/3scale-operator/pull/308/files#r371898758
This specific change seems that is added because you must be using a different operator-sdk version or go version.
However, in this file there are legit changes so we must include some of them

miguelsorianod:
Functionality wise looks good to me :+1: . I've also tried to deploy it in my local cluster and trying to enable and disable PodDisruptionBudget and it works correctly too :+1:

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
miguelsorianod: looks good to me :+1:

wei-lee:
@miguelsorianod I have now reverted changes to the `go.sum` file and cleaned up the commit history.

I also regenerated the files by using 0.8.0 version of the operator-sdk by running `operator-sdk generate k8s` and `operator-sdk generate openapi`. However, the generated files are still different from master and the crd tests are failing. Do you have any idea where I did wrong? Am I using the wrong version of the operator-sdk?

miguelsorianod:
Hi @wei-lee,

The changes in the pkg/apis/apps/v1alpha1/zz_generated.deepcopy.go and pkg/apis/apps/v1alpha1/zz_generated.openapi.go files are expected and correct now. The changes are expected because are due to the addition of the PodDisruptionBudget changes so it's ok in that regard :+1: .

About the  deploy/crds/capabilities* files please exclude those changes from the commits. Those changes happens because some fields in the CRDs are not generated by the operator-sdk's openapi generator tool but we wanted to include them to offer security at deploy time.
The same happens with some of the fields in 'deploy/crds/apps_v1alpha1_apimanager_crd.yaml'. From there we only want the changes related to the PodDisruptionBudget addition. Please do not add the changes that remove some parts of the configurationSecretRef and awsSecretCredentials fields. We know this is unfortunate but we do not know of a way to include them automatically as the openapi tool does not seem to include those fields or how to make it include them (at least in the version used in operator-sdk 0.8.0)

wei-lee:
@miguelsorianod understood. Thanks for the detailed explanation. I will make the changes you suggested.

wei-lee:
@miguelsorianod ok, all issues are addressed now. Please let me know if there is anything else you need.

Thanks for the help! 👍

miguelsorianod:
Hi @wei-lee,

The final state of the changes looks good now :+1: 

Regarding the commit history, I think the commits do not correspond with the changes. On the update templates commit I do not see that the amp-ha template changes are there, which should be the only file in that commit.
I also see some changes related to go.sum and some changes we talked about in the deepcopy generated files in the commits that shouldn't be shown be added in any commit. Can you fix that? Thank you.

wei-lee:
@miguelsorianod is it ok if i just merge the 2 commits to 1? The commits are already squashed so it will be tricky to reorganise them again.

eguzki:
@wei-lee this is one of my favorite git tricks.

Merge both commits into one, and then split them as you wish. 

**Split commit in two**

https://emmanuelbernard.com/blog/2014/04/14/split-a-commit-in-two-with-git/
```
git rebase -i <oldsha1>~1
# mark the expected commit as `edit` (replace pick in front of the line), save and close
git reset HEAD^
git add ...
git commit -m "First part"
git add ...
git commit -m "Second part"
git rebase --continue
```

wei-lee:
@eguzki thanks for the trick. That's really useful. Will try that now.

wei-lee:
@miguelsorianod thanks again @eguzki for the trick, it is easier than I thought. Now it should be ok now.

PR #309, opened 2020-01-22
Deprecate amazonSimpleStorageService for simpleStorageService in fileStorage

miguelsorianod:
Now AWS S3 configuration is provided solely through a secret that contains
keys that provide the S3 specific configuration.
The system.fileStorage.amazonSimpleStorageService and its fields
have been deprecated in favour of system.fileStorage.simpleStorageService


miguelsorianod:
I've done the following to try it "live":

1. Execute the operator in my local environment with make local with the 2.7.1 version of the operator
2. Deploy 3scale with S3 using the 2.7.1 CR format (old amazonSimpleStorageService section)
3. Change to this PR branch
4. Execute oc apply on the new CRD to have the new CRD schema for the APIManager
5. Stop the operator
6. Temporarily edit 3scale operator and 3scale version strings to newer versions (to trigger an operator update)
7. Execute the operator again, this time in this branch PR to have the new changes
8. This has triggered an upgrade of the operator and I've verified the old section is removed and the new one added. I've also verified that the S3 secret has the credentials and the configuration (bucket name) there. Finally, I've also tried to upload a new logo to the developer portal and it has worked correctly, verifying that it is uploaded to the desired bucket. Also, system-environment ConfigMap does not have does fields anymore.

pkg/apis/apps/v1alpha1/apimanager_types.go
```diff
@@ -220,22 +221,30 @@ type SystemFileStorageSpec struct {
 	// +optional
 	PVC *SystemPVCSpec `json:"persistentVolumeClaim,omitempty"`
 	// +optional
-	S3 *SystemS3Spec `json:"amazonSimpleStorageService,omitempty"`
+	// Deprecated
+	DeprecatedS3 *DeprecatedSystemS3Spec `json:"amazonSimpleStorageService,omitempty"`
+	// +optional
+	S3 *SystemS3Spec `json:"simpleStorageService,omitempty"`
```
miguelsorianod: The changes here have been:

We maintain amazonSimpleStorageService json field but we rename the old S3 *SystemS3Spec field and struct type to DeprecatedS3 *DeprecatedSystemS3Spec, and we create a new field S3 and struct type SystemS3Spec with the new attribute, but using a new json name called "simpleStorageService"

pkg/3scale/amp/operator/system_reconciler.go
```diff
@@ -227,7 +227,7 @@ func (r *SystemReconciler) Reconcile() (reconcile.Result, error) {
 func (r *SystemReconciler) reconcileS3Storage() error {
 	// Nothing for reconcile.
 	// Check all required fields exist
-	awsCredentialsSecretName := r.apiManager.Spec.System.FileStorageSpec.S3.AWSCredentials.Name
+	awsCredentialsSecretName := r.apiManager.Spec.System.FileStorageSpec.S3.ConfigurationSecretRef.Name
```
miguelsorianod: System reconciler now only handles the new S3 type struct fields (as the migration should take care to make sure that is used)

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -480,10 +480,16 @@ func (u *UpgradeApiManager) upgradeSystemS3EnvVars() (reconcile.Result, error) {
 }
 
 func (u *UpgradeApiManager) upgradeAwsSecret() (reconcile.Result, error) {
+	if u.Cr.Spec.System.FileStorageSpec.DeprecatedS3 == nil {
+		// We check whether S3 section has been migrated. If it has we assume
+		// everything went ok
+		return reconcile.Result{}, nil
```
miguelsorianod: See this

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -521,13 +527,14 @@ func (u *UpgradeApiManager) upgradeAwsSecret() (reconcile.Result, error) {
...
 	changed := false
 
-	if u.Cr.Spec.System.FileStorageSpec.S3.AWSBucket != "" {
-		u.Cr.Spec.System.FileStorageSpec.S3.AWSBucket = ""
-		changed = true
-	}
-
-	if u.Cr.Spec.System.FileStorageSpec.S3.AWSRegion != "" {
-		u.Cr.Spec.System.FileStorageSpec.S3.AWSRegion = ""
+	if u.Cr.Spec.System.FileStorageSpec.DeprecatedS3 != nil {
```
miguelsorianod: We now migrate the old section content to the new section

pkg/apis/apps/v1alpha1/apimanager_types.go
```diff
@@ -519,6 +528,7 @@ func (apimanager *APIManager) setSystemFileStorageSpecDefaults() (bool, error) {
 
 	if systemSpec.FileStorageSpec != nil &&
 		systemSpec.FileStorageSpec.PVC != nil &&
+		systemSpec.FileStorageSpec.DeprecatedS3 != nil &&
```
miguelsorianod: DeprecatedS3 should never be used but during the migration it is set so we manage that case and make sure that only one of them is used

miguelsorianod:
Ready for review @eguzki

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -49,7 +49,7 @@ func (u *UpgradeApiManager) Upgrade() (reconcile.Result, error) {
 		return res, err
 	}
 
-	if u.Cr.Spec.System.FileStorageSpec != nil && u.Cr.Spec.System.FileStorageSpec.S3 != nil {
+	if u.Cr.Spec.System.FileStorageSpec != nil && (u.Cr.Spec.System.FileStorageSpec.DeprecatedS3 != nil || u.Cr.Spec.System.FileStorageSpec.S3 != nil) {
```
eguzki: Before upgrading, there is no way the user has filled *both* fields 
* `spec.system.fileStorage.amazonSimpleStorageService`
* `spec.system.fileStorage.simpleStorageService`

The upgrading procedure of S3 have some steps and the last one should be updating apimanager CR. Then you do not need to check both fields existence at the same time

pkg/3scale/amp/operator/upgrade.go
eguzki: check comment above. I think this check has no sense as has been checked before to reach `upgradeAwsSecret()`

pkg/3scale/amp/operator/upgrade.go
eguzki: see comments above. You can assume it is already not nil. As it should be the last step of the upgrade procedure

pkg/apis/apps/v1alpha1/apimanager_types.go
eguzki: `setSystemFileStorageSpecDefaults` is part of the reconciliation and you can assume it has been migrated. No need to check

pkg/3scale/amp/operator/upgrade.go
miguelsorianod: I think this might still be needed.
When the methods make changes they return that want to requeue and the exeuction stops at that point, requeuing the event and reevaluating the logic. So, the first time upgradeAwsSecret is executed DeprecatedS3 is not nil but it will be reexecuted again when it is nil, when the requeueing due to upgradeApimanagerCRS3Attrs is produced. It should be "idempotent"

pkg/3scale/amp/operator/upgrade.go
miguelsorianod: There is no way the user has filled both fields, but it might have one of the two set depending on the stage of the upgrade procedure that is being executed at one point. This is checking that either of them is set, not both of them at the same time

pkg/3scale/amp/operator/upgrade.go
miguelsorianod: Yes I was unsure of this, just added it in case something goes wrong in some of the previous steps

pkg/apis/apps/v1alpha1/apimanager_types.go
miguelsorianod: Even though it's been migrated there's nothing preventing that it is set by the user (or by the operator itself by mistake) after the migration. You do the same between checking PVC and S3. You just make sure that one of the fields in that section is set (a union).

miguelsorianod:
@eguzki left responses for your comments. Please review them.
Thank you.

pkg/apis/apps/v1alpha1/apimanager_types.go
eguzki: As I see, in reconciliation loop the operator does not care anymore on that field.

I would add a check to see that DeprecateS3 is not null and when that is the case, report some warning using logs, nothing else.

pkg/3scale/amp/operator/upgrade.go
eguzki: When updating S3 section of the CR is the last step of the upgrading S3 procedure (like it is now and as should be), this change is just allowing entering upgrading S3 process one more time and not doing anything, as everything has already been upgraded. Am I right?

miguelsorianod: You are right.
I think it's not a bad idea to allow re-entering after all the changes to ensure the current state is the desired state at all points in time, after having applied all changes, like you do with reconciliation.
I'll change it anyway to simplify the code a little bit anyway :+1:

doc/apimanager-reference.md
```diff
@@ -101,7 +101,8 @@ This resource is the resource used to deploy a 3scale API Management solution.
 | **Field** | **json/yaml field**| **Type** | **Required** | **Default value** | **Description** |
 | --- | --- | --- | --- | --- | --- |
 | PVC | `persistentVolumeClaim` | \*SystemPVCSpec | No | nil | Used to use a PersistentVolumeClaim as the System's file storage. See [SystemPVCSpec](#SystemPVCSpec) |
-| S3  | `amazonSimpleStorageService` | \*SystemS3Spec | No | nil | Used to use S3 as the System's file storage. See [SystemS3Spec](#SystemS3Spec) |
+| DeprecatedS3  | `amazonSimpleStorageService` | \*DeprecatedSystemS3Spec | No | nil | DEPRECATED [DeprecatedSystemS3Spec](#DeprecatedSystemS3Spec). Used to use S3 as the System's file storage. See [SystemS3Spec](#SystemS3Spec) |
```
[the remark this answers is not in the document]
miguelsorianod: It's not very useful. Just in the case the user wants to go see the structure. We can remove them all in another PR :+1:

pkg/3scale/amp/operator/upgrade.go
miguelsorianod: I'll remove this check as we have prevented the upgradeSystemS3 to reenter after updating the CR sections at the end

pkg/apis/apps/v1alpha1/apimanager_types.go
miguelsorianod: Added a log line in system_reconciler.go file. Please review it.

miguelsorianod:
Applied changes. Ready for review @eguzki

pkg/3scale/amp/operator/upgrade.go
eguzki: updating s3 section of the CR should be the last step. Would be good to add a comment to prevent anyone adding more steps after.

miguelsorianod: Comment added :+1:

PR #312, opened 2020-02-03
template and operator test default images

eguzki:
Template and operator test default images instead of enforcing nightly images. 

Stable branch tests were testing nightly images

### DIFF
--- a/pkg/3scale/amp/product/3scale_release.go
+++ b/pkg/3scale/amp/product/3scale_release.go
@@ -1,5 +1,5 @@
 package product
 
 var (
-	ThreescaleRelease = "2.9"
+	ThreescaleRelease = "master"
 )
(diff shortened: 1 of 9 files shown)


pkg/3scale/amp/product/3scale_release.go
```diff
@@ -1,5 +1,5 @@
 package product
 
 var (
-	ThreescaleRelease = "2.7"
```
thomasmaas: why master?

eguzki: master branch deploys nightly images, so `ThreescaleRelease = "2.7"` is not correct. It could also be "coming release". 

Basically this is needed to use default AMP_RELEASE param when deploying templates or via operator. As the tests were setting `--param AMP_RELEASE=master`.

PR #317, opened 2020-02-07
Add routes existence tests in templates deployments

miguelsorianod:
Adds a command in CircleCI that checks the existence of the default routes that should be created when deploying 3scale.
It has been added into the nightly templates deployment and the one that is executed in each PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -118,13 +118,48 @@ commands:
       - run:
           name: Deploy 3scale from amp-eval template without imagestreams
           command: |
+            WILDCARD_DOMAIN="lvh.me"
             ruby -ryaml -rjson -e 'puts YAML.load(ARGF).tap{|t| t["objects"].reject!{|o| o["kind"]=="ImageStream"}}.to_json' pkg/3scale/amp/auto-generated-templates/amp/amp-eval.yml | \
-              oc new-app -f- --param WILDCARD_DOMAIN=lvh.me --param AMP_RELEASE=master
+              oc new-app -f- --param WILDCARD_DOMAIN=${WILDCARD_DOMAIN} --param AMP_RELEASE=master
             oc wait --for=condition=available --timeout=-1s $(oc get dc --output=name)
 
             oc get events | egrep ' Failed ' || :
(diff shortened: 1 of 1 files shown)


miguelsorianod:
Yesterday it was tested that when the routes weren't created the new command didn't progress, as expected.

This can be partially seen in this job: https://circleci.com/gh/3scale/3scale-operator/9082 where you can see that it took 8 hours approximately to finish correctly: 9 hr ago (8:18:10)
The reason it ended successfully instead of failing is because CircleCI restarted everything eventually due to long execution and the new nightly images were available.
I've made sure that no output is shown when the wait for a route is being performed and it's not successful so the default no_output_timeout is triggered successfully ending the job before.

.circleci/config.yml
```diff
@@ -118,13 +118,48 @@ commands:
       - run:
           name: Deploy 3scale from amp-eval template without imagestreams
           command: |
+            WILDCARD_DOMAIN="lvh.me"
```
eguzki: why env var?

.circleci/config.yml
```diff
@@ -118,13 +118,48 @@ commands:
...
+              oc new-app -f- --param WILDCARD_DOMAIN=${WILDCARD_DOMAIN} --param AMP_RELEASE=master
             oc wait --for=condition=available --timeout=-1s $(oc get dc --output=name)
 
             oc get events | egrep ' Failed ' || :
             oc get events -o json | jq '[.items[] | select(.reason == "Failed") | debug ] | length == 0'  --exit-status
           no_output_timeout: 30m
+      - check-3scale-templates-deployed-routes
+
+  check-3scale-templates-deployed-routes:
+    parameters:
```
eguzki: any job using this wildcard_domain parameter?

miguelsorianod: I implemented it as a parameter to make the command more reusable/configurable. 
In the command itself the value of the parameter is expanded

.circleci/config.yml
miguelsorianod: Seemed more consistent with the other command that was created. Also seemed easier to modify/read instead of the middle of a long command

PR #325, opened 2020-02-17
Prepare 2.8 stable release

eguzki:


pkg/3scale/amp/manual-templates/amp/apicast.yml
```diff
@@ -121,7 +121,7 @@ parameters:
   value: "2.7.0"
```
miguelsorianod: The AMP_RELEASE value should be updated for apicast template too

eguzki:

eguzki:
@miguelsorianod new commits added. Review please

miguelsorianod:
Looks good to me but a bug in system image has been detected.
This will require updating the porta image (and its quay url). We can merge it now and then update it in another PR when the new image is available and tested by qe or we can wait and update it in this PR.

PR #330, opened 2020-02-21
Add explanation of PodDisruptionBudgets in documentation

miguelsorianod:


eguzki:
PDB was only enabled when CR setting was set AND external databases deployment config set. Isn't it? If that is true, it should be mentioned.

miguelsorianod:
PDB is enabled always even if external databaes deployment config is not set. They are independent.

PR #331, opened 2020-02-28
Upgrade operator sdk from v0.8.0 to v0.15.2

miguelsorianod:
This PR updates operator-sdk from v0.8.0 to v0.15.2.

Conceptually, the set of changes have been:

* Update all the needed dependencies
* Make the necessary changes to allow the operator code to compile
* Incorporate the new operator-sdk changes that were not code-breaking changes, but that we want to have all the functionalities that the operator-sdk v0.15.2 provide and to be in sync with the code that it generates/uses
* Regenerate CRDs, CSVs and OpenAPI
* Updated Go version to 1.13.7

To perform the previous actions the files on our project have been compared with what an operator-sdk v0.15.2 empty project with an API, a Controller, and their corresponding CRDs, OpenAPI code and CSVs are created.

Some remarkable changes:
* The PlanCost type attributes of the capabilities code have been changed from float64 to string due to float64 not supported in openapi anymore. See the corresponding commit for details. This is a breaking change that affects the "Plan" datatype and capabilities. Because capabilities are tech preview we go forward with the breaking changes.
* The default operator-sdk code to generate metrics was giving us permission errors. For what I have observed, it seems it tries to get metrics for all defined types (GVKs) in the Schema object that the operator has. This Schema object seems to include Kubernetes types that we don't use in our code (and thus we don't have permissions for it). I've implemented a function to filter the GVKs for the ones I specified.


Makefile
```diff
@@ -13,64 +13,69 @@ endif
...
+OPERATOR_SDK := operator-sdk
+GO := go
+OC := oc
+DOCKER := docker
 
 help: Makefile
 	@sed -n 's/^##//p' $<
 
 ## vendor: Populate vendor directory
 vendor:
```
eguzki: Vendoring is not required anymore, is it? 

This `vendor` target still useful?

eguzki: I mean, vendoring is required, obviously, but `vendor` directory at the root not required anymore, is it?

miguelsorianod: Go modules do not exclude vendoring.
I see the vendor tag still useful to vendor the dependencies. operator-sdk command uses the vendor directory if its required in the project directory (https://github.com/operator-framework/operator-sdk/blob/master/CHANGELOG.md#removed-4)

Makefile
```diff
@@ -13,64 +13,69 @@ endif
 OPERATORCOURIER := $(shell command -v operator-courier 2> /dev/null)
 LICENSEFINDERBINARY := $(shell command -v license_finder 2> /dev/null)
 DEPENDENCY_DECISION_FILE = $(PROJECT_PATH)/doc/dependency_decisions.yml
+OPERATOR_SDK := operator-sdk
+GO := go
```
eguzki: better 
```
GO ?= go
OPERATOR_SDK ?= operator-sdk
```

miguelsorianod: changed :+1:

PR #332, opened 2020-03-02
register openshift schemes in main

eguzki:


miguelsorianod:
What's the motivation of this?

Originally the registering of schemes was placed in the pkg/apis/addtoscheme_apps_v1alpha1.go because the types of those schemas are used on the code related to that API Group / types, and where the CRD itself is added into the schema. At the end in the main all the previously added schemas are checked.

eguzki:
Following the pattern of other operators 

https://github.com/integr8ly/grafana-operator/blob/master/cmd/manager/main.go

https://github.com/keycloak/keycloak-operator/blob/master/cmd/manager/main.go

Besides, openshift schemes does not belong to the APIManager CRD. If we ever add new CRD requiring openshift schemes, these schemes are going to be registered twice? It is small change but I think it makes all the sense

PR #333, opened 2020-03-02
Monitoring resources

eguzki:
Deploy monitoring resources.

Features:
* New APIManager CRD attribute to enable / disable
* Reconcilliate  ServiceMonitors for metric enabled components
* Reconcilliate GrafanaDashboards 
* Reconcilliate PrometheusRules 

Others:
* Embedding static files, i.e. resources, in Go using https://github.com/go-bindata/go-bindata
  * Grafana dashboards are large json files and it is very convenient to work embedding files in the binary.
* Templates with `import "text/template"`

Questions open: 
* Implementation for templates? -> NO
* If servicemonitors/grafana/prometheusrules CRD's are not installed in the cluster, currently log and continue. Other operators implement periodical checking and install resources when CRD's are available. 

Missing:
- [ ] ~Tests~
- [ ] ~components using options~
- [x] Adapt SaaS grafana dashboard spec for components in on prem deployment.
- [x] Adapt SaaS prometheus rules spec for components in on prem deployment.
- [x] Upgrading step from previous releases
- [x] upstream doc


cmd/manager/main.go
```diff
@@ -127,6 +132,18 @@ func main() {
...
 	}
 
+	// Setup Scheme for all monitoring resources
+	if err := monitoringv1.AddToScheme(mgr.GetScheme()); err != nil {
+		log.Error(err, "")
+		os.Exit(1)
+	}
+
+	// Setup Scheme for all grafana resources
+	if err := grafanav1alpha1.AddToScheme(mgr.GetScheme()); err != nil {
```
miguelsorianod, line 142: Does adding the scheme mean that in order to be able to start the operator now the grafana-related CRDs have to be there beforehand? I think in case of prometheus the CRDs are already there because OpenShift clusters install them but I think it's not the case for Grafana-related CRDs.
I understand we require the CRDs even if we are not using the functionality.

If so, I guess we should provide documentation team with steps to document the deployment of grafana-related CRDs?

pkg/3scale/amp/component/apicast_monitoring.go
```diff
@@ -0,0 +1,189 @@
...
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Service",
+			APIVersion: "v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "apicast-production-metrics",
+			Labels: map[string]string{
+				"app":                          appsv1alpha1.Default3scaleAppLabel,
+				"threescale_component":         "apicast",
+				"threescale_component_element": "production",
```
miguelsorianod, line 28: Maybe we should take into account the PR #367 for this?

pkg/3scale/amp/component/backend.go
```diff
@@ -394,6 +405,10 @@ func (backend *Backend) buildBackendWorkerEnv() []v1.EnvVar {
 	result = append(result,
 		helper.EnvVarFromSecret("CONFIG_EVENTS_HOOK", "system-events-hook", "URL"),
 		helper.EnvVarFromSecret("CONFIG_EVENTS_HOOK_SHARED_SECRET", "system-events-hook", "PASSWORD"),
+		// TODO: Add as EnvVar or reference to CM?
+		// TODO: Enable metrics endpoint only when monitoring is enabled by CR? Other components have monitoring enabled no matter monitoring field is enabled or not.
+		v1.EnvVar{Name: "CONFIG_WORKER_PROMETHEUS_METRICS_PORT", Value: BackendWorkerMetricsPortStr},
```
miguelsorianod, line 410: helper.EnvVarFromValue can be used for this

cmd/manager/main.go
eguzki: No. Adding the scheme does no require the CRD to be registered in the cluster.

Reconcilers ask the cluster if the CRD's are installed. If not, CRs are safely discarded with a log message:

```
{"level":"info","ts":1591386801.247215,"logger":"controller_apimanager","msg":"Install grafana-operator in your cluster to create grafanadashboards objects","APIManager Controller":"apimanager1","Error creating grafanadashboard object":"backend"}
```

For that kind of checks, a new discovery client is being added to the base reconciler.

Here you can see the reconcilers checking for specific CRD to be available in the cluster: 
https://github.com/3scale/3scale-operator/pull/333/files?file-filters%5B%5D=.go&file-filters%5B%5D=.tpl#diff-bd922f847ce6c805b996495ed4e5a612R126

pkg/3scale/amp/component/apicast_monitoring.go
eguzki: Metering labels only apply to pods

pkg/3scale/amp/component/backend.go
eguzki: It is already one line thing. I prefer like this because it is already clean and clear.
```
func EnvVarFromValue(name string, value string) v1.EnvVar {    
    return v1.EnvVar{                                          
        Name:  name,                                           
        Value: value,                                          
    }                                                          
}                                                              
```

pkg/3scale/amp/component/apicast_monitoring.go
miguelsorianod: That's correct. I was referring to reusing the created options in that PR for "common" labels (not the metric ones)

pkg/3scale/amp/component/backend.go
miguelsorianod: :+1:  as you prefer

pkg/3scale/amp/component/apicast_monitoring.go
eguzki: Ok, when #367 is merged, I will rebase this and apply changes. Leaving this conversation unresolved as reminder

eguzki: Done

Makefile
```diff
@@ -39,6 +39,16 @@ download:
 	@echo Download go.mod dependencies
 	@go mod download
 
+## install-tools: Installing tools from tools.go
+install-tools: download
+	@echo Installing tools from tools.go
+	@cat tools.go | grep _ | awk -F'"' '{print $$2}' | xargs -tI % go install %
```
miguelsorianod, line 45: `go install` will install the binary on user's $GOPATH/bin. If there's that same tool already installed I guess it'll overwrite any existing one (that might be another version by another project or installed by the user itself).

Do you think it might be a good idea to change the output path? (I guess for that go build has to be used instead of go install)

miguelsorianod: I've seen go-bindata seems to be a used tool. Maybe placing it outside of $GOPATH/bin would make go generate not work appropriately though

doc/operator-monitoring-resources.md
```diff
@@ -0,0 +1,51 @@
+# 3scale Monitoring Resources
+
+The 3scale monitoring resources get installed when 3scale is installed on Openshift using the 3scale Operator.
```
miguelsorianod, line 3: Should we clarify that is only installed when the appropriate setting is enabled?

miguelsorianod: I see it's specified a little bit below in `Enabling 3scale monitoring`

pkg/3scale/amp/auto-generated-templates/amp/amp-eval-s3.yml
```diff
@@ -789,6 +789,10 @@ objects:
               secretKeyRef:
                 key: password
                 name: backend-internal-api
+          - name: CONFIG_LISTENER_PROMETHEUS_METRICS_PORT
```
miguelsorianod, line 792: I think we should add this to templates upgrade procedure. New 3scale installations through templates will have this environment variables so we should be consistent between existing and new templates installation.

pkg/3scale/amp/component/system_monitoring.go
```diff
@@ -0,0 +1,111 @@
...
+		Spec: grafanav1alpha1.GrafanaDashboardSpec{
+			Json: assets.TemplateAsset("monitoring/system-grafana-dashboard-1.json.tpl", data),
+			Name: fmt.Sprintf("%s/system-grafana-dashboard-1.json", ns),
+		},
+	}
+}
+
+func SystemSidekiqPrometheusRules(ns string) *monitoringv1.PrometheusRule {
+	return &monitoringv1.PrometheusRule{
+		ObjectMeta: v12.ObjectMeta{
```
miguelsorianod, line 81: A little bit of nitpick:
```
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	v12 "k8s.io/apimachinery/pkg/apis/meta/v1"
```

Using `metav1` should be enough. If this happens in other files the same would apply.

pkg/3scale/amp/component/backend.go
```diff
@@ -387,6 +398,10 @@ func (backend *Backend) buildBackendWorkerEnv() []v1.EnvVar {
 	result = append(result,
 		helper.EnvVarFromSecret("CONFIG_EVENTS_HOOK", "system-events-hook", "URL"),
 		helper.EnvVarFromSecret("CONFIG_EVENTS_HOOK_SHARED_SECRET", "system-events-hook", "PASSWORD"),
+		// TODO: Add as EnvVar or reference to CM?
+		// TODO: Enable metrics endpoint only when monitoring is enabled by CR? Other components have monitoring enabled no matter monitoring field is enabled or not.
```
miguelsorianod, line 402: Seems good to me enabled by default (unless there's a considerably noticeable performance hit doing so). As far as I know for APIcast for example does the same.

pkg/3scale/amp/component/backend.go
```diff
@@ -387,6 +398,10 @@ func (backend *Backend) buildBackendWorkerEnv() []v1.EnvVar {
 	result = append(result,
 		helper.EnvVarFromSecret("CONFIG_EVENTS_HOOK", "system-events-hook", "URL"),
 		helper.EnvVarFromSecret("CONFIG_EVENTS_HOOK_SHARED_SECRET", "system-events-hook", "PASSWORD"),
+		// TODO: Add as EnvVar or reference to CM?
```
miguelsorianod, line 401: It's one of the usual issues that we come up with. As always, it has it's advantages and disadvantages and there doesn't seem to be a real "winner".

If you use Environment variable you don't have to worry about the ConfigMap value, querying it etc... but you have duplicated values and it's harder for the user to keep track of what's set.

If you use a ConfigMap you have to worry about the ConfigMap, but you have a centralized value for all possible references and the user can keep track of the set value for all of the objects that reference it. You also have the downside of that changing the ConfigMap value does not automatically trigger a redeployment of the elements that reference it so it might be an odd behaviour too.

pkg/apis/capabilities/v1alpha1/zz_generated.openapi.go
```diff
@@ -11,27 +11,27 @@ import (
 
 func GetOpenAPIDefinitions(ref common.ReferenceCallback) map[string]common.OpenAPIDefinition {
 	return map[string]common.OpenAPIDefinition{
-		"github.com/3scale/3scale-operator/pkg/apis/capabilities/v1alpha1.API":               schema_pkg_apis_capabilities_v1alpha1_API(ref),
```
miguelsorianod, line 14: How have you generated the openapi autogenerated files?

If I generate them I get changes reverting back to the full path.

In my case I'm using `operator-sdk generate openapi`

```
msoriano@localhost:~/go/src/github.com/3scale/3scale-operator (monitoring-tools)$ operator-sdk version
operator-sdk version: "v0.15.2", commit: "ffaf278993c8fcb00c6f527c9f20091eb8dd3352", go version: "go1.13.3 linux/amd64"
```

Makefile
```diff
@@ -39,6 +39,16 @@ download:
 	@echo Download go.mod dependencies
 	@go mod download
 
+## install-tools: Installing tools from tools.go
+install-tools: download
```
miguelsorianod, line 43: The `assets` Makefile target should have this as a prerequisite or some check verifying that go-bindata exists.

Due to I originally didn't execute install-tools I received this error:

```
msoriano@localhost:~/go/src/github.com/3scale/3scale-operator (monitoring-tools)$ make assets
Generate Go embedded assets files by processing source
go generate github.com/3scale/3scale-operator/pkg/assets
pkg/assets/assets.go:9: running "go-bindata": exec: "go-bindata": executable file not found in $PATH
make: *** [Makefile:50: assets] Error 1
```

pkg/assets/assets.go
```diff
@@ -0,0 +1,37 @@
+package assets
+
+import (
+	"bytes"
+	"text/template"
+)
+
+// Important: Run "make generate" to regenerate code after modifying/adding/removing any asset
```
miguelsorianod, line 8: Outdated information. It should be "make assets" for what I've seen in the Makefile.

pkg/assets/assets.go
```diff
@@ -0,0 +1,37 @@
...
+func SafeStringAsset(name string) string {
+	data, err := Asset(name)
+	if err != nil {
+		panic(err)
+	}
+
+	return string(data)
+}
+
+// SafeStringAsset Returns asset data as string
```
miguelsorianod, line 22: It seems it's an incorrect description (copy-paste from method above)?

Makefile
eguzki: This is regular way of working with go tools. 

If there are issues with versioning, as you highlight, customizing the GOPATH env var is the way to go

doc/operator-monitoring-resources.md
eguzki: fixed

pkg/3scale/amp/auto-generated-templates/amp/amp-eval-s3.yml
eguzki: Indeed, this has to be included in the templa upgrade procedure

pkg/3scale/amp/component/system_monitoring.go
eguzki: :+1:

pkg/3scale/amp/component/backend.go
eguzki: :+1: I will remove comments, then.

pkg/3scale/amp/component/backend.go
eguzki: If there is no clear winner, let them be env vars then. KISS.

pkg/apis/capabilities/v1alpha1/zz_generated.openapi.go
eguzki: Same  operator-sdk version

```
$ operator-sdk version
operator-sdk version: "v0.15.2", commit: "ffaf278993c8fcb00c6f527c9f20091eb8dd3352", go version: "go1.13.3 linux/amd64"
```

I tried again and it keeps the changes. They are not reverted back as in your case. My guess, this is because my workspace is not included in GOPATH, (go modules new way of working) and in your case, the workspace is included in GOPATH, hence the module name has to be included. 

TBH, I do not even know what is the use of `operator-sdk generate openapi`. CRD's and openapi validation is generated by `operator-sdk generate crds`. What is the use of `pkg/apis/capabilities/v1alpha1/zz_generated.openapi.go`?

Makefile
eguzki: :+1:

pkg/assets/assets.go
eguzki: :+1:

pkg/assets/assets.go
eguzki: :+1:

eguzki:
Some fixes have been included in https://github.com/3scale/3scale-operator/pull/401

eguzki:
No need to review templates.

pkg/apis/capabilities/v1alpha1/zz_generated.openapi.go
miguelsorianod: For what I've seen, it contains the Go representation of the validation block. Is the Go Kind type's OpenAPI model. It seems it's useful because it allows you to create a full OpenAPI Specification and generate a client from it.

Makefile
miguelsorianod: :+1:

pkg/apis/capabilities/v1alpha1/zz_generated.openapi.go
eguzki: We don't use it in any way, right?

eguzki:
Upgrade path is still missing. Otherwise when upgraded 3scale enabled monitoring, some components may not reconcile env vars to enable metrics

pkg/3scale/amp/component/apicast_monitoring.go
```diff
@@ -0,0 +1,198 @@
...
+		Spec: grafanav1alpha1.GrafanaDashboardSpec{
+			Json: assets.TemplateAsset("monitoring/apicast-grafana-dashboard-2.json.tpl", data),
+			Name: fmt.Sprintf("%s/apicast-grafana-dashboard-2.json", ns),
+		},
+	}
+}
+
+func ApicastPrometheusRules(ns string) *monitoringv1.PrometheusRule {
+	return &monitoringv1.PrometheusRule{
+		ObjectMeta: v12.ObjectMeta{
```
miguelsorianod, line 144: Same issue as in comment https://github.com/3scale/3scale-operator/pull/333#discussion_r443666233. I've also observed it in other *_monitoring.go files

doc/operator-monitoring-resources.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+## Monitored components
+
+* Kubernetes resources at namespace level where 3scale is installed
+* Apicast Staging
+* Apicast Production
+* 3scale Backend worker
+* 3scale Backend listener
+* System sidekiq
```
miguelsorianod, line 37: Are there environment variables for system-sidekiq? or is always enabled directly by code?

Are there environment variables for zync and/or zync-que? or is always enabled directly by code?

It seems it works differently than backend and apicast

miguelsorianod: :+1:

PR #338, opened 2020-03-18
Fix apicast 3scale portal endpoint

eguzki:
Simplifying `system-master-apicast` secret and apicast options structure. Master base Url not required.

- [x] upgrade procedure

### DIFF
--- a/pkg/3scale/amp/operator/upgrade.go
+++ b/pkg/3scale/amp/operator/upgrade.go
@@ -5,11 +5,13 @@ import (
 	"fmt"
 	"reflect"
 
+	"github.com/3scale/3scale-operator/pkg/3scale/amp/component"
 	appsv1alpha1 "github.com/3scale/3scale-operator/pkg/apis/apps/v1alpha1"
 	"github.com/go-logr/logr"
 	"github.com/google/go-cmp/cmp"
 	appsv1 "github.com/openshift/api/apps/v1"
 	imagev1 "github.com/openshift/api/image/v1"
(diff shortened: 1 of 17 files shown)


miguelsorianod:
Looks good.

If I'm not missing anything the upgrade procedure should remove the obsolete entry from the `system-master-apicast` secret and nothing else to do (think about how you'd do it in templates too even if not writing the procedure at the moment).

eguzki:
@miguelsorianod PR ready for review

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -581,3 +589,30 @@ func (u *UpgradeApiManager) findTagReference(tagRefName string, tagRefs []imagev
...
+		Name:      component.SystemSecretSystemMasterApicastSecretName,
+		Namespace: u.Cr.Namespace,
+	}
+	err := u.Client.Get(context.TODO(), secretNamespacedName, existing)
+	// NotFound also regarded as error, as secret is expected to exist
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if _, ok := existing.Data["BASE_URL"]; ok {
```
miguelsorianod: JSON patch is not really needed in this case.

In Go you can delete a map key. Go allows you to use:
```
delete(mymap, mymapkey)
```
The delete function doesn't return anything, and will do nothing if the specified key doesn't exist.

So you can check if the key exists or not and if it does delete it and update the object.

If you still want to use JSON patch I'm ok with it too as it does the work.

PR #340, opened 2020-03-20
Fix CI jobs that create the ImageStreams separately

miguelsorianod:
Since PR https://github.com/3scale/3scale-operator/pull/292 ImageStreams are no longer created with the "latest" tag but with the version-specific tag.
CircleCI tasks that relied on ImageStream building via the build.yml template to create the ImageStreams independently were always being created with "latest" tags.

This PR makes the name of the ImageStream tags of the 3scale components configurable in the build.yml and updates CircleCI jobs that rely on them to use the version-specific tag. In the case of this PR, "master".

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -178,6 +178,9 @@ commands:
           no_output_timeout: 30m
       - check-3scale-templates-deployed-routes
   push-3scale-images-to-quay:
+    parameters:
+      components_imagestream_tag_name:
+        type: string
     steps:
       - docker-login
       - run:
@@ -185,11 +188,12 @@ commands:
           command: |
             oc whoami --show-token | docker login -u $(oc whoami) --password-stdin 172.30.1.1:5000
             project=$(oc project -q)
+            imagestream_tag_name=<< parameters.components_imagestream_tag_name >>
 
             oc image mirror $(for component in apicast zync ; do
-              echo 172.30.1.1:5000/$project/amp-$component:latest=quay.io/3scale/$component:nightly
-            done) 172.30.1.1:5000/$project/amp-backend:latest=quay.io/3scale/apisonator:nightly \
-            172.30.1.1:5000/$project/amp-system:latest=quay.io/3scale/porta:nightly --insecure
+              echo 172.30.1.1:5000/$project/amp-$component:${imagestream_tag_name}=quay.io/3scale/$component:nightly
+            done) 172.30.1.1:5000/$project/amp-backend:${imagestream_tag_name}=quay.io/3scale/apisonator:nightly \
+            172.30.1.1:5000/$project/amp-system:${imagestream_tag_name}=quay.io/3scale/porta:nightly --insecure
 
   create-secrets:
     steps:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -178,18 +178,22 @@ commands:
...
+      components_imagestream_tag_name:
+        type: string
     steps:
       - docker-login
       - run:
           name: Push nightly images to quay.io
           command: |
             oc whoami --show-token | docker login -u $(oc whoami) --password-stdin 172.30.1.1:5000
             project=$(oc project -q)
+            imagestream_tag_name=<< parameters.components_imagestream_tag_name >>
```
eguzki: why local env var?

.circleci/config.yml
```diff
@@ -475,30 +488,11 @@ jobs:
 
 workflows:
   version: 2
-  operator:
```
eguzki: Do not remove operator workflow

.circleci/config.yml
```diff
@@ -507,24 +501,3 @@ workflows:
           requires:
             - install-operator
             - build-push-3scale-nightly-images
-    <<: *nightly-trigger
-  operator-release:
```
eguzki: Do not remove operator-release workflow

.circleci/config.yml
miguelsorianod: If you see the commit title related to that, it is just for testing purposes

.circleci/config.yml
miguelsorianod: see my comment above

.circleci/config.yml
miguelsorianod: because I prefer not to keep referencing the paramter the way `<< parameters.components_imagestream_tag_name >>`
on several lines

miguelsorianod:
Nightly workflow CircleCI job https://circleci.com/workflow-run/0775956f-4233-4cf8-9f0f-69ba43408369 with a correct execution with the changes.
Now I'm going to revert the temporary commit and let execute all the non-nightly workflows.

miguelsorianod:
Tests pass now. Ready @eguzki

.circleci/config.yml
eguzki: why not? it is clear

.circleci/config.yml
eguzki: Remove the commit, open new branch out of this branch, apply the commit and test it.

.circleci/config.yml
miguelsorianod: I don't want to repeat the reference in several places of the code (reuse the value) and I don't want to limit references to circlec-ci specific syntax as much as possible

.circleci/config.yml
miguelsorianod: It's tested  https://circleci.com/workflow-run/0775956f-4233-4cf8-9f0f-69ba43408369  (nightly only) and current state is without the nightly execution

.circleci/config.yml
eguzki: :woman_shrugging:

PR #352, opened 2020-04-17
Refactor templates-related code into templates package

miguelsorianod:
This PR refactors the following templates-related code into the templates package:
* Objects list created by each component in the templates
* Post-Processing methods (which only apply to templates)

### DIFF
--- a/pkg/3scale/amp/component/s3.go
+++ b/pkg/3scale/amp/component/s3.go
@@ -2,6 +2,7 @@ package component
 
 import (
 	"github.com/3scale/3scale-operator/pkg/common"
+	"github.com/3scale/3scale-operator/pkg/helper"
 
 	appsv1 "github.com/openshift/api/apps/v1"
 	templatev1 "github.com/openshift/api/template/v1"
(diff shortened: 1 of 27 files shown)


miguelsorianod:
Additionally, for a future PR I've observed we might take a look at completely removing s3 and ha components themselves in some way

pkg/3scale/amp/component/s3.go
```diff
@@ -27,143 +23,9 @@ func NewS3(options *S3Options) *S3 {
...
-				res = append(res[:idx], res[idx+1:]...) // This deletes the element in the array
-				break
-			}
-		}
-	}
-
-	return res
-}
-
+// TODO Template-only object. Decide what to do with this component.
```
eguzki, line 26: why don't you move it to templates package?

miguelsorianod: I thought about it but I left it for a future PR where we might decide to remove it completely. That's why I left the TODO there.
We still need some parts of it in the component package anyway (constants) so I preferred to leave it this way at the moment (same applies with the s3_options file).

PR #356, opened 2020-04-20
Generic reconciler refactor

eguzki:
Refactor all reconcilers to use same base reconciler.

Remove all duplicated reconcilers.

- [x] unittests
- [x] enhancement in logging


eguzki:
`+855 LOC added`
`−2,131 LOC added`

That was also one of the objectives. Remove redundancy and duplicated code.

eguzki:
@miguelsorianod ready for initial review. 

Unittests will be fixed and refactored when the implementation is accepted.

pkg/reconcilers/base_reconciler.go
```diff
@@ -0,0 +1,133 @@
...
+
+// MutateFn is a function which mutates the existing object into it's desired state.
+type MutateFn func(existing, desired common.KubernetesObject) (bool, error)
+
+func CreateOnlyMutator(existing, desired common.KubernetesObject) (bool, error) {
+	return false, nil
+}
+
+type BaseReconciler struct {
+	mgr             manager.Manager
```
miguelsorianod, line 28: We used to pass the client and scheme directly. Is it a good idea to pass the whole manager dependency?
When doing tests how are we going to mock this? When we passed the client we could use the FakeClient and pass it.

eguzki: I was thinking about mocking Manager with fake client. 

But, I agree. No big benefit. Will add client and scheme as fields.

pkg/reconcilers/base_reconciler.go
```diff
@@ -0,0 +1,137 @@
...
+
+func CreateOnlyMutator(existing, desired common.KubernetesObject) (bool, error) {
+	return false, nil
+}
+
+type BaseReconciler struct {
+	client          client.Client
+	scheme          *runtime.Scheme
+	apiClientReader client.Reader
+	ctx             context.Context
```
miguelsorianod, line 30: I'm curious about this. What's the usefulness on being able to pass a context.Context object and not keep using the context.TODO directly? Do we have any use case?

pkg/reconcilers/base_reconciler.go
```diff
@@ -0,0 +1,137 @@
...
+
+	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+var log = logf.Log.WithName("reconcilers")
```
miguelsorianod, line 17: What's the advantage of creating a logger here as a package-level variable compared to receiving the one from the controller?
Does it also have some impact when dealing with testing? (possible mocking of the log etc...)

pkg/controller/apimanager/apimanager_controller.go
```diff
@@ -104,7 +104,7 @@ var _ reconcile.Reconciler = &ReconcileAPIManager{}
 
 // ReconcileAPIManager reconciles a APIManager object
 type ReconcileAPIManager struct {
-	operator.BaseControllerReconciler
+	*reconcilers.BaseReconciler
```
miguelsorianod, line 107: Is there some motivation on passing a pointer instead of a value?

pkg/reconcilers/base_reconciler.go
eguzki: The context allows some features, like cancelling operations. Thus, the same context instance should be used in all operations. We are not using them, but this is preparation for that. 

It is commonly used pattern in operators.

pkg/reconcilers/base_reconciler.go
eguzki: Yes, this is a mistake. I will refactor a little bit.

pkg/controller/apimanager/apimanager_controller.go
eguzki: Avoid unnecessary copy of the structures.

pkg/reconcilers/base_reconciler.go
miguelsorianod: :+1: Do you have any example of an operator making use of it? It seems interesting in that case

eguzki: for instance, Keycloak:
https://github.com/keycloak/keycloak-operator/blob/master/pkg/controller/keycloak/keycloak_controller.go#L53

eguzki:
ready for more reviews @miguelsorianod 

Shall I start implementing unittests?

miguelsorianod:
No let me review first so you don't have to potentially change unit tests.

I'll review it as soon as possible

pkg/controller/apimanager/apimanager_controller.go
```diff
@@ -28,7 +26,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
-var log = logf.Log.WithName("controller_apimanager")
+var log = logf.Log.WithName("apimanager.controller")
```
miguelsorianod, line 29: The naming scheme seems to be different than what operator-sdk generates for controller log names, which is of the form "controller_<controllername>". We might find it's different on what we might find in most custom controllers around there (for example https://github.com/keycloak/keycloak-operator/blob/bae3b1e49a1a8c8d4008da9735f1db7cb7ea40f5/pkg/controller/keycloak/keycloak_controller.go#L38)

Not

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
```diff
@@ -1,42 +1,43 @@
...
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
-	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 type BaseAPIManagerLogicReconciler struct {
-	BaseLogicReconciler
+	*reconcilers.BaseReconciler
 	apiManager *appsv1alpha1.APIManager
+	logger     logr.Logger
```
miguelsorianod, line 25: It seems we use this logger now instead of the one contained in reconcilers.BaseReconciler, although we need to create one for reconcilers.BaseReconciler.

What's the idea of having two loggers? Are we assuming the BaseReconciler will be shared among N different kinds of "logic reconcilers" ?

pkg/reconcilers/base_reconciler.go
miguelsorianod, line 17: Should this still exist?

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
```diff
@@ -65,31 +66,86 @@ func (r BaseAPIManagerLogicReconciler) ensureOwnerReference(obj common.Kubernete
...
+
+func (r *BaseAPIManagerLogicReconciler) ReconcileResource(obj, desired common.KubernetesObject, mutatefn reconcilers.MutateFn) error {
+	desired.SetNamespace(r.apiManager.GetNamespace())
+	if err := r.setOwnerReference(desired); err != nil {
 		return err
 	}
 
-	r.Logger().Info(fmt.Sprintf("Updated object %s", ObjectInfo(obj)))
-	return r.Client().Update(context.TODO(), obj) // don't wrap error
+	return r.BaseReconciler.ReconcileResource(obj, desired, r.APIManagerMutator(mutatefn))
```
miguelsorianod, line 122: If an object processed by APIManager cannot follow what is done by APIManagerMutator, for example, we don't want to add ownerReference to a specific object (because we want to keep it after deletion of the APIManager object itself), what would be the solution for it? Would the idea be to not use the ReconcileResource method and create a custom one for that specific use case?

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
```diff
@@ -65,31 +66,86 @@ func (r BaseAPIManagerLogicReconciler) ensureOwnerReference(obj common.Kubernete
...
-	if err := r.setOwnerReference(obj); err != nil {
-		return err
+func (r *BaseAPIManagerLogicReconciler) ReconcilePodDisruptionBudget(desired *v1beta1.PodDisruptionBudget, mutatefn reconcilers.MutateFn) error {
+	if !r.apiManager.IsPDBEnabled() {
+		common.TagObjectToDelete(desired)
 	}
+	return r.ReconcileResource(&v1beta1.PodDisruptionBudget{}, desired, mutatefn)
+}
+
+func (r *BaseAPIManagerLogicReconciler) ReconcileImagestream(desired *imagev1.ImageStream, mutatefn reconcilers.MutateFn) error {
```
miguelsorianod, line 76: Maybe it would be a good idea to return the existing object on the reconcile methods?

The existing object might be useful in some situations outside the ReconcileXX method itself.

For example, in a Job object, we want to make sure the job is successfully created, and do something with it after that when exists. In this case we use the resulting existing object to do some stuff with it later:
```
	err := r.ReconcileResource(existing, desired, operator.CreateOnlyMutator)
	if err != nil {
	}
	if existing.Status.Succeeded != *desired.Spec.Completions {
          ... // Do other stuff
        }
```

miguelsorianod: Another usage is when you want to perform something just after creating the object. In that case existing and desired should be equal but only the spec part, not the status part. Using the returned 'status' section might be sometimes used and that is not included in the desired object even if the object is being created for the first time

pkg/controller/apimanager/apimanager_controller.go
```diff
@@ -138,7 +132,7 @@ func (r *ReconcileAPIManager) upgradeAPIManager(cr *appsv1alpha1.APIManager) (re
 // The Controller will requeue the Request to be processed again if the returned error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *ReconcileAPIManager) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	logger := r.Logger().WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	logger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
```
miguelsorianod, line 135: Why is log instead of r.Logger() being used?
r *ReconcileAPIManager is a BaseReconciler, and has been instantiated with the `log` variable

pkg/helper/object_meta_merger.go
```diff
@@ -1,18 +1,28 @@
...
 // EnsureObjectMeta ensure Labels, Annotations
-func EnsureObjectMeta(existing, desired *metav1.ObjectMeta) bool {
+func EnsureObjectMeta(existing, desired common.KubernetesObject) bool {
 	updated := false
 
-	MergeMapStringString(&updated, &existing.Labels, desired.Labels)
-	MergeMapStringString(&updated, &existing.Annotations, desired.Annotations)
+	existingLabels := existing.GetLabels()
+	if existingLabels == nil {
+		existingLabels = map[string]string{}
```
miguelsorianod, line 14: I think the new code might not be needed.

If you look at the MergeMapStringString, the initialization and set is performed (it's doing a merge) there.

pkg/reconcilers/deploymentconfig.go
```diff
@@ -0,0 +1,89 @@
...
+
+	if desired.Spec.Replicas != existing.Spec.Replicas {
+		existing.Spec.Replicas = desired.Spec.Replicas
+		update = true
+	}
+
+	return update
+}
+
+func DeploymentConfigContainerResourcesReconciler(desired, existing *appsv1.DeploymentConfig) bool {
```
miguelsorianod, line 64: What's the difference between a "Reconciler" and a "Mutator" in this case? why haven't we called this "DeploymentConfigContainerResourcesMutator"?

pkg/reconcilers/imagestreams.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/3scale/3scale-operator/pkg/common"
+	imagev1 "github.com/openshift/api/image/v1"
+)
+
+func GenericImagestreamMutator(existingObj, desiredObj common.KubernetesObject) (bool, error) {
```
miguelsorianod, line 11: nitpick. `GenericImagestreamMutator``should be better called `GenericImageStreamMutator` (Uppercase S in stream)

pkg/controller/apimanager/apimanager_controller.go
eguzki:
```{"level":"info","ts":1588004166.3609462,"logger":"kubebuilder.controller","msg":"Starting EventSource","controller":"apimanager-controller","source":"kind source: /, Kind="}
{"level":"info","ts":1588004166.361117,"logger":"kubebuilder.controller","msg":"Starting EventSource","controller":"apimanager-controller","source":"kind source: /, Kind="}
{"level":"info","ts":1588004166.3612475,"logger":"kubebuilder.controller","msg":"Starting EventSource","controller":"apimanager-controller","source":"kind source: /, Kind="}
{"level":"info","ts":1588004166.3614237,"logger":"kubebuilder.controller","msg":"Starting EventSource","controller":"binding-controller","source":"kind source: /, Kind="}
{"level":"info","ts":1588004166.3615575,"logger":"kubebuilder.controller","msg":"Starting EventSource","controller":"binding-controller","source":"kind source: /, Kind="}
{"level":"info","ts":1588004166.3616793,"logger":"kubebuilder.controller","msg":"Starting EventSource","controller":"binding-controller","source":"kind source: /, Kind="}
{"level":"info","ts":1588004166.3618264,"logger":"kubebuilder.controller","msg":"Starting EventSource","controller":"binding-controller","source":"kind source: /, Kind="}
{"level":"info","ts":1588004166.361956,"logger":"kubebuilder.controller","msg":"Starting EventSource","controller":"binding-controller","source":"kind source: /, Kind="}
{"level":"info","ts":1588004166.362078,"logger":"kubebuilder.controller","msg":"Starting EventSource","controller":"binding-controller","source":"kind source: /, Kind="}
{"level":"info","ts":1588004166.3622398,"logger":"kubebuilder.controller","msg":"Starting EventSource","controller":"tenant-controller","source":"kind source: /, Kind="}
```
Which one do you want to use? Up to you.

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
eguzki: The idea is that the logger at `BaseAPIManagerLogicReconciler` contains named value `APIManager Controller=apiManager.Name`. Because `BaseAPIManagerLogicReconciler` life cycle is just one reconciliation loop and has information about APImanager CR. 

On the other hand, BaseReconciler instance life cycle spans to the life of the operator's process.  And its logger is generic about `apimanager.controller` controller.  

[Logr](https://godoc.org/github.com/go-logr/logr) interface allows creating new logger objects out of other more generic logger objects calling `WithValues` or `WithName` methods. And this is used in BaseAPIManagerLogicReconciler constructor:
```
func NewBaseAPIManagerLogicReconciler(b *reconcilers.BaseReconciler, apiManager *appsv1alpha1.APIManager) *BaseAPIManagerLogicReconciler {
    return &BaseAPIManagerLogicReconciler{                                                          
        BaseReconciler: b,                                                                          
        apiManager:     apiManager,                                                                 
        logger:         b.Logger().WithValues("APIManager Controller", apiManager.Name),            
    }                                                                                               
}                                                                                                   
```

pkg/reconcilers/base_reconciler.go
eguzki: It is being used by public functions of the package. Those are not methods with `BaseController` type receiver

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
eguzki: Exactly! 

It is a pity we do not have an example, as all object processed by the apimanager controller receive labels and ensure ownership. 

If there was such an example, it would be using directly  `BaseReconciler.ReconcileResource` method, like this:
```
r.BaseReconciler.ReconcileResource(obj, desired, myMutator)
```

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
eguzki: I think that status will be eventually updated and you will not have this info in the "existing" object. You will need to read it from the cluster again to be sure. 

Anyway, `BaseAPIManagerLogicReconciler.ReconcileImagestream` does not need to return existing object in anyway.

If you ever need existing object after reconciliation, using the base method, 
```
r.BaseReconciler.ReconcileResource(existingObj, desired, myMutator)
```
the existing object, after the call, will be the result of the reconciliation.

pkg/controller/apimanager/apimanager_controller.go
eguzki: Right. Upodating it.

pkg/controller/apimanager/apimanager_controller.go
miguelsorianod: The one in the logs you've shown me is not a controller generated by operator-sdk, which is what I was referring to.

I personally prefer the one with underscore as it's how operator-sdk generates it for you, but both are fine to me so I'll be ok with any decision you make

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
miguelsorianod: I see. Makes sense :+1:

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
miguelsorianod: :+1:

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
miguelsorianod: > I think that status will be eventually updated and you will not have this info in the "existing" object. You will need to read it from the cluster again to be sure.

That depends on the object type. It's often the case you say. For some object types use the returned object in the same reconciliation loop iteration it's often useful and you might have the information already there.

You can avoid performing a Get just after performing a ReconcileImageStream call in the same reconcile loop call. You also will avoid eventual consistency caching in that iteration etc... . That's one of the reasons why the create method returns the object.

miguelsorianod: In any case I guess we can leave it this way and if we see we often resort to the situation I explained we can reevaluate it

pkg/helper/object_meta_merger.go
eguzki: :+1:

pkg/reconcilers/deploymentconfig.go
eguzki: Mutators functions must belong to the type `MutateFn`

```
type MutateFn func(existing, desired common.KubernetesObject) (bool, error)
```

pkg/reconcilers/imagestreams.go
eguzki: :+1:

pkg/controller/apimanager/apimanager_controller.go
eguzki: :+1: let it be `controller_apimanager`

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
eguzki: :+1: As I said, you can always use `r.BaseReconciler.ReconcileResource(existingObj, desired, myMutator)`, and there you have required info.

pkg/reconcilers/base_reconciler.go
miguelsorianod: :+1: it confused me seeing it in the base_reconciler.go file

pkg/reconcilers/deploymentconfig.go
miguelsorianod: I understand that, I was just wondering what criteria do you use to decide whether you want to make it a MutateFn directly or not. Why not making it a MutateFn function?

eguzki: Could have the same signature. But then you would have to type assert from `common.KubernetesObject` to `*appsv1.DeploymentConfig`.  Maybe they should be private functions of the package?

miguelsorianod: I guess it does not really matter that much. If they are not going to be used outside the package there's no need to export them. Up to you I'm fine with both approaches.

eguzki: Making them private, then.

eguzki: They cannot be private. Some of them, for instance `DeploymentConfigResourcesMutator`, are being used by mutators not included in this package.

eguzki:
All feedback has been resolved

eguzki:
```
+1,733 LOC added
−5,119 LOC removed
```

eguzki:
PR ready for review and merge to master @miguelsorianod

pkg/helper/object_meta_merger.go
```diff
@@ -1,18 +1,22 @@
...
 
-	MergeMapStringString(&updated, &existing.Labels, desired.Labels)
-	MergeMapStringString(&updated, &existing.Annotations, desired.Annotations)
+	existingLabels := existing.GetLabels()
+	existingAnnotations := existing.GetAnnotations()
+
+	MergeMapStringString(&updated, &existingLabels, desired.GetLabels())
+	MergeMapStringString(&updated, &existingAnnotations, desired.GetAnnotations())
+
+	existing.SetLabels(existingLabels)
```
miguelsorianod, line 18: I think the changes in the body of this method are not needed.

If you look at the MergeMapStringString, the initialization and set is performed (it's doing a merge) there.

See https://github.com/3scale/3scale-operator/pull/356#discussion_r416040478 where this was commented

pkg/controller/apimanager/apimanager_controller_test.go
```diff
@@ -74,22 +73,14 @@ func TestAPIManagerControllerCreate(t *testing.T) {
...
+	for i := 0; i < 100 && !endLoop; i++ {
+		res, err := r.Reconcile(req)
+		if err != nil {
+			t.Fatal(err)
+		}
 
-	if res.Requeue {
-		t.Error("reconcile did not finish end of reconciliation as expected. APIManager should have been reconciled at this point")
+		endLoop = !res.Requeue
 	}
```
miguelsorianod, line 84: Should we check the value of endLoop after the loop? In the case endLoop is false after the loop it would mean a reconcile requeue has been received even if the contents we check in the test seem correct

pkg/helper/object_meta_merger.go
eguzki: Note that the function signature has been changed. The generic reconciler is based on `common.KubernetesObject` interface. That interface does not implement `metav1.ObjectMeta`. You no longer have direct access to labels and annotations. Instead, you have getters/setters methods for that.

pkg/controller/apimanager/apimanager_controller_test.go
eguzki: good catch.

pkg/helper/object_meta_merger.go
miguelsorianod: I noticed the signature change. I was not aware of the common.KubernetesObject interface not implementing metav1.ObjectMeta :+1:

PR #357, opened 2020-04-23
Capabilities v2

eguzki:
Reconcile cluster k8s resources with 3scale using the 3scale API.

Include APIaaP objects

Some considerations for the implementation:
* Some types have owner and cannot be shared. For instance Products. Product belong to one account provider and cannot be shared between account providers. Same for limits. Limits are owned by some application plan and cannot be shared by multiple application plans.
* Objects can be invalid. Invalid objects will be marked as such, updated status with error message and will be filtered out by operator controllers. User should delete them and do not try to create new objects with the same spec. Invalidating an object A will trigger invalidation cascade to all objects with references to the object A. The alternative to invalidate objects is, object removal. Invalidating objects approach lets the user know what happened.
  * For instance, ActiveDocs belonging to a account provider and product, and those account provider and product are not related.
* Backend usage: 1 product cannot be using same backend multiple times. Unique key is composition of <product_id, backend_id>.
  * Proposed solution: In the spec of the product, there is a map called backend usages with the key being backend_id and the value being the backend usage object reference. This way it is ensured 1 product not using same backend multiple times.
* Deletion cascade: Implement using ownerreferences? Ownerreferences does not work for Applicationa. Application object has two owners, application plan and account. If any of those owners are deleted, application should be deleted as well (orphan object). K8s garbage collector works only when ALL the owners have been deleted. Hibrid solution? ownerreferences for simple relations, operator based deletetion cascade for complex relations like the applications?
* Orphan objects: when some controller finds an orphan object (some referenced object does not exist), it will wait for the referenced object to exist. The controller will start retry with exponential back-off pattern until the referenced object is found (which could never happen). This way it is allowed temporary inconsistency, like for instance create product resource before backend resource. Product object will be ofphan until backend is found.
* Objects lifecycle
  * Based on status.Conditions k8s common pattern
  * states:
    * InSync: Sync process going on
    * Orphan: Temporarily invalid. Waiting for some resource to exist.
    * Synced: Sync'ed
    * Invalid: Invalid object. Spec should be changed.
* Implement 3scale state cache in status? Avoid too many calls to 3scale API for every reconciliation loop. Assume 3scale state (read from 3scale API endpoints) does not change if 3scale operator does not change it (using 3scale API delete, put, post methods)
* If Account Provider object is deleted, the products, backends, accounts, activedocs will not be deleted from 3scale. It is just considered "disconnected". Current PoC implements deletion of all services from 3scale when "binding" object is deleted (finalizers are used for that regard)
* 3scale API does not allow deleting a backend if it is being used by a Product. This constraint is not easy to implement as k8s resources. If the user deletes a backend. All products using that deleted backend will be marked as orphan and will not sync until reference is found or backend usage is deleted.
* 3scale does not allow updating the ownership of some object types. For instance an application plan owned by a product cannot be changed and owned by another product. This kind of relations will be implemented by defining objects inside owner spec instead of having their own CRD. For instance application plans will not have their own CRD.
  * Detailed explanation: When there is a relation 1:N from A -> B, A instance owns multiple B instances. The owner of those B instance cannot be changed to some other C object. It is required to enforce ownership as fixed, not changeable. The reconciliation logic for the scenario where ownership can be updated is very complex. For example: Product A owns several Metrics and Application plans. One of these plans, plan P, owns a limit, L, related to some metric M. If application plan P's owner is changed to product B the entire system will be inconsistent. As the application plan P would be owned by B and it would own limits with metrics owned by a different product A. This scenario is not allowed in 3scale.
* Relations 1:N will be implemented using one of the following options, depending on each relationship type:
  * B instance contains a reference to A. This way you make sure B instance is not owned by multiple A's. But this solution has a big weakness. There is nothing like relational integrity, that prevents the user to update the reference hold by B to A and change to a reference to C or even something not existing. Any change has to be validated by the controller. Example: Product 1 -> N ActriveDocs
  * A maintains a list/map of B references. Then 1:N relation has to be validated by the controller making sure B instance is not owned by several A instances. Example: Product 1 -> N Backend (backend usage relationship)
   * A maintaing a list/map of B objects as subresources. B object type will not have their own CRD. Ownership is fixed. Example: Product 1 -> N Application plans
* 3scale entities are linked to a Provider Account. This provider account will be referenced by the credentials secret to sync with 3scale using REST API. Each object will need provider account to sync. The look up process can be as follows:
    * If provider_account_reference attribute is found, referenced provider account will be used.
    * If no provider_account_reference attribute is found, some secret with hardcoded name will be looked up in the namespace. For example, `default_provider_account` secret. If found, that one will be used.
    * If no provider_account_reference attribute is found AND provider account default secret in the namespace is not found either, then, 3scale default provider account (3scale-admin) will be looked up using system-seed secret in the current namespace.
    * If nothing is successfully found, the object will be marked as orphan.

Design proposal:

Backend
```
name: Operated Backend A                                         
privateBaseURL: https://example.com:8443                         
systemName: OperatedBackendA                                     
metrics:
  hits:                                                          
    description: Number of API hits                              
    friendlyName: Hits                                           
    unit: "hit"                                                  
methods:                                                         
  method01:                                                      
    friendlyName: Method01                                       
mappingRules:                                                    
  - httpMethod: GET                                              
    pattern: "/pets"                                             
    metricMethodRef: metric02                                    
  - httpMethod: POST                                             
    pattern: "/pets"                                             
    metricMethodRef: hits 
```
Product
```
provider_account_reference: REF
deployment:                                                 
  apicastSelfManaged:                                       
    stagingPublicBaseURL: "https://staging2.example.com"    
    productionPublicBaseURL: "https://prod2.example.com"    
    authentication:                                         
      appKeyAppID:  {}          
metrics:                                                            
  hits:                                   
    description: Number of API hits       
    friendlyName: Hits
    unit: "hit"                           
methods:                                  
  method01:                               
    friendlyName: Method01                
mappingRules:                         
  - httpMethod: GET                   
    pattern: "/pets"                  
    metricMethodRef: method01         
  - httpMethod: GET                   
    pattern: "/pets/id"               
    metricMethodRef: method01
backendUsages:      
  backendA:         
    path: /A        
applicationPlans:                  
  plan01:                          
    name: "My Plan 01"
    limits: 
      - period: month            
        value: 100               
        metricMethodRef:         
          systemName: hits       
      - period: month            
        value: 200               
        metricMethodRef:         
           systemName: hits       
           backend: backendA       
    pricingRules:                  
      - from: 1                    
        to: 100                    
        pricePerUnit: "15.45"      
        metricMethodRef:           
          systemName: hits         
      - from: 1                    
        to: 100                    
        pricePerUnit: "15.45"      
        metricMethodRef:           
          systemName: hits         
          backend: backendA        

```
Account (**NOT INCLUDED IN THIS PR**)
```
provider_account_reference: REF
applications: 
  app_id: 
    attributes: {}
```

ActiveDocs (**NOT INCLUDED IN THIS PR**)
```
provider_account_reference: REF
product_reference: REF
attributes: {}
```


cmd/manager/main.go
```diff
@@ -262,13 +262,7 @@ func filterGKVsFromAddToScheme(gvks []schema.GroupVersionKind) []schema.GroupVer
...
 
 		// Custom resource types
-		schema.GroupVersionKind{Group: "capabilities.3scale.net", Kind: "Plan", Version: matchAnyValue},
-		schema.GroupVersionKind{Group: "capabilities.3scale.net", Kind: "API", Version: matchAnyValue},
-		schema.GroupVersionKind{Group: "capabilities.3scale.net", Kind: "Limit", Version: matchAnyValue},
-		schema.GroupVersionKind{Group: "capabilities.3scale.net", Kind: "MappingRule", Version: matchAnyValue},
 		schema.GroupVersionKind{Group: "capabilities.3scale.net", Kind: "Tenant", Version: matchAnyValue},
-		schema.GroupVersionKind{Group: "capabilities.3scale.net", Kind: "Metric", Version: matchAnyValue},
-		schema.GroupVersionKind{Group: "capabilities.3scale.net", Kind: "Binding", Version: matchAnyValue},
 		schema.GroupVersionKind{Group: "apps.3scale.net", Kind: "APIManager", Version: matchAnyValue},
```
miguelsorianod, line 266: Do we want to add the new types here? APIManager and Tenant metrics are gathered (among other types that can be seen)

eguzki: Yes we want. Will add in another PR :+1:

pkg/common/status_conditions.go
```diff
@@ -0,0 +1,203 @@
...
+//
+// Conditions should be added to explicitly convey properties that users and
+// components care about rather than requiring those properties to be inferred
+// from other observations. Once defined, the meaning of a Condition can not be
+// changed arbitrarily - it becomes part of the API, and has the same
+// backwards- and forwards-compatibility concerns of any other part of the API.
+type Condition struct {
+	Type               ConditionType          `json:"type"`
+	Status             corev1.ConditionStatus `json:"status"`
+	Reason             ConditionReason        `json:"reason,omitempty"`
```
miguelsorianod, line 60: It is recommended to add +optional to Reason, Message and LastTransitionTime (although they already are with the omitempty json field) https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

pkg/apis/capabilities/v1beta1/product_types.go
```diff
@@ -0,0 +1,1095 @@
...
+}
+
+func (a *AuthenticationSpec) SecuritySecretToken() *string {
+	// authentication is oneOf by CRD openapiV3 validation
+	if a.UserKeyAuthentication != nil {
+		return a.UserKeyAuthentication.SecuritySecretToken()
+	}
+
+	if a.AppKeyAppIDAuthentication == nil {
+		panic("product authenticationspec: both userkey and appid_appkeyare nil")
```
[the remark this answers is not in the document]
miguelsorianod: Forget about my previous comment.

It seems, as you specified in comment code, that OpenAPI v3 validation is oneOf. I guess you added it manually on the CRD file. This will be hard to maintain and error-prone due to the generator will undo the manual changes but it should work. It'll be really easy to introduce the undoing of changes when we apply big changes to the CRDs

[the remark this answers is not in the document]
eguzki: If the CRD has openAPI v3 validation, how can the user violate that?

The CRD validation states:
```
oneOf:                          
 - properties:                  
     userkey:                   
       type: object             
   required: ["userkey"]        
 - properties:                  
     appKeyAppID:               
       type: object             
   required: ["appKeyAppID"]    
```

If you try to set none of the auth keys, you will get:
```
validation failure list:
"spec.deployment.apicastSelfManaged.authentication" must validate one and only one schema (oneOf). Found none valid
spec.deployment.apicastSelfManaged.authentication.userkey in body is required
```

If you try to set both auth keys, you will get:

```
validation failure list:
"spec.deployment.apicastSelfManaged.authentication" must validate one and only one schema (oneOf). Found 2 valid alternatives
```

How can you override CRD validation?

pkg/common/status_conditions.go
eguzki: :+1:

pkg/apis/capabilities/v1beta1/product_types.go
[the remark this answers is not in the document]
miguelsorianod: See my previous comment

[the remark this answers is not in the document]
eguzki: > This will be hard to maintain and error-prone due to the generator will undo the manual changes but it should work. It'll be really easy to introduce the undoing of changes when we apply big changes to the CRDs

Hopefully this validation will be supported by operator-sdk or underlaying kubebuilder tool https://github.com/operator-framework/operator-sdk/issues/1844

Any validation supported at CRD level, the less code we have to introduce in the operator.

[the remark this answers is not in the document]
miguelsorianod: :+1: let's hope it's introduced by operator-sdk/kubebuilder tool soon : (

pkg/apis/capabilities/v1beta1/product_types.go
```diff
@@ -0,0 +1,1095 @@
...
+
+// ProductStatus defines the observed state of Product
+// +k8s:openapi-gen=true
+type ProductStatus struct {
+	// +optional
+	ID *int64 `json:"productId,omitempty"`
+	// +optional
+	State *string `json:"state,omitempty"`
+
+	// ObservedGeneration reflects the generation of the most recently observed MachineSet.
```
miguelsorianod, line 764: Minor typo. `MachineSet` is incorrect

pkg/apis/capabilities/v1beta1/product_types.go
```diff
@@ -0,0 +1,1095 @@
...
+	// of terminal errors would be invalid combinations of settings in the
+	// spec, values that are unsupported by the product controller, or the
+	// responsible product controller itself being critically misconfigured.
+	//
+	// Any transient errors that occur during the reconciliation of Products
+	// can be added as events to the Product object and/or logged in the
+	// controller's output.
+	// +optional
+	// TODO enum
+	ErrorReason *string `json:"errorReason,omitempty"`
```
miguelsorianod, line 788: What's the reasoning of ErrReason and ErrorMessage being attributes of the status section and not conditions?
Seems ok to me but I'd like to know the reasoning so we can apply same criteria in other objects

pkg/apis/capabilities/v1beta1/product_types.go
eguzki: :+1:

pkg/apis/capabilities/v1beta1/product_types.go
eguzki: As the comment says, they are final errors that cannot be fixed retrying or anything like that. Not related to conditions and first level error field in Status section. 

I agree it could be redundant now. There exists a condition type specific for that.

pkg/apis/capabilities/v1beta1/backend_types.go
```diff
@@ -0,0 +1,307 @@
...
+	// Name is human readable name for the backend
+	Name string `json:"name"`
+
+	// SystemName identifies uniquely the product within the account provider
+	// Default value will be sanitized Name
+	// +optional
+	SystemName string `json:"systemName,omitempty"`
+
+	// +kubebuilder:validation:Pattern=`^https?:\/\/.*$`
+	PrivateBaseURL string `json:"privateBaseURL,omitempty"`
```
miguelsorianod, line 59: +optional comment tag is recommended to be added

pkg/apis/capabilities/v1beta1/product_types.go
```diff
@@ -0,0 +1,1095 @@
...
+
+	// Limits
+	// +optional
+	Limits []LimitSpec `json:"limits,omitempty"`
+
+	// TODO Features
+}
+
+// Methodpec defines the desired state of Product's Method
+type Methodpec struct {
```
miguelsorianod, line 139: Typo in the struct name (and comment). I guess it should be `MethodSpec`

pkg/apis/capabilities/v1beta1/backend_types.go
```diff
@@ -0,0 +1,307 @@
...
+	// In other words, if metric's system_name is A, there is no metric or method with system_name A.
+	// +optional
+	Metrics map[string]MetricSpec `json:"metrics,omitempty"`
+
+	// Methods
+	// Map: system_name -> MethodSpec
+	// system_name attr is unique for all metrics AND methods
+	// In other words, if metric's system_name is A, there is no metric or method with system_name A.
+	// +optional
+	Methods map[string]Methodpec `json:"methods,omitempty"`
```
miguelsorianod, line 79: Typo in the struct name. I guess it should be MethodSpec

pkg/helper/task_runner.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"fmt"
+	"time"
+
+	"github.com/go-logr/logr"
+)
+
+type task struct {
+	Name string
+	Run  func(interface{}) error
```
miguelsorianod, line 12: Why is the interface{} parameter added as a parameter of the Run method?

pkg/helper/task_runner.go
```diff
@@ -0,0 +1,52 @@
...
+// TaskRunner abstracts task running engine
+type TaskRunner interface {
+	Run() error
+	// AddTask register tasks to be executed sequentially
+	// Tasks will be executed in order. First in, first to be executed.
+	AddTask(string, func(interface{}) error)
+}
+
+type taskRunnerImpl struct {
+	ctx      interface{}
```
miguelsorianod, line 24: What does ctx mean in this case? What's the idea of this?

pkg/controller/backend/backend_controller.go
```diff
@@ -0,0 +1,223 @@
...
+		err := r.Client().Update(r.Context(), backendResource)
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("Failed setting backend defaults: %w", err)
+		}
+
+		logger.Info("resource defaults updated. Requeueing.")
+		return reconcile.Result{Requeue: true}, nil
+	}
+
+	backendAPI, specErr := r.reconcileSpec(backendResource)
```
miguelsorianod, line 153: Is there never a scenario where we want to requeue due to some reason before reconciling the whole spec? For example because step B requires step A be reconciled and changes effective before running step B

pkg/helper/backendapi_entity.go
```diff
@@ -0,0 +1,360 @@
...
+			return metric.Element.ID, nil
+		}
+	}
+
+	return 0, fmt.Errorf("backend [%s] hits not found", b.backendAPIObj.Element.SystemName)
+}
+
+// sanitizeBackendMetricList sanitizes systemName from backend metrics
+// Update is made in place
+// "system_name": "hits.45498" -> "system_name": "hits"
```
miguelsorianod, line 336: I'm curious about this. Why is a number received from the response?

pkg/helper/backendapi_entity.go
```diff
@@ -0,0 +1,360 @@
...
+			if metric.Element.SystemName == systemName {
+				metricList.Metrics = append(metricList.Metrics, metric)
+			}
+		}
+	}
+
+	return metricList, nil
+}
+
+func (b *BackendAPIEntity) getMetricsAndMethods() (*threescaleapi.MetricJSONList, error) {
```
miguelsorianod, line 307: Why in getMetricsAndMethods we only list metrics?

miguelsorianod: Is it because ListBackendapiMetrics golang Porta client returns metrics AND methods?

pkg/controller/product/product_controller.go
```diff
@@ -0,0 +1,416 @@
...
+		jsonData, err := json.MarshalIndent(product, "", "  ")
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		reqLogger.V(1).Info(string(jsonData))
+	}
+
+	// Ignore deleted Products, this can happen when foregroundDeletion is enabled
+	// https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#foreground-cascading-deletion
+	if product.DeletionTimestamp != nil {
```
miguelsorianod, line 128: Can this also happen for other CRs like APIManager?

pkg/helper/product_entity.go
```diff
@@ -0,0 +1,490 @@
...
+//
+// PRIVATE
+//
+//
+
+func (b *ProductEntity) resetBackendUsages() {
+	b.backendUsages = nil
+}
+
+func (b *ProductEntity) resetMethods() {
```
miguelsorianod, line 359: Why are all this reset* methods implemented in ProductEntity and BackendAPIEntity? Is it a caching mechanism that we invalidate when performing update/delete/create ? If so, why do we reset when creating or updating instead of updating the values?

pkg/controller/product/3scale_reconciler.go
```diff
@@ -0,0 +1,95 @@
...
+
+func (t *ThreescaleReconciler) Reconcile() (*helper.ProductEntity, error) {
+	productEntity, err := t.reconcile3scaleProduct()
+	if err != nil {
+		return nil, err
+	}
+	t.productEntity = productEntity
+
+	taskRunner := helper.NewTaskRunner(nil, t.logger)
+	taskRunner.AddTask("SyncProduct", t.syncProduct)
```
miguelsorianod, line 41: Why is syncProduct separated from reconcile3scaleProduct? other syncers contain creation/update/delete

pkg/controller/backend/backend_controller.go
```diff
@@ -0,0 +1,223 @@
+package backend
```
miguelsorianod, line 1: If I've looked correctly at the code, I understand currently deletion of Backends and deletion of Products on the 3scale side when deleting the corresponding CRs is not supported right?

pkg/controller/product/application_plan_reconciler.go
```diff
@@ -0,0 +1,391 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (a *applicationPlanReconciler) syncPlan(_ interface{}) error {
+	params := threescaleapi.Params{}
+
```
miguelsorianod, line 78: I understand we only try to add reconciliation code for fields that can be changed in the 3scale API side right?

For example, if we have a field in the CR but that field cannot be changed once the 3scale API entity is created, we have not added code to reconcile nor code verifying that the user is not doing anything "incorrect".

So, if a user changes a "create-only" field we simply ignore the change. Is that correct?

pkg/apis/capabilities/v1beta1/product_types.go
eguzki: Removing fields

pkg/apis/capabilities/v1beta1/backend_types.go
eguzki: it is not optional. removing `omitempty` json tag

pkg/apis/capabilities/v1beta1/product_types.go
eguzki: :+1:

pkg/apis/capabilities/v1beta1/backend_types.go
eguzki: fixed

pkg/helper/task_runner.go
eguzki: A context.

Some context that tasks can share to pass messages or objects.

It was used initially, but after a refactor, all tasks belong to the same type and they can share information using the type receiver. 

I left there for being more flexible for future use. But can be removed now and added later if needed.

pkg/helper/task_runner.go
eguzki: see my comment https://github.com/3scale/3scale-operator/pull/357#pullrequestreview-439353338

pkg/controller/backend/backend_controller.go
eguzki: `reconcileSpec` tries to reconcile backend spec with 3scale control plane. Does not know anything about requeing or k8s stuff.

There are some errors it can raise that the backend controller will read them and requeue, but that logic belong to the backend controller.

pkg/helper/backendapi_entity.go
eguzki: 3scale API design issues. 3scale Apisonator does not know about backends, so product metrics and backend metrics called `hits` or any other conflicting name have to be differentiated.

pkg/helper/backendapi_entity.go
eguzki: exactly. 

To get the list of metrics, you need to:
* get the list of metrics and methods
* get the list of methods
* get the difference: (metrics and methods ) - (methods)

eguzki: :shrug:

pkg/controller/product/product_controller.go
eguzki: Sure! if you delete APIManager with `foregrounddeletion` enabled you will get reconciliation loop with `DeletionTimestamp` set and if not checked, it will try to reconciliate spec as always. It can also happen if finalizers are used.

pkg/helper/product_entity.go
eguzki: ProductEntity and BackendEntity implement some 3scale API related code and implement caching as well. 

Regarding caching, the http response body is cached that is basically array of objects. When you delete or create an object, you could manipulate (add, delete, update elements) the cached array to be more efficient. For simplicity, it is reseted. Can be a improvement, indeed.

pkg/helper/task_runner.go
miguelsorianod: No need for now

pkg/controller/backend/backend_controller.go
miguelsorianod: You might want to be able to requeue without an error between steps of the reconciler, which does not seem possible with this approach as the information that the backend controller receives is an error.
Might be worth to add that possibility in the future.

pkg/helper/backendapi_entity.go
miguelsorianod: :+1:

pkg/helper/backendapi_entity.go
miguelsorianod: :+1:

pkg/helper/product_entity.go
miguelsorianod: :+1:

pkg/controller/product/3scale_reconciler.go
eguzki: It is a good question. There is no good answer. 

`reconcile3scaleProduct` is kind of "CreateSimpleOnlyIfNotExist" method.
`syncProduct` reconciliates all attributes provided it already exists.

All tasks assume the product already exists, and they can access `productEntity` object to manipulate it. Add metrics, mappingrules, etc...

The full reconciliation of the product and attributes (not sub resources like metrics and so on) could be done in `reconcile3scaleProduct`

miguelsorianod: :+1:

pkg/controller/backend/backend_controller.go
eguzki: Backend delete https://issues.redhat.com/browse/THREESCALE-5473
Product Delete https://issues.redhat.com/browse/THREESCALE-5475

miguelsorianod: :+1:

pkg/controller/product/product_controller.go
miguelsorianod: Ok. We should think on adding it to the others then in the future to avoid this possible race condition

pkg/controller/product/application_plan_reconciler.go
eguzki: One example is the `system-name`. It is a create only attr that 3scale API does not allow to update. The operator will not try to reconciliate, it will use as ID. If you change it in you spec, new element will be created or some other element can be reconciled.

So, yes. The reconlliation logic does take this kind of attributes into account and operate accordingly. If not, it is a bug.

miguelsorianod: :+1:

PR #358, opened 2020-04-29
Prototype apimanager backup/restore - PVC-based subset

miguelsorianod:
Tasks:

- [x] Update CSV contents when the functionality is ready


miguelsorianod:
Here is a first prototype of how backup/restore functionality would be for 3scale APIManager based installation with the operator.

I've added some initial documentation. See:
doc/operator-backup-and-restore.md
doc/apimanagerrestore-reference.md
doc/apimanagerbackup-reference.md

Current functionality supports the scenario of a PVC-based backup to a PVC-based restore, using PVC as System's FileStorage (both in backup and restore)

Let's have some feedback about the functionality and code and iterate a little bit.

Thank you.

doc/apimanagerbackup-reference.md
```diff
@@ -0,0 +1,111 @@
...
+  all [data that is backed up](#data-that-is-backed-up)
+* Providing the desired size of the PersistentVolumeClaim through the
+  `resources` field. In this case the `storageClass` field
+  can be also set to specify what StorageClass will be used for the
+  PersistentVolumeClaim if desired. Make sure enough resources are defined
+  to contain all [data that is backed up](#data-that-is-backed-up)
+
+| **json/yaml field**| **Type** | **Required** | **Default value** | **Description** |
+| --- | --- | --- | --- | --- |
+| `resources` | [PersistentVolumeClaimResourcesSpec](#PersistentVolumeClaimResourcesSpec) | No | See [PersistentVolumeClaimResourcesSpec](#PersistentVolumeClaimResourcesSpec) |
```
eguzki, line 90: From backup CRD, `resources` are required?

deploy/crds/apps.3scale.net_apimanagerbackups_crd.yaml
```diff
@@ -0,0 +1,154 @@
...
+                    needed: Detailed info: https://docs.okd.io/3.11/dev_guide/persistent_volumes.html#persistent-volumes-volumes-and-claim-prebinding'
+                  properties:
+                    resources:
+                      properties:
+                        requests:
+                          type: string
+                      required:
+                      - requests
+                      type: object
+                    storageClass:
```
eguzki, line 55: storageClass inside `PersistentVolumeClaimResourcesSpec`??

doc/operator-backup-and-restore.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Backing up 3scale
+
+The backup functionality of a 3scale installation deployed by an `APIManager`
+custom resource is provided.
+
+### Backup compatible scenarios
+
+To see what 3scale installation configurations can be backed up see the
+[APIManagerBackup reference](#apimanagerbackup-reference.md). Specifically the following sections should be
```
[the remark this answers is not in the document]
miguelsorianod: :+1:

deploy/crds/apps.3scale.net_apimanagerbackups_crd.yaml
```diff
@@ -0,0 +1,154 @@
...
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          type: object
+        spec:
+          description: APIManagerBackupSpec defines the desired state of APIManagerBackup
+          properties:
+            apiManagerName:
+              description: 'INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
```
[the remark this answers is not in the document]
miguelsorianod: :+1: when PR is more advanced I'll review it

deploy/crds/apps.3scale.net_apimanagerbackups_crd.yaml
```diff
@@ -0,0 +1,154 @@
...
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          type: object
+        spec:
+          description: APIManagerBackupSpec defines the desired state of APIManagerBackup
+          properties:
+            apiManagerName:
```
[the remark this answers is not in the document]
miguelsorianod: The purpose is that currently there are two methods to retrieve the APIManager to backup:
1. You provide the APIManager name that you want to backup through this 'apiManagerName' optional attribute and the operator tries to find that name.
1. In case apiManagerName is not provided it tries to find an existing APIManager in the namespace where the 3scale-operator has been deployed

deploy/crds/apps.3scale.net_apimanagerbackups_crd.yaml
```diff
@@ -0,0 +1,154 @@
...
+                      - requests
+                      type: object
+                    storageClass:
+                      type: string
+                    volumeName:
+                      type: string
+                  required:
+                  - resources
+                  type: object
+                simpleStorageService:
```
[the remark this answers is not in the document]
miguelsorianod: I left this here because some work was already done for it and so we could get an idea (even in the PR) on the organization that it might take adding more stuff. Shall we remove it just before merging the PR?

deploy/crds/apps.3scale.net_apimanagerrestores_crd.yaml
```diff
@@ -0,0 +1,122 @@
...
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          type: object
+        spec:
+          description: APIManagerRestoreSpec defines the desired state of APIManagerRestore
+          properties:
+            restoreSource:
+              description: 'INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
```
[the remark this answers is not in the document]
miguelsorianod: :+1:

deploy/crds/apps.3scale.net_apimanagerrestores_crd.yaml
```diff
@@ -0,0 +1,122 @@
...
+        status:
+          description: APIManagerRestoreStatus defines the observed state of APIManagerRestore
+          properties:
+            apiManagerBackupSharedInSecret:
+              type: boolean
+            apiManagerBackupSharedInSecretCleanup:
+              type: boolean
+            apiManagerRestoreStepFinished:
+              type: boolean
+            completed:
```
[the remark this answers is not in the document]
miguelsorianod: The thing with conditions is that they can only take the value "true/false", not an "enum" of states.
Conditions are also a little bit harder to manipulate (searching, adding, removing elements etc...) so I went the simpler way.

Additionally, we could end up wanting more complex information than just a state, that's why I was thinking whether it would be a good idea to structure more the status, like:

.status.pvcDestination.configmapsAndSecrets.BackupFinished and other kind of potential information like .status.pvcDestination.logs

etc...

How would we fit the potential new information and lack of inherent structure of conditions?

doc/apimanagerbackup-reference.md
```diff
@@ -0,0 +1,111 @@
+# APIManagerBackup reference
+
+The following Custom Resources are provided:
+
+`APIManagerBackup`
+
+This resource is the resource used to backup a 3scale API Management solution
+deployed using an APIManager custom resource.
+
+## Backup scenarios scope
```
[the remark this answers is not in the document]
miguelsorianod: Where does it belong?

doc/apimanagerbackup-reference.md
miguelsorianod: Yes.

When you create a PVC, if you do not provide an existing PV name you have to specify the minimum size that you want. See PersistentVolumeClaimResourcesSpec section with details on what you can configure

deploy/crds/apps.3scale.net_apimanagerbackups_crd.yaml
miguelsorianod: It is not inside `PersistentVolumeClaimResourcesSpec`. It is inside `PersistentVolumeClaimBackupSource`

miguelsorianod:
Thanks for the first review

deploy/crds/apps.3scale.net_apimanagerbackups_crd.yaml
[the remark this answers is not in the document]
eguzki: Only one apimanager object per namespace is allowed.

deploy/crds/apps.3scale.net_apimanagerbackups_crd.yaml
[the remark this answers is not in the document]
eguzki: My point is, if we do not provide any support for s3 storage backup, that attribute does not make sense.  If you are planning to support s3 storage before this functionality is merged to master, then that attribute makes sense in early stages.

doc/apimanagerbackup-reference.md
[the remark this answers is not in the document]
eguzki: Up to you. `operator-backup-and-restore.md` or somewhere else?

doc/apimanagerbackup-reference.md
eguzki: `Required` field value :eyes:

deploy/crds/apps.3scale.net_apimanagerbackups_crd.yaml
eguzki: I know, it was not a question, but a proposal.

The options are providing either:
* volumeName
* Resources and (optionally) storageClass

So those three attributes of `PersistentVolumeClaimBackupSource`  are not sibling attributes. There are two groups. Hence my proposal.

doc/operator-backup-and-restore.md
```diff
@@ -0,0 +1,135 @@
+# 3scale installation Backup and Restore using the operator
+
+## Table of contents
```
eguzki, line 3: Check TOC anchors, some of them are invalid

eguzki:
Dropped some comments

doc/operator-backup-and-restore.md
miguelsorianod: :+1:

deploy/crds/apps.3scale.net_apimanagerbackups_crd.yaml
miguelsorianod: This section intended to mimic the Kubernetes PVC settings, where they are in the same level indeed https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#persistentvolumeclaimspec-v1-core.
I personally prefer to keep it as similar as PVC itself.

doc/apimanagerbackup-reference.md
miguelsorianod: :+1: I'll add the optional annotation

doc/apimanagerbackup-reference.md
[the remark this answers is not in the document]
miguelsorianod: If you look at the documentation, from operator-backup-and-restore.md I link to this reference scopes.
Do you want me to move those to operator-backup-and-restore.md? what about the other information like what is backed up/restored that it is also in the reference documentation?

It somewhat seemed it might be too verbose to be in the general operator-backup-and-restore.md documentation and I decided to link to the reference of those sections from the general documentation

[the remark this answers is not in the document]
eguzki: Ok, up to you

deploy/crds/apps.3scale.net_apimanagerbackups_crd.yaml
eguzki: ok, being consistent with PVC structure is desired :+1:

deploy/crds/apps.3scale.net_apimanagerbackups_crd.yaml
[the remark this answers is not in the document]
miguelsorianod: I added that option just as an additional possibility where the user could made that explicit on the spec. I understand you don't see it necessary and you want it removed. I don't see it adds a lot of value so I have removed it

doc/apimanagerbackup-reference.md
miguelsorianod: Added

doc/operator-backup-and-restore.md
miguelsorianod: I've fixed the ones I've been able to detect broken

deploy/crds/apps.3scale.net_apimanagerbackups_crd.yaml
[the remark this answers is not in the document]
miguelsorianod: I don't intend to implement S3 at this point. I've removed S3-related code for this PR

doc/apimanagerbackup-reference.md
```diff
@@ -0,0 +1,111 @@
...
+
+Backup functionality is available when the following databases are
+configured externally:
+* System database (MySQL or PostgreSQL)
+* Backend Redis database
+* System Redis database
+
+## Data that is backed up
```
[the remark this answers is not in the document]
miguelsorianod: Can you check the 'Backup compatible scenarios' and 'Restore' compatible scenarios sections in the operator-backup-and-restore.md document? From there I link to the corresponding reference sections where I go into detail on several aspects. Is that ok?

miguelsorianod:
Rebased from latest contents of master that have the reconcilers refactoring

pkg/reconcilers/base_reconciler.go
```diff
@@ -101,7 +102,11 @@ func (b *BaseReconciler) ReconcileResource(obj, desired common.KubernetesObject,
 
 	// item found successfully
 	if common.IsObjectTaggedToDelete(desired) {
```
miguelsorianod, line 104: Notice/review new deletion implementation additions please. They are used in the new logic reconcilers

pkg/reconcilers/base_reconciler.go
```diff
@@ -126,7 +136,7 @@ func (b *BaseReconciler) UpdateResource(obj common.KubernetesObject) error {
 	return b.Client().Update(b.ctx, obj)
 }
 
-func (b *BaseReconciler) DeleteResource(obj common.KubernetesObject) error {
+func (b *BaseReconciler) DeleteResource(obj common.KubernetesObject, options ...client.DeleteOption) error {
```
miguelsorianod, line 139: Added ability to set options, needed for cascade deleting

pkg/reconcilers/base_reconciler.go
eguzki: :+1:

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
```diff
@@ -0,0 +1,398 @@
...
+
+	// TODO create Mutator function for Jobs??
+	existing := &batchv1.Job{}
+	// Check if this step has finished
+	if r.cr.SecretsAndConfigMapsBackupStepFinished() {
+		return reconcile.Result{}, nil
+	}
+
+	// Check if backup substep has completed
+	if !r.cr.SecretsAndConfigMapsBackupSubStepFinished() {
```
eguzki, line 157: `r.cr.SecretsAndConfigMapsBackupStepFinished()`

`r.cr.SecretsAndConfigMapsBackupSubStepFinished()`

names are very confusing. What is that substep about? Maybe a better name?

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
```diff
@@ -0,0 +1,398 @@
...
+		err := r.ReconcileResource(existing, desired, reconcilers.CreateOnlyMutator)
```
eguzki, line 158: I think it is quite ugly to use `existing`, although it works for now.

I would not have implemented with `ReconcileResource` method because Jobs do not need reconcilliation. 

I would have done

```
r.GetResource(NamespacedName(desired), existing)
if notFound(){
     r.CreateREsource(desired)
     // you can expect the job to not be finished instantly, so requeue
      return reconcile.Result{Requeue: true}, nil
}

// continue with the existing object checking for completion...
```

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
```diff
@@ -0,0 +1,398 @@
...
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		r.Logger().Info("Job finished successfully. Requeing", "Job Name", desired.Name)
+		return reconcile.Result{Requeue: true}, nil
+	}
+
+	// Check if cleanup substep has completed
+	if !r.cr.SecretsAndConfigMapsCleanupSubStepFinished() {
+		common.TagToObjectDeleteWithPropagationPolicy(desired, metav1.DeletePropagationForeground)
```
eguzki, line 179: I am curious. Why do you need [Foreground cascading deletion](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#foreground-cascading-deletion). Which are resources owned by the job that needs to be deleted first?

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
```diff
@@ -0,0 +1,398 @@
...
+		err := r.ReconcileResource(existing, desired, reconcilers.CreateOnlyMutator)
```
eguzki, line 180: if `r.cr.SecretsAndConfigMapsCleanupSubStepFinished()` is false, you can assume Job object has been created. can't you?

Why not like this?
```
err = r.GetResource(common.ObjectKey(desired), existing)      
if err != nil 
    if !errors.IsNotFound(err) {                    
    }
    // Job Not found
    // Here check config maps cleanup done
       r.cr.Status.SecretsAndConfigMapsCleanupSubStepFinished = &substepFinished
     ...
}                                                             


```

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
```diff
@@ -0,0 +1,398 @@
...
+		// if we do not do it this way it could happen that we mark the state of this
+		// substep as completed but the Job is not completely deleted and we move to the
+		// next step, which would fail because Job is not deleted nor the PVC associated
+		// to it, which would provoke an error when trying to reuse it for the next job PVC
+		err = r.GetResource(common.ObjectKey(desired), existing)
+		if err != nil && !errors.IsNotFound(err) {
+			return reconcile.Result{}, err
+		}
+
+		substepFinished := true
```
eguzki, line 197: Here the job might still be alive, right? You do not check it does not exists.

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
```diff
@@ -0,0 +1,398 @@
...
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		r.Logger().Info("Job cleant up successfully. Requeuing", "Job Name", desired.Name)
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (r *APIManagerBackupLogicReconciler) reconcileAPIManagerCustomResourceBackupToPVCJob() (reconcile.Result, error) {
```
eguzki, line 209: Same sort of comments as for `reconcileBackupSecretsAndConfigMapsToPVCJob`

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
```diff
@@ -0,0 +1,398 @@
...
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		r.Logger().Info("Job finished successfully. Requeing", "Job Name", desired.Name)
+		return reconcile.Result{Requeue: true}, nil
+	}
+
+	// Check if cleanup substep has completed
+	if !r.cr.SystemFileStorageCleanupSubStepFinished() {
+		// TODO delete performed with ReconcileResource does not have the option to delete
```
eguzki, line 303: In Kubernetes >=1.9, for all kinds in the apps/v1 group version, dependent objects are deleted by default. Why do you need to specify cascading deletion policy?

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
eguzki: I see, created pods are dependent of the job. But the need of foreground cascading deletion I still do not see.

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
```diff
@@ -0,0 +1,398 @@
...
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		r.Logger().Info("Job cleant up successfully. Requeuing", "Job Name", desired.Name)
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (r *APIManagerBackupLogicReconciler) reconcileBackupSystemFileStoragePVCToPVCJob() (reconcile.Result, error) {
```
eguzki, line 268: Same sort of comments as for reconcileBackupSecretsAndConfigMapsToPVCJob

eguzki:
Maybe the operator should look out there to know in which state is the pipeline, instead of checking status booleans. For instance, check if jobs are running. If secrets and maps are in the PV and the job does not exist, it means secrets and maps have been backed up and the stage is completed. 

Just food for thought.

miguelsorianod:
Hi @eguzki ,

Thank you very much for your feedback. I'll take a look and answer it as soon as possible :+1:

pkg/common/kubernetes_types.go
```diff
@@ -39,3 +56,17 @@ func IsObjectTaggedToDelete(obj KubernetesObject) bool {
 	annotation, ok := annotations[DeleteTagAnnotation]
 	return ok && annotation == "true"
 }
+
+func GetDeletePropagationPolicyAnnotation(obj KubernetesObject) *metav1.DeletionPropagation {
```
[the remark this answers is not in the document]
miguelsorianod: I think it was to differentiate between unset and set in the ReconcileResource method. metav1.DeletionPropagation is also its own type with its set of allowed values

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
```diff
@@ -0,0 +1,398 @@
...
+
+func NewAPIManagerBackupLogicReconciler(b *reconcilers.BaseReconciler, cr *appsv1alpha1.APIManagerBackup) (*APIManagerBackupLogicReconciler, error) {
+	res := &APIManagerBackupLogicReconciler{
+		BaseReconciler: b,
+		logger:         b.Logger().WithValues("APIManagerBackup Controller", cr.Name),
+		cr:             cr,
+	}
+
+	if !cr.BackupCompleted() {
+		// We only set the apiManagerBackup field when
```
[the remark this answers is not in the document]
miguelsorianod: :+1: Doing the way you commented has the same effect than what I was doing if I'm not wrong. Visually if the backup is completed with your proposed approach you return earlier which I like it so I'll change it :+1: . Please review when doing the new round of review

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
```diff
@@ -0,0 +1,398 @@
...
+}
+
+func (r *APIManagerBackupLogicReconciler) reconcileAPIManagerSourceStatusField() (reconcile.Result, error) {
+	apiManager := r.apiManagerBackup.APIManager()
+
+	if r.cr.Status.APIManagerSourceName == nil {
+		r.cr.Status.APIManagerSourceName = &apiManager.Name
+		err := r.UpdateResourceStatus(r.cr)
+		if err != nil {
+			return reconcile.Result{}, nil
```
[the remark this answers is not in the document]
miguelsorianod: Yes. That was an error. Thanks! I have also detected it in another condition a little bit below and in the restore_logic_reconciler. I'll fix them :+1:

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
```diff
@@ -0,0 +1,398 @@
...
+		}
+		r.Logger().Info("APIManager source name set in status. Requeuing", "APIManager source name", r.cr.Status.APIManagerSourceName)
+		return reconcile.Result{Requeue: true}, err
+	}
+	if *r.cr.Status.APIManagerSourceName != apiManager.Name { // TODO should we reconcile this case?
+		r.cr.Status.APIManagerSourceName = &apiManager.Name
+		err := r.UpdateResourceStatus(r.cr)
+		if err != nil {
+			return reconcile.Result{}, nil
```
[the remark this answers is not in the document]
miguelsorianod: Yes. See comment above :+1:

pkg/apis/apps/v1alpha1/apimanagerbackup_types.go
```diff
@@ -0,0 +1,173 @@
...
+	APIManagerCustomResourceCleanupSubStepFinished *bool `json:"apiManagerCustomResourceCleanupSubStepFinished,omitempty"`
+	// +optional
+	SystemFileStorageBackupSubStepFinished *bool `json:"systemFileStorageBackupSubStepFinished,omitempty"`
+	// +optional
+	SystemFileStorageCleanupSubStepFinished *bool `json:"systemFileStorageCleanupSubStepFinished,omitempty"`
+	// +optional
+	Completed *bool `json:"completed,omitempty"`
+
+	// +optional
+	APIManagerSourceName *string `json:"apiManagerSourceName,omitempty"`
```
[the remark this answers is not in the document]
miguelsorianod: I see a value for it. Whereas what you say it's true, the APIManager deployed in the same namespace does not necessarily have to be the same than the one the backup was performed. There could be also the situation that there's no APIManager anymore in the namespace but you still have one or more backups on it.
I think it's useful having track of how was the APIManager called when a given specific backup was performed.

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
```diff
@@ -0,0 +1,398 @@
...
+}
+
+func (r *APIManagerBackupLogicReconciler) reconcileBackupDestinationPVC() error {
+	desired := r.apiManagerBackup.BackupDestinationPVC()
+	if desired == nil {
+		return nil
+	}
+
+	// TODO create mutator function for PVC ?
+	err := r.ReconcileResource(&v1.PersistentVolumeClaim{}, desired, reconcilers.CreateOnlyMutator)
```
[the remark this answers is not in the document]
miguelsorianod: That's a good question and I thought about it before.
Adding an OwnerReference to the PVC would mean that when the APIManager is deleted the PVC would be deleted too (and the PV behind it too).
Adding/Removing it has some advantages and disadvantages as we know.

I decided to keep the PVC even after deleting the APIManagerBackup object so the PVC/PV can still be used even if the APIManagerBackup object does not exist anymore, or the user wants to reuse the PV for something else.

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
miguelsorianod: I agree with you that names are very confusing. That's the inherent problem of placing them in the same level of structure or in conditions where you have to namespace the names themselves and you cannot really define structure for them.
I could not come up with more clear names having them all in the same level.

An alternative I thought was to just create structure for them as I commented you in the https://github.com/3scale/3scale-operator/pull/358#discussion_r421516552 discussion. Allowing us to namespace the state and give it structure and potentially having other contextual information for some of them. Additionally, we could end up wanting more complex information than just a state, that's why I was thinking whether it would be a good idea to structure more the status, like:

.status.pvcDestination.configmapsAndSecrets.BackupFinished and other kind of potential information like .status.pvcDestination.logs

Of course this means that we would incorporate it as part of the CRD schema (although that's done also with conditions or fields at the same level  I think), but maybe being all of them optional would be acceptable? Of course in that case too we would have to perform nil checks in several places around the code as with any optional nested fields.

What do you think? As you suggested in another comment I totally think we should do code review/discuss this part

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
miguelsorianod: Ok I'll have it in the back of my mind. I might change it. Let's leave it this way for now until other parts are discussed :+1:

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
miguelsorianod: You are correct about pods are dependent on the job and they are not deleted automatically unless you provide a cascading deletion policy.

miguelsorianod: Regarding foreground or background cascading deletion, the reason I used foreground is to be sure the Jobs and its dependent elements (Pods) are deleted before going to the next step.
If I were to use background I think it might happen that the Job is deleted but its pods would not be deleted yet. The code checks the jobs (and not its pods) so there might be a scenario where job would be deleted and pods not? (and PVCs are at pod level at the end, and are needed on next stages).

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
miguelsorianod: For what I observed, when using Go client they don't by default. When using kubectl it's the default behaviour but it isn't when using Go apparently

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
```diff
@@ -0,0 +1,398 @@
...
+		}
+		r.Logger().Info("Job cleant up successfully. Requeuing", "Job Name", desired.Name)
+		return reconcile.Result{Requeue: true}, nil
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (r *APIManagerBackupLogicReconciler) reconcileBackupCompletion() (reconcile.Result, error) {
+	if !r.cr.BackupCompleted() {
```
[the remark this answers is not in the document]
miguelsorianod: I agree. It somehow seems more robust checking that so it does not try to perform the update always just in case there was some error before.
If you look at the TODO I left there, some of my comments where:
```
// TODO make this more robust only setting it in case all substeps have been completed?
// It might be a little bit redundant because the steps are checked during the reconciliation
```
So I was/am not sure whether it would be a good idea that the state is consistent at that point. It _should_ be at that point but it might not in reality (if there is some error in our side). It's kind of, is it worth it adding this sort of redundancy for added security?

What do you think? I could see both solutions and I have points in favour and against each of them

pkg/backup/apimanager_backup.go
```diff
@@ -0,0 +1,346 @@
...
+			// TODO BackoffLimit field controls how many times the job is retried
+			Template: v1.PodTemplateSpec{
+				Spec: v1.PodSpec{
+					Volumes: []v1.Volume{
+						b.pvcBackupDestinationPodVolume(),
+					},
+					Containers: []v1.Container{
+						v1.Container{
+							Name:  "backup-cfgmaps-secrets",
+							Image: "registry.redhat.io/openshift4/ose-cli:4.2",
```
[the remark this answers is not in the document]
miguelsorianod: Yes they should. Because of the current state of the PR / solution I decided to code it that way at that moment. When we advance more in this PR let's discuss changing it before merging it :+1:

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
miguelsorianod: I'm not sure I'm following you.
If r.cr.SecretsAndConfigMapsCleanupSubStepFinished() is false it means the Job object has been created and finished correctly indeed.

In this step you have to delete the Job in case it still exists so you have to end up calling the ReconcileResource object anyway with the deletion tag, and you can only mark as finished when it's fully deleted. You cannot call the ReconcileResource to delete and then unconditionally continue. I tried to explain that in a TODO just below. With the code you shown me it still lacks the deletion part, which has to be added anyway.

If you look a little bit below you'll see that I might be doing sort of what you say, but after performing the deletion:
```
		err = r.GetResource(common.ObjectKey(desired), existing)
		if err != nil && !errors.IsNotFound(err) {
		}
```

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
miguelsorianod: I understand you refer to your comment over here https://github.com/3scale/3scale-operator/pull/358#discussion_r425739955. I answered there. If you were not referring to that please tell me

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
miguelsorianod: You are right. I've added check here and on other places where the deletion is performed.

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
miguelsorianod: I understand you refer to your comment over here #358 (comment). I answered there. If you were not referring to that please tell me

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
miguelsorianod: EDIT: I've added more code related to feedback you left me in https://github.com/3scale/3scale-operator/pull/358#discussion_r425740366. Take that into account for this discussion also

miguelsorianod:
> Maybe the operator should look out there to know in which state is the pipeline, instead of checking status booleans. For instance, check if jobs are running. If secrets and maps are in the PV and the job does not exist, it means secrets and maps have been backed up and the stage is completed. I think it would be far more robust process, and easier to implement.

I thought about that and initially that's how I implemented it. The problem I found with that solution is that the affirmation "If secrets and maps are in the PV and the job does not exist, it means secrets and maps have been backed up and the stage is completed" is hard to check / is not necessarily true:
If the job does not exist it can mean two things: That it has finished completely and you have deleted it in a previous reconcile loop, or that it has not started yet. How can you check that if you don't have some state stored? The part of "if secrets and maps are in the pv" you can only check by looking at the pvc content, which already requires a job to mount and you end up with the same problem again about this state of the job, or store the state somewhere.

miguelsorianod:
@eguzki applied changes based on some of your feedback and answered it. Please check again. Thank you.

eguzki:
> If the job does not exist it can mean two things: That it has finished completely and you have deleted it in a previous reconcile loop, or that it has not started yet. How can you check that if you don't have some state stored? The part of "if secrets and maps are in the pv" you can only check by looking at the pvc content, which already requires a job to mount and you end up with the same problem again about this state of the job, or store the state somewhere.

The job itself let's you know state machine's current state.

How about the following idea in Go/PseudoCode:

```
func ProcessPipeline(jobs []Jobs{secretsJob, pvcJob}) (result, error) {
   for _, job := range jobs {
     existingJob := Job{}
     err := client.GetResource(job.NamespacedName, existingJob)
     if err != nil && !errors.IsNotFound(err){
        return Result{}, err
     }
     if errors.IsNotFound(err){
        err := client.CreateResource(job)
         if err != nil {
            return Result{}, err
         }
         // SInce the job has just started it is assumed it will take same time, thus requeue
          return Result{Requeue: true}, nil
     }

     // Job found
     if job.Failed() {
         backup.Status.Error = job.GetError()
          client.Status().Update(backup)
          // do not try to requeue
          return Result{}, nil
     }
     if !job.Completed() {
        // Just requeue
          return Result{Requeue: true}, nil
     }
   }

   // Alll jobs are completed
   // Update Backup resource as completed
   backup.Status.Completed = true
   client.Status().Update(backup)
}

// jobs owner reference is the backup CR
// so jobs will be deleted when backup CR is deleted.
mypipelineA := []Jobs{secretsJob, pvcJob}

ProcessPipeline(mypipelineA)
```

I think it has some advantages:
* It is more robust than current implementation. Logic is not based on booleans set in the status. State machine logic is based on cluster state and reconciled against a desired state (all jobs are done in order). Status field only shows information to the user and it is not used to coordinate pipeline.
* It is simpler than the current implementation. Scales better when pipelines are more complex. No need to check a set of boolean flags that will grow if pipelines get more complex.
* Code can be reused for any pipeline. For instance, for the pipeline that will be implemented for S3 backup. The Job object can even be abstracted to be any Job, not just OCP Job. Then a job could be anything to be done. It must implement `Get`, `Create`, `Completed` and `Failed` interface methods.

miguelsorianod:
I agree what you proposed would be the ideal solution and that's how I started to implement it in the beginning.

> The job itself let's you know state machine's current state.

However, you are assuming the Job itself exists as completed state when it finishes.

PVCs (ReadWriteOnce) are specified in the Job (and thus it is mounted in the Pod that is deployed by the Job). The current logic deletes the Job to make sure the volume is completely "free'd" to be used for the next Job (also to cleanup but that's more 'secondary).

Maybe when the Job is finished the PVC it could be mounted even if it already exists in a previous Job due to the pod being terminated. I can take a look at it. If it can't be done the proposed solution would not work. What do you think?

Additionally, if it is possible, do you think we should cleanup the Jobs (and thus its pods) or let them be there? In case you think we should clean-up them, at what point would we do it? If we do it before setting the complete mark then we would lose track of them in the next reconcile loop so we would have the same problem. We could do it after that but it would require another boolean flag or some mark.

deploy/crds/apps.3scale.net_apimanagerbackups_crd.yaml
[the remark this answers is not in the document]
miguelsorianod: Done in #375 :+1:

deploy/crds/apps.3scale.net_apimanagerrestores_crd.yaml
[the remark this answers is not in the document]
miguelsorianod: Done in #375 +1

deploy/crds/apps.3scale.net_apimanagerrestores_crd.yaml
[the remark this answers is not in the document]
miguelsorianod: Followed a different approach based on K8s objects direct state observation and auxiliary bools in case they were needed

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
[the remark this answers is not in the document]
miguelsorianod: This is obsolete as the state check was refactored following another approach. Marking as resolved.

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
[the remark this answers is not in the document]
miguelsorianod: Marking as resolved. If you have any comment to add please tell me.

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
miguelsorianod: This is obsolete as the state check was refactored following another approach. Marking as resolved.

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
miguelsorianod: This is obsolete as the state check was refactored following another approach. Marking as resolved.

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
miguelsorianod: This is obsolete as the state check was refactored following another approach. Marking as resolved.

pkg/controller/apimanagerbackup/apimanagerbackup_logic_reconciler.go
miguelsorianod: Marking as resolved. If you have any comment about this please tell me.

miguelsorianod:
Hi,

This functionality should be able to be tested at this moment so we are going to merge the PR to the 'backup-restore-integration' branch. cc @3scale/qe you should be able to start looking into this functionality to test it.
Keep in mind that we still expect changes/additions to happen in that branch but we think the functionality should be stable enough to be tested already.

PR #359, opened 2020-05-04
Add restore wait for apimanager readiness and resync of zync routes

miguelsorianod:
This PR adds the following:
* Waits until APIManager resource has all non-external Deployments in "ready" state.
* Trigger of zync domains resync via a Kubernetes Job

This PR is built on top of PR https://github.com/3scale/3scale-operator/pull/358 and the base merge branch is the one from that PR.

We also should evaluate/consider whether we want to include the functionality in the restore itself. Doing so implies that the restore does not finish until the deployments are ready (it keeps waiting) and the domains resync finishes without an exit code error

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -63,6 +63,12 @@ rules:
   verbs:
   - get
 - apiGroups:
+  - ""
+  resources:
+  - pods/exec
+  verbs:
+  - create
+- apiGroups:
   - apps
   resources:
   - replicasets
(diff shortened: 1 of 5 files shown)


miguelsorianod:
if at the end changes are ok and we want to include them there's a pending point: update documentation to reflect the new changes.

miguelsorianod:
Ready for review @eguzki. I've also updated documentation.

PR #360, opened 2020-05-15
Zync BuildConfig with UBI7 base image

guicassolato:
Changes the image of the base ImageStream used by the zync BuildConfig from `centos/ruby-24-centos7` to `registry.access.redhat.com/ubi7/ruby-25`. This not only changes the underlying operating system to RHEL7, but also upgrades Ruby to 2.5.

This change follows https://github.com/3scale/zync/pull/328, which aligned the recommended settings for building Zync in Opesnhift with its current Dockerfile in the upstream and Ruby version upon which CI tests are run and downstream images are built.

I understand this will change the nightly builds, making them aligned to current development stage of the upstream. https://github.com/3scale/3scale-operator/blob/cc029c11494d99c91fb94e9778fb6a713373dfa7/.circleci/config.yml#L226.

### DIFF
--- a/pkg/3scale/amp/manual-templates/amp/build.yml
+++ b/pkg/3scale/amp/manual-templates/amp/build.yml
@@ -11,12 +11,12 @@ objects:
     annotations:
     labels:
       app: zync
-    name: ruby-24-centos7
+    name: ruby-25-ubi7
   spec:
     tags:
     - from:
         kind: DockerImage
-        name: centos/ruby-24-centos7
+        name: registry.access.redhat.com/ubi7/ruby-25
       name: latest
 
 - apiVersion: v1
@@ -40,7 +40,7 @@ objects:
       dockerStrategy:
         from:
           kind: ImageStreamTag
-          name: ruby-24-centos7:latest
+          name: ruby-25-ubi7:latest
       type: Docker
 
 - apiVersion: v1
--- a/pkg/3scale/amp/manual-templates/amp/build.yml
+++ b/pkg/3scale/amp/manual-templates/amp/build.yml
@@ -16,7 +16,7 @@ objects:
     tags:
     - from:
         kind: DockerImage
-        name: registry.access.redhat.com/ubi7/ruby-25
+        name: registry.redhat.io/ubi7/ruby-25
       name: latest
 
 - apiVersion: v1
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -195,6 +195,16 @@ commands:
             done) 172.30.1.1:5000/$project/amp-backend:${imagestream_tag_name}=quay.io/3scale/apisonator:nightly \
             172.30.1.1:5000/$project/amp-system:${imagestream_tag_name}=quay.io/3scale/porta:nightly --insecure
 
+  create-redhat-registry-io-secret:
+    steps:
+      - run:
+          name: Create registry.redhat.io secret
+          command: |
+            oc create secret docker-registry threescale-registry-auth \
+              --docker-password="${REGISTRY_REDHAT_IO_PASSWORD}" \
+              --docker-username="${REGISTRY_REDHAT_IO_USERNAME}" \
+              --docker-server="${REGISTRY_REDHAT_IO_SERVER}"
+
   create-secrets:
     steps:
       - run:
@@ -377,6 +387,7 @@ jobs:
       - checkout
       - install-openshift
       - create-secrets
+      - create-redhat-registry-io-secret
       - build-nightly-amp:
           components_imagestream_tag_name: << parameters.components_imagestream_tag_name >>
       - oc-observe


pkg/3scale/amp/manual-templates/amp/build.yml
```diff
@@ -11,12 +11,12 @@ objects:
...
     labels:
       app: zync
-    name: ruby-24-centos7
+    name: ruby-25-ubi7
   spec:
     tags:
     - from:
         kind: DockerImage
-        name: centos/ruby-24-centos7
+        name: registry.access.redhat.com/ubi7/ruby-25
```
eguzki, line 19: From https://access.redhat.com/RegistryAuthentication
```
Over the coming year, Red Hat will standardize on registry.redhat.io and registry.connect.redhat.com as the primary container registries for Red Hat and certified partners, and will eventually decommission registry.access.redhat.com. While customers can continue to use registry.access.redhat.com until it is decommissioned, it is recommended that you prepare to use registry.redhat.io.
```

Using the new `registry.redhat.io` registry requires authenticated access. The circleCI job building images should create secret with the operator developer account credentials living in circleCi. Like this:

```
  create-redhat-registry-io-secret:
    steps:
      - run:
          name: Create registry.redhat.io secret
          command: |
            oc create secret docker-registry threescale-registry-auth \
              --docker-password="${REGISTRY_REDHAT_IO_PASSWORD}" \
              --docker-username="${REGISTRY_REDHAT_IO_USERNAME}" \
              --docker-server="${REGISTRY_REDHAT_IO_SERVER}"
```

miguelsorianod: I think that's not exactly like that @eguzki.

UBI images are a special case where they keep being published in registry.access.redhat.com if I'm not mistaken.

miguelsorianod: https://www.redhat.com/en/blog/introducing-red-hat-universal-base-image
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/building_running_and_managing_containers/index#using_init_red_hat_base_images

eguzki: Should be
`registry.redhat.io/ubi7/ruby-25`

https://catalog.redhat.com/software/containers/detail/5c889d63bed8bd28d0e79ee2?container-tabs=gti&gti-tabs=registry-tokens

guicassolato: Ok, @eguzki. I'll make sure to include that in the workflow and change the base domain of the image. Thanks.

miguelsorianod: For example, operator-sdk itself uses registry.access:
https://github.com/operator-framework/operator-sdk/blob/424a61d56000e6e3d91d352faa1bd4f7c814661f/internal/scaffold/build_dockerfile.go

It seems with the specific case of ubi images it's sort of a "grey" area

eguzki: `registry.access.redhat.com` is deprecated as a registry. Will eventually be decommissioned. If some RedHat blog uses that registry is their fault.

miguelsorianod: It's not a blog only, it's the official documentation in rhel8 and some of the products

eguzki: Then it is not deprecated? If it is not, I am ok with using it.

miguelsorianod: https://access.redhat.com/articles/4238681 another article where it says the authenticated or unauthenticated registry can be used.

Regarding whether it's deprecated or not... in general registry.access.redhat.com is deprecated. However, in the special case of ubi images I'm not so sure about that. UBI are meant to be easily accessible. Using an authenticated registry to get them kind of defeats the purpose and I guess that's why lots of documentation and product are using registry.access.redhat.com. I don't have any official confirmation about being one way or the other though. If you are interested we can try to ask someone that is involved in that

If we were to use registry.redhat.io wouldn't forks experience problems with it as they don't have access to the secret defined in CircleCI?

miguelsorianod: Also, is some external team using the file? is QE using it? If we decide to change to .io we have to tell them too

miguelsorianod: Some additional information I've found: https://bugzilla.redhat.com/show_bug.cgi?id=1719994#c4

eguzki: @miguelsorianod there is no issue about using the new repository `registry.redhat.io/ubi7/ruby-25`.

The job building the nightly images is not run by GH forks and external PR's. 

Even if we needed to run in all PR's (including external PR's), we could use some robot developer account. `registry.redhat.io` is authenticated, but free. Anyone can get developer account credentials. Thus, it is not a big deal exposing some test robot developer account red hat developer credentials in our circleci job.

miguelsorianod: As I said, based on the information I've seen and shown you, I don't see the need to move to registry.redhat.io at the moment in the special case of UBI images. It seems it's going to stay there for the foreseeable future and it adds a layer of authentication to the usage of that template that it's not necessary at all, increasing the needed steps/complexity to use it.

> Even if we needed to run in all PR's (including external PR's), we could use some robot developer account. registry.redhat.io is authenticated, but free. Anyone can get developer account credentials. Thus, it is not a big deal exposing some test robot developer account red hat developer credentials in our circleci job.

I'm not a fan of placing credentials in the repository, even if "anyone can get developer account credentials".

I prefer not doing it at the moment for the reasons I already exposed. If the situation were to change in the future regarding UBI images this could be changed of course.

That said, being limited to the nightly jobs only, and that those are not run on all PRs, it's not a blocking change for me. If you decide to do it cc @3scale/qe has to be aware of this. Also ideally in that case in zync @guicassolato it would be better to be consistent with the decision taken here.

guicassolato: > in zync @guicassolato it would be better to be consistent with the decision taken here.

I think this is particularly relevant to https://github.com/3scale/zync/pull/330.

gsaslis: +1 on no need to move to registry.redhat.io yet. 

Let's only do that when $whoever_responsible deems it absolutely necessary, because using the repo that **requires** auth adds an extra step. 

Yes, it's free, and, yes everyone can get an account, but it doesn't mean that they will. :) 
(they might simply give up)

If we _can_ use a registry that doesn't require auth, let's save everyone the additional hassle.

eguzki: Guys, RH decided to move to authenticated registry. Not a big deal. Get used to it.

gsaslis: I am not sold on that, tbh. (Otherwise, why is the unauthenticated registry around 2+ years later and docs all over still mention the unauthenticated one, as per @miguelsorianod's comments?)

eguzki:
After discussing about the image registry issue, we will use the registry.redhat.io . @miguel prefers using the old unauthenticated one. But not he is not blocking the issue. I will make sure everything works as expected and nightly images are built as always.

PR #362, opened 2020-05-20
Use public images for tests

eguzki:
Production urls are only valid after GA. Before GA, production urls will return 404 Not found because it is on GA when images are made public.

So before GA, the operator stable branch should be using 3scale operand images from the component's stable branches. Even nightly images are wrong. That is because some component could be merging to master some new feature (maybe breaking thing) just after creating component's stable branch.

After GA, before any update in the operator code, the stable branch of the operator should be updated with a PR like this. That is the main reason of this PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -129,14 +129,9 @@ commands:
   deploy-3scale-eval-from-template:
     steps:
       - run:
-          name: Deploy 3scale from amp-eval template with nightly images
+          name: Deploy 3scale from amp-eval template
           command: |
             oc new-app --file pkg/3scale/amp/auto-generated-templates/amp/amp-eval.yml \
-              --param AMP_BACKEND_IMAGE=quay.io/3scale/apisonator:nightly \
-              --param AMP_ZYNC_IMAGE=quay.io/3scale/zync:nightly \
-              --param AMP_APICAST_IMAGE=quay.io/3scale/apicast:nightly \
-              --param AMP_SYSTEM_IMAGE=quay.io/3scale/porta:nightly \
-              --param MEMCACHED_IMAGE=registry.redhat.io/3scale-amp2/memcached-rhel7:3scale2.7 \
               --param WILDCARD_DOMAIN=lvh.me --param TENANT_NAME=3scale
             oc wait --for=condition=available --timeout=-1s $(oc get dc --output=name)
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -123,7 +123,7 @@ commands:
             oc wait --for=condition=available --timeout=-1s $(oc get dc --output=name)
 
             oc get events | egrep ' Failed ' || :
-            oc get events -o json | jq '[.items[] | select(.reason == "Failed") | debug ] | length == 0'  --exit-status
+            oc get events -o json | jq '[.items[] | select(.reason == "Failed") | debug ] | length == 0'
           no_output_timeout: 30m
 
   deploy-3scale-eval-from-template:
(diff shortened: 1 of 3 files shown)


miguelsorianod:
This works now because 2.8 is already GA.

The reason nightly were used for tests is due for a new release we need to test latest available images (nightly) and the registry.redhat.io images are not still released.

What's the idea  from now on? For new releases we code the nightly images and after GA we open a PR to change it as this one? Keep in mind that if that's the case, the branch will always be on commits newer than the tagged ones.

eguzki:
I know. That is why I open this PR. Otherwise any new change would trigger a tests running operand nightly images instead of release images.

miguelsorianod:
So what's the way to go for future releases, what I mentioned before?:

For new releases we code the nightly images and after GA we open a PR to change it as this one? Keep in mind that if that's the case, the branch will always be on commits newer than the tagged ones.

eguzki:
The operator commits are newer than the tagger ones. This is expected in case there is a newer operator release. 

But keep in mind we are talking about the operands urls.

miguelsorianod:
> The operator commits are newer than the tagged ones. This is expected in case there is a newer operator release.

I mean after the release, for a given release branch.

eguzki:
> I mean after the release, for a given release branch that is already GA. I'm not talking about master or new unreleased product branches.

Yes. The stable branch has to be updated with production base urls of the operands. It you change a bit of the operator code and you need to prepare new operator release, tests should be done using correct operand urls, not nightly urls. Nightly urls refer to operands in  some other 3scale release not the same as the current stable branch.

miguelsorianod:
> Yes. The stable branch has to be updated with production base urls of the operands. It you change a bit of the operator code and you need to prepare new operator release, tests should be done using correct operand urls, not nightly urls. Nightly urls refer to operands in some other 3scale release not the same as the current stable branch.

When you prepare a new release branch for a new operator release the tests cannot be done using correct operand urls. The new image urls are still not there until GA, and the urls of the previous release cannot be used as they contain old code. Only elements for a new branch to be prepared for a release I can see being used are nightly urls (until GA)

eguzki:
What are you asking to me?

miguelsorianod:
The following:
> What's the idea from now on? For new releases we code the nightly images and after GA we open a PR to change it as this one? Keep in mind that if that's the case, the branch will always be on commits newer than the tagged ones.

eguzki:
Already answered. https://github.com/3scale/3scale-operator/pull/362#issuecomment-631560506

More questions?

miguelsorianod:
The answer you quoted is answering that you are opening this PR to change to use productized images.

For a new release are we going to open the new branch with nightly images coded until GA, at which point we will another PR after GA (tagging) and change it to the productized images URLs?

Previous answers are confusing me like:
> It you change a bit of the operator code and you need to prepare new operator release, tests should be done using correct operand urls, not nightly urls.

That's only valid after GA, not when you prepare a new operator release

eguzki:
Yes. After GA. 

Production urls are only valid after GA. Before GA, production urls will return 404 Not found because it is on GA when images are made public.

So before GA, the operator stable branch should be using 3scale operand images from the component's stable branches. Even nightly images are wrong. That is because some component could be merging to master some new feature (maybe breaking thing) just after creating component's stable branch. 

After GA, before any update in the operator code, the stable branch of the operator should  be updated with a PR like this. That is the main reason of this PR. 

Sorry for the confusion. Let me know if you need more info.

miguelsorianod:
Thank you for the clarification.

> So before GA, the operator stable branch should be using 3scale operand images from the component's stable branches. Even nightly images are wrong. That is because some component could be merging to master some new feature (maybe breaking thing) just after creating component's stable branch.

Are components (operands) creating stable images of new unreleased content that will go into a new 3scale release that we can use? up until now nightly images were being used in the new release branches before GA because there was no such thing. If so, where can we find them?

Thank you.

eguzki:
Either we:
* ask them to create them and keep them updated for any update in the stable branch. 
* our stable branch has a nightly job to create stable release components image tag. For example apicast-2.8-stable:nightly

miguelsorianod:
:+1: ok, so this is an approach that cannot be done at the moment unless some changes are made.

I guess if we were to decide to go with the second proposal, we would need the information from each of those components about what are their stable branches each release.

Let's merge this and open a Jira issue to define the concrete way to go forward.

PR #363, opened 2020-05-20
add pullSecret to buildconfig to access private repos

eguzki:
Image builder also needs pullsecret to access private repos.

### DIFF
--- a/pkg/3scale/amp/manual-templates/amp/build.yml
+++ b/pkg/3scale/amp/manual-templates/amp/build.yml
@@ -39,7 +39,7 @@ objects:
     strategy:
       dockerStrategy:
         pullSecret:
-          name: threescale-registry-auth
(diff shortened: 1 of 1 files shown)


miguelsorianod:
Looks good to me. Maybe would it be a good idea to add it as an OpenShift Template Parameter so a user of the template can set any secret name they want? (we can set a default to the parameter being threescale-registry-auth)

eguzki:
IMO it does not have any use to add template parameter because build.yaml is for internal use only. But I added anyway.

miguelsorianod:
> IMO it does not have any use to add template parameter because build.yaml is for internal use only. But I added anyway.

I think we might not be the only 3scale subteam using the build.yml. I think QE were relying on it and some other devs that wanted to manually build their images. At least a while ago.

PR #367, opened 2020-05-22
Metering Labels

eguzki:
Metering labels

https://docs.google.com/document/d/1bd9VnZ9jKwDBOOKAJltAtvhz3COUYYmRe3-BGE6fDGE/edit?ts=5de56bf6

- [x] 3scale-operator
- [x] apicast-production
- [x] apicast-staging
- [x] backend-cron
- [x] backend-listener
- [x] backend-redis
- [x] backend-worker
- [x] system-app
- [x] system-memcache
- [x] system-mysql
- [x] system-postgresql          
- [x] system-redis
- [x] system-sidekiq
- [x] system-sphinx
- [x] zync
- [x] zync-database
- [x] zync-que


eguzki:
Implemented first for apicast deployment.

@miguelsorianod Let's verify implementation before going for other deployments.

pkg/3scale/amp/component/apicast_options.go
```diff
@@ -16,6 +16,8 @@ type ApicastOptions struct {
 	StagingResourceRequirements    v1.ResourceRequirements `validate:"-"`
 	ProductionReplicas             int32
 	StagingReplicas                int32
+	StagingLabels                  map[string]string `validate:"required"`
```
miguelsorianod, line 19: The labels `app`, `threescale_component`, `threescale_component_element` are used on different types of K8s objects. You can see them in Services, DeploymentConfigs(don't confuse it with PodTemplates in the DCs), ConfigMaps, PDBs.

Maybe we should name this options as StagingPodTemplateLabels and ProductionPodTemplateLabels?

Additionally, is there some way we can reuse the values of labels for the other K8s objects? If you look at the pkg/3scale/amp/component/apicast.go you'll see the labels are mentioned are found in other objects too.

pkg/3scale/amp/operator/apicast_options_provider.go
```diff
@@ -22,16 +24,23 @@ func NewApicastOptionsProvider(apimanager *appsv1alpha1.APIManager) *ApicastOpti
 }
 
 func (a *ApicastOptionsProvider) GetApicastOptions() (*component.ApicastOptions, error) {
+	imageOpts, err := NewAmpImagesOptionsProvider(a.apimanager).GetAmpImagesOptions()
```
miguelsorianod, line 27: This is weird but the alternative would be duplicate some part of the code of GetAmpImageOptions. I guess we can leave it this way to avoid duplication. Let's hope we don't end up having K8s requests in the GetAmpImagesOptions in the long term as we have with other options providers.

pkg/3scale/amp/operator/apicast_options_provider.go
```diff
@@ -49,3 +58,21 @@ func (a *ApicastOptionsProvider) setReplicas() {
 	a.apicastOptions.ProductionReplicas = int32(*a.apimanager.Spec.Apicast.ProductionSpec.Replicas)
 	a.apicastOptions.StagingReplicas = int32(*a.apimanager.Spec.Apicast.StagingSpec.Replicas)
 }
+
+func (a *ApicastOptionsProvider) setStagingLabels(image string) map[string]string {
+	labels := helper.MeteringLabels("apicast-staging", helper.ParseVersion(image), helper.ApplicationType)
+	labels["app"] = a.apicastOptions.AppLabel
+	labels["threescale_component"] = "apicast"
```
miguelsorianod, line 65: We'll end up with duplicated key/values between templates and operator for all components. I'd be nice to not have them duplicate. Can you think of some way to avoid that? I guess constants in the components package would be a way. Not sure whether we would have lots of them though. What do you think?

pkg/3scale/amp/component/apicast_options.go
eguzki: done

pkg/3scale/amp/operator/apicast_options_provider.go
eguzki: done

eguzki: duplication of templates is not a big deal. They will soon be removed

eguzki:
resolved duplication issues.

miguelsorianod:
Looks good, it can be applied to others :+1:

eguzki:

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -614,6 +628,97 @@ func (u *UpgradeApiManager) upgradeSystemMasterApicastSecret() (reconcile.Result
...
+	if !u.apiManager.IsExternalDatabaseEnabled() {
+		redis, err := Redis(u.apiManager)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		deploymentConfigs = append(deploymentConfigs, redis.SystemDeploymentConfig())
+		deploymentConfigs = append(deploymentConfigs, redis.BackendDeploymentConfig())
+	}
+
+	if u.apiManager.IsSystemPostgreSQLEnabled() {
```
miguelsorianod, line 677: Might it be that this check is not completely right?

IsSystemPostgreSQLEnabled checks that external databases is not set and postgresql field is set.
The else condition happens when one of the condition parts is false.
This means that in the else we could have external databases enabled. In that case we should not provide system mysql.

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -614,6 +628,97 @@ func (u *UpgradeApiManager) upgradeSystemMasterApicastSecret() (reconcile.Result
...
+		deploymentConfigs = append(deploymentConfigs, systemMySQL.DeploymentConfig())
+	}
+
+	for _, desired := range deploymentConfigs {
+		err = u.ensurePodTemplateLabels(desired)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+
```
miguelsorianod, line 697: It would probably be a good idea when changes happen so the upgrade part is reevaluated again to ensure changes have been applied and there are no new changes, before finishing the upgrade completely. We could do it at the end of the for loop if there's been a change or at each iteration if there's been a change.

What do you think?

pkg/3scale/amp/operator/upgrade.go
eguzki: good catch. Fixing it

pkg/3scale/amp/operator/upgrade.go
eguzki: done

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -614,6 +628,103 @@ func (u *UpgradeApiManager) upgradeSystemMasterApicastSecret() (reconcile.Result
...
+	err := u.Client().Get(context.TODO(), types.NamespacedName{Name: desired.Name, Namespace: u.apiManager.Namespace}, existing)
+	if err != nil {
+		return false, err
+	}
+
+	updated := false
+	if !reflect.DeepEqual(existing.Spec.Template.Labels, desired.Spec.Template.Labels) {
+		diff := cmp.Diff(existing.Spec.Template.Labels, desired.Spec.Template.Labels)
+		u.Logger().V(1).Info(fmt.Sprintf("DC %s template lables changed: %s", desired.Name, diff))
+		existing.Spec.Template.Labels = desired.Spec.Template.Labels
```
miguelsorianod, line 717: I think in the case of Labels (and annotations too) we should follow a merge approach instead of a set approach like it's done in `BaseAPIManagerLogicReconciler.ReconcileResource` (through the APIManagerMutator mutateFn).
The comparison in this code would change too (instead of checking whether labels are equal it should be checked that the desired subset is equal)

miguelsorianod:
Applied changes look good to me :+1: 

I noticed another change and I left feedback on it.

pkg/3scale/amp/operator/upgrade.go
eguzki: :+1: 

Changed to merge labels

eguzki:
All changes applied and feedback answered

PR #369, opened 2020-05-27
Remove obsolete APIManagerRestore status fields

miguelsorianod:
Obsolete since PR #361


miguelsorianod:
@eguzki, performed some additional to remove dead code and unneeded code

PR #370, opened 2020-05-27
Fix APIManagerRestore's shared secret step dependencies

miguelsorianod:
Fix APIManagerRestore shared secret dependency between steps
    
Several restore steps always expected the shared secret
to always be available. However, one of the last steps is removing
them. That made steps fail in some situations where reconciliation
loop was reexecuted again.
This commit adds a new Status field 'APIManagerToRestoreRef' that
contains the name of the APIManager to restore, and the steps
that needed the secret now can do their tasks using this flag.

Don't merge until PR #369 is merged


pkg/controller/apimanagerrestore/apimanagerrestore_logic_reconciler.go
```diff
@@ -321,6 +330,21 @@ func (r *APIManagerRestoreLogicReconciler) apiManagerFromSharedBackupSecret() (*
 }
 
 func (r *APIManagerRestoreLogicReconciler) reconcileRestoreAPIManager() (reconcile.Result, error) {
+	apiManagerToRestoreName := r.cr.Status.APIManagerToRestoreRef.Name
```
miguelsorianod, line 333: At this point, if there are no programming errors APIManagerToRestoreRef should never be nil and Name should be a non-empty string. Thus, no checks related to that are performed each time the attribute is referenced in steps that are after the step that should set this status value

pkg/apis/apps/v1alpha1/apimanagerrestore_types.go
```diff
@@ -30,6 +30,10 @@ type APIManagerRestoreStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// +optional
+	APIManagerToRestoreRef *v1.LocalObjectReference `json:"apimanagerToRestore,omitempty"`
```
miguelsorianod, line 35: An alternative to LocalObjectReference would be ObjectReference, which has the namespace field too that we could set. The thing with ObjectReference is that it also has a bunch of other fields like 'UID types.UID', 'resourceVersion', ...
What do you think? do you think LocalObjectReference good enough?

miguelsorianod:
Ready for review @eguzki

pkg/controller/apimanagerrestore/apimanagerrestore_logic_reconciler.go
```diff
@@ -321,6 +330,21 @@ func (r *APIManagerRestoreLogicReconciler) apiManagerFromSharedBackupSecret() (*
...
+
+	err := r.GetResource(types.NamespacedName{Name: apiManagerToRestoreName, Namespace: r.cr.Namespace}, &appsv1alpha1.APIManager{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			apiManagerAlreadyCreated = false
+		}
+		return reconcile.Result{}, err
+	}
+
+	if apiManagerAlreadyCreated {
```
eguzki, line 344: this condition is always evaluated to `true`

miguelsorianod: It shouldn't. In line 339 I set it to false :thinking:

pkg/controller/apimanagerrestore/apimanagerrestore_logic_reconciler.go
miguelsorianod: The reason I haven't add it as a comment is due to it appears in several places and it's kind of a "way of proceeding".
Adding it on all occurrences each time we reference the attribute can be kind of verbose. What do you want to do?

miguelsorianod:
@eguzki feedback answered

eguzki:
As I see, this status field should only be used in one step. CAnnot be shared between steps or they will be coupled. It is desirable to keep steps/stages as decoupled as possible.

The step main structure should be:
```
if backup.status.Field != nil{
   // job done
   return
}
doSomeTask()
backup.status.Field = Done
client.Status().update(backup)
```

pkg/controller/apimanagerrestore/apimanagerrestore_logic_reconciler.go
eguzki: Not straightforward/obvious comments are always welcome.

pkg/controller/apimanagerrestore/apimanagerrestore_logic_reconciler.go
eguzki: are you sure? :eyes:

miguelsorianod: :+1: oh I see, I'm always returning when I set it to false. Fixing it...

miguelsorianod:
> As I see, this status field should only be used in one step. CAnnot be shared between steps or they will be coupled. It is desirable to keep steps/stages as decoupled as possible.

In this specific case I don't see the problem. Several tasks need the APIManager name to perform their tasks. Once that is available it will always be as it won't be changed by any task. It's also an informational field.

I agree with the general approach. It's basically more or less going back to the implementation that we I had implemented before trying to rely on the k8s jobs uniquely.

pkg/controller/apimanagerrestore/apimanagerrestore_logic_reconciler.go
miguelsorianod: I added it on the first occurrencly only. There's no point on keep repeating the same comment each time that attribute is referenced

pkg/controller/apimanagerrestore/apimanagerrestore_logic_reconciler.go
miguelsorianod: Fixed

miguelsorianod:
Ready @eguzki

PR #372, opened 2020-05-27
Add backup restore crd tests

miguelsorianod:
Add APIManagerBackup and APIManagerRestore to the CRD tests

### DIFF
--- a/deploy/crds/apps.3scale.net_v1alpha1_apimanagerbackup_cr.yaml
+++ b/deploy/crds/apps.3scale.net_v1alpha1_apimanagerbackup_cr.yaml
@@ -3,5 +3,6 @@ kind: APIManagerBackup
 metadata:
   name: example-apimanagerbackup
 spec:
-  # Add fields here
-  size: 3
+  backupSource:
+    persistentVolumeClaim:
+      volumeName: "mypreexistingvolume"
--- a/deploy/crds/apps.3scale.net_v1alpha1_apimanagerrestore_cr.yaml
+++ b/deploy/crds/apps.3scale.net_v1alpha1_apimanagerrestore_cr.yaml
@@ -3,5 +3,7 @@ kind: APIManagerRestore
 metadata:
   name: example-apimanagerrestore
 spec:
-  # Add fields here
-  size: 3
+  restoreSource:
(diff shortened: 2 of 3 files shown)


miguelsorianod:
While executing the tests the following CRD test fails: `TestCompleteCRD`

It fails with errors of the following style:
```
--- FAIL: TestCompleteCRD (0.02s)
    crd_validation_test.go:77: 
        	Error Trace:	crd_validation_test.go:77
        	Error:      	Discrepancy between CRD and Struct
        	Test:       	TestCompleteCRD
        	Messages:   	CRD: apps.3scale.net_apimanagerbackups_crd.yaml: Missing or incorrect schema validation at /spec/backupSource/persistentVolumeClaim/resources/requests, expected type object
    crd_validation_test.go:77: 
        	Error Trace:	crd_validation_test.go:77
        	Error:      	Discrepancy between CRD and Struct
        	Test:       	TestCompleteCRD
        	Messages:   	CRD: apps.3scale.net_apimanagerbackups_crd.yaml: Missing or incorrect schema validation at /spec/backupSource/persistentVolumeClaim/resources/requests/i, expected type object
```

I think that the problem is that the `schema.GetMissingEntries` method is detecting unexported struct fields as missing entries, which are not in the openapi spec understandably.
I don't know if there's much that we can do in that regard, as the method GetMissingEntries does not have any parameter that controls exported/unexported field reporting, and the result that it returns only contains the field path and the type.

miguelsorianod:
A solution that I can think of is just excluding certain path prefixes. It would require maintenance, as changes in the CRDs would impact that test.

miguelsorianod:
Looking at several operators that make use of the `MissingEntries` methods:
* Some don't use attributes with unexported fields so they don't reproduce the problem
* Some directly exclude status section so they don't reproduce the problem (that seems wrong to me)
* Some hardcode the paths to exclude as I proposed before

miguelsorianod:
Based on the possible approaches that I can think of and what I've observed in other operators I implemented the missing entries exclusion path omissions mechanism, adding the ones where we were reproducing the described issue.
This has the downsides already explained in previous messages.

miguelsorianod:
Ready for review @eguzki

PR #374, opened 2020-06-02
Add  APIManagerBackup cr cleanup doc

miguelsorianod:


miguelsorianod:
This is not ideal but for the moment to free the used elements during the backup should be enough

PR #377, opened 2020-06-03
Update OLM-related files with APIManagerBackup/Restore changes

miguelsorianod:


miguelsorianod:
When updating CSV contents I've detected there are other pending changes unrelated to APIManagerBackup/Restore but I've not included them as I've seen them included in the #367 PR

PR #378, opened 2020-06-03
Update GetResource to get namespacedname info from the ObjectKey

miguelsorianod:
ObjectInfo is often empty when calling GetResource so the logging of the object's name and namespace ended up often being empty

### DIFF
--- a/pkg/reconcilers/base_reconciler.go
+++ b/pkg/reconcilers/base_reconciler.go
@@ -3,6 +3,7 @@ package reconcilers
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/3scale/3scale-operator/pkg/common"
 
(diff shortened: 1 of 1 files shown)


pkg/reconcilers/base_reconciler.go
```diff
@@ -122,7 +122,7 @@ func (b *BaseReconciler) ReconcileResource(obj, desired common.KubernetesObject,
 }
 
 func (b *BaseReconciler) GetResource(objKey types.NamespacedName, obj common.KubernetesObject) error {
-	b.Logger().Info(fmt.Sprintf("Get object %s", common.ObjectInfo(obj)))
```
eguzki, line 125: 1) `common.ObjectInfo` returns type/name, not name/namespace. 
2) Indeed, sometimes only the type is empty. I think it only happens with our CRD's. Try to find out why
3) base_reconciler should be consistent. We should log the same for Get, Create, Update, Delete, UpdateStatus

miguelsorianod: When getting the object the user of the method often sends an empty object, where TypeMeta has not been set. I guess it happens because of this

miguelsorianod: See:

```
	existing := batchv1.Job{}
	fmt.Printf("objectinfo: '%s'\n", common.ObjectInfo(&existing))
```

```
objectinfo: '/'
```

miguelsorianod: cc @eguzki

miguelsorianod: I've changed the implementation to get the kind from the object type

PR #379, opened 2020-06-03
[WIP] Create backuprestore pipeline elems

miguelsorianod:
In APIManagerBackup and APIManagerRestore the different tasks can be viewed as a series of pipeline steps. It would be great to abstract this view into code.

This PR provides a possible solution. Let's discuss about whether it would be suited for our use cases and iterate from here in case we see it's a good idea.


miguelsorianod:
I've started by trying to reimplement APIManagerRestore with the new pipeline approach.
For APIManagerRestore with the current state of the PR it should be implemented already but it has not been tested at all so later surprises might appear.

Let's discuss the implemented approach first though.

miguelsorianod:
cc @eguzki for a first round of reviews. Maybe it would be a good idea to sit down together and perform pair review

PR #380, opened 2020-06-08
Add step to cleanup Backup/Restore K8s jobs

miguelsorianod:
Volumes that are mounted in the Backup/Restore jobs (like system-storage) cannot be deleted even if the jobs (and its pods) are in 'Completed' state.

This prevented the full deletion of an APIManager object while an APIManagerBackup existed.

Add a step that cleanups the K8s jobs.

### DIFF
--- a/deploy/crds/apps.3scale.net_apimanagerbackups_crd.yaml
+++ b/deploy/crds/apps.3scale.net_apimanagerbackups_crd.yaml
@@ -110,6 +110,11 @@ spec:
                 - type
                 type: object
               type: array
+            mainStepsCompleted:
+              description: Set to true when main steps have been completed. At this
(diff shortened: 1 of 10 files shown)


miguelsorianod:
We added a bool to control when all the "main" steps of the backup are completed. The reason for that is that we cannot execute the main steps when the cleanup job is to be executed / is executed, as the previous steps check their own completion state based on the state of the K8s jobs they execute and the cleanup job deletes the jobs themselves.

We also have decided not to execute the cleanup job by checking whether the  'completed' bool is set, as we don't want to execute anything after 'completed' is set.

miguelsorianod:
Added minor commit that updates backup/restore documentation to remove the part where we required to delete the APIManagerBackup, which is no longer needed

PR #382, opened 2020-06-08
Create APIManagerBackup/Restore unique predictable K8s job names from CRs UIDs

miguelsorianod:
Until now K8s Job names created by APIManagerBackup and APIManagerRestore controllers were created using the APIManagerBackup/Restore CR name as part of the K8s Job and a unique job suffix. This was done in order to have predictable Job names and with multiple existing APIManagerBackup/APIManagerRestore CR objects at the same time.

The problem with that approach is that the size of the name of the resulting K8s was depending on the CR name length that the user provided when creating the CRs. K8s labels have a limitation of 63 characters and the combination of the CR name + the unique job suffix made K8s Job names reach the limit of characters for labels in some occasions.

The new approach creates K8s job names based on a prefix job name (defined by us in code) and the UID of the APIManagerBackup/APIManagerRestore. This makes Job names unique among CR instantiations and allows us to control the length of the created Job names (UIDs are always 36 characters long)

For example, if a user has created the following APIManagerBackup:

```
kind: APIManagerBackup
metadata:
  name: example-apimanagerbackup-pvc
spec:
  backupDestination:
    persistentVolumeClaim:
      resources:
        requests: "1Gi"
```

and the APIManagerBackup has the UID `174fdad6-9121-4217-9a0c-19b2f52555d6` assigned once created.

The K8s Jobs created would be:
```
msoriano@localhost:~/go/src/github.com/3scale/3scale-operator/example-manifests/backuprestore (backuprestore-uid-based-jobnames)$ oc get jobs
NAME                                                          COMPLETIONS   DURATION   AGE
backup-apimanager-cr-174fdad6-9121-4217-9a0c-19b2f52555d6     1/1           13s        21s
backup-cfgmaps-secrets-174fdad6-9121-4217-9a0c-19b2f52555d6   1/1           28s        53s
backup-system-fs-pvc-174fdad6-9121-4217-9a0c-19b2f52555d6     0/1           6s         6s
```

And the K8s Pods that would be automatically created by the Jobs K8s controller would be:
```
backup-apimanager-cr-174fdad6-9121-4217-9a0c-19b2f52555d6-4q6wx   0/1     Completed           0          27s
backup-cfgmaps-secrets-174fdad6-9121-4217-9a0c-19b2f52555d6bpnp   0/1     Completed           0          59s
backup-system-fs-pvc-174fdad6-9121-4217-9a0c-19b2f52555d6-skt9r   0/1     ContainerCreating   0          12s
```

Notice how in the case of pods, random characters are added by K8s Jobs controller automatically (this is a behavior that always happens with pods managed by other K8s elements like Jobs, DeploymentConfigs, Jobs...), and that the original name might be truncated if necessary to allow the random characters that K8s. This can be observed in the example above ( a '-' character is removed in one of the pod names above). That is expected and shouldn't be a problem as the pods are managed by the K8s Job controller.

### DIFF
--- a/pkg/backup/apimanager_backup.go
+++ b/pkg/backup/apimanager_backup.go
@@ -102,6 +102,11 @@ func (b *APIManagerBackup) BackupSecretsAndConfigMapsToPVCJob() *batchv1.Job {
 		return nil
 	}
 
+	jobName, err := helper.UIDBasedJobName("backup-cfgmaps-secrets", b.options.APIManagerBackupUID)
+	if err != nil {
+		panic(err)
+	}
+
 	var completions int32 = 1
 	return &batchv1.Job{
 		TypeMeta: metav1.TypeMeta{
@@ -109,7 +114,7 @@ func (b *APIManagerBackup) BackupSecretsAndConfigMapsToPVCJob() *batchv1.Job {
 			Kind:       "Job",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      fmt.Sprintf("job-backup-cfgmaps-secrets-%s", b.options.APIManagerBackupName),
+			Name:      jobName,
 			Namespace: b.options.Namespace,
 		},
 		Spec: batchv1.JobSpec{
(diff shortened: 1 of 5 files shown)


miguelsorianod:
ready for review @eguzki .

PD: If you want to suggest a Job prefix name change for one or more of the existing K8s Jobs names defined now it's an opportunity to do it

PR #383, opened 2020-06-09
[WIP] Increase currently supported embedded/external databases granularity

miguelsorianod:
This PR performs a restructuration of the APIManager CRD to allow configurability at individual level for the databases that currently can be configured external so they can be individually selected as embedded or external instead of an all-or-nothing behavior, and also having configuration-level specific to each of them (in order to accommodate for new functionality that is specific to each DB).

Remarkable aspects of this PR:
* Deprecated the  `spec.system.database.mysql` and `spec.system.database.postgresql` sections. They have been replaced by the sections `spec.system.database.embedded` and `spec.system.database.external` (mutually exclusive) sections. Inside `spec.system.database.embeded` the `spec.system.database.embedded.mysql` and `spec.system.database.embedded.postgresql` sections have been added.
* Deprecated the fields `spec.system.redisImage`, `spec.backend.redisImage`
* Deprecated the highAvailability attribute that controlled whether all externally configurable DBs were external or not (in an all-or-nothing behavior for all of them) and it has been replaced by allowing each of those DBs to be configured as embedded or external individually
* Split the 'redis' component into SystemRedis and BackendRedis. The reason for this is that now we allow to control individually whether we want each of them being embedded or not.
* Added a CommonEmbeddedRedis component that includes the creation of data that is shared between SystemRedis and BackendRedis (at this moment, the 'redis-config' ConfigMap). This has resulted in additions/changes in the component, templates and operator packages.
* Upgrade procedure to migrate CR fields
* Renamed deprecated fields in CRD Go code with the 'Deprecated' prefix. This has changed the names at Go code level but NOT the json field names (to keep compatibility)
* This changes do NOT apply to templates. It is an operator-only functionality

PR #384, opened 2020-06-10
Add affinity and tolerations APIManager configurability for DeploymentConfigs

miguelsorianod:
This PR intends to add configurability of affinity and tolerations in APIManager for DeploymentConfigs (database and non-database ones)

The implementation has the following remarkable points:
* Reconciliation of 'Affinity' and 'Tolerations' has been implemented for all DeploymentConfigs. This means the reconciliation is executed even if the DeploymentConfig does not have configurability at APIManager level and even if the DeploymentConfig is a database one. It still should work without issues as the default values for those are nil and no changes should be detected. The only new difference in behaviour for database DCs and DCs where APIManager configurability is not provided is that if the user manually changes the values for them directly in the DCs the changes will be reverted by the operator to nil for them.
* Affinity and Tolerations are deep objects (several levels and multiple fields) and contain arrays (where elements order matters). Reconciliation has been implemented with a DeepEqual comparison and seems to work ok. Changing the order of the specified array elements in APIManager is detected as a change even if the elements are the same, so a reconciliation happens in that case too. This behaviour is no different than changing array elements in a DeploymentConfig in OpenShift.
* Operator-only functionality

### DIFF
--- a/deploy/crds/apps.3scale.net_apimanagers_crd.yaml
+++ b/deploy/crds/apps.3scale.net_apimanagers_crd.yaml
@@ -41,9 +41,676 @@ spec:
                   type: boolean
                 productionSpec:
                   properties:
+                    affinity:
+                      description: Affinity is a group of affinity scheduling rules.
+                      properties:
+                        nodeAffinity:
+                          description: Describes node affinity scheduling rules for
+                            the pod.
+                          properties:
+                            preferredDuringSchedulingIgnoredDuringExecution:
+                              description: The scheduler will prefer to schedule pods
+                                to nodes that satisfy the affinity expressions specified
+                                by this field, but it may choose a node that violates
+                                one or more of the expressions. The node that is most
+                                preferred is the one with the greatest sum of weights,
+                                i.e. for each node that meets all of the scheduling
+                                requirements (resource request, requiredDuringScheduling
+                                affinity expressions, etc.), compute a sum by iterating
+                                through the elements of this field and adding "weight"
+                                to the sum if the node matches the corresponding matchExpressions;
+                                the node(s) with the highest sum are the most preferred.
+                              items:
+                                description: An empty preferred scheduling term matches
+                                  all objects with implicit weight 0 (i.e. it's a
+                                  no-op). A null preferred scheduling term matches
+                                  no objects (i.e. is also a no-op).
+                                properties:
+                                  preference:
+                                    description: A node selector term, associated
+                                      with the corresponding weight.
+                                    properties:
+                                      matchExpressions:
+                                        description: A list of node selector requirements
+                                          by node's labels.
+                                        items:
+                                          description: A node selector requirement
+                                            is a selector that contains values, a
+                                            key, and an operator that relates the
+                                            key and values.
+                                          properties:
+                                            key:
+                                              description: The label key that the
+                                                selector applies to.
+                                              type: string
+                                            operator:
+                                              description: Represents a key's relationship
+                                                to a set of values. Valid operators
+                                                are In, NotIn, Exists, DoesNotExist.
+                                                Gt, and Lt.
+                                              type: string
+                                            values:
+                                              description: An array of string values.
+                                                If the operator is In or NotIn, the
+                                                values array must be non-empty. If
+                                                the operator is Exists or DoesNotExist,
+                                                the values array must be empty. If
+                                                the operator is Gt or Lt, the values
+                                                array must have a single element,
+                                                which will be interpreted as an integer.
+                                                This array is replaced during a strategic
+                                                merge patch.
+                                              items:
+                                                type: string
+                                              type: array
+                                          required:
+                                          - key
+                                          - operator
+                                          type: object
+                                        type: array
+                                      matchFields:
+                                        description: A list of node selector requirements
+                                          by node's fields.
+                                        items:
+                                          description: A node selector requirement
+                                            is a selector that contains values, a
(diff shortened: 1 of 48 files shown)


eguzki:
Looks go. 
Go ahead and implement for all 3scale components. 

For database DC as well.

pkg/reconcilers/deploymentconfig.go
```diff
@@ -43,6 +44,12 @@ func GenericDeploymentConfigMutator(existingObj, desiredObj common.KubernetesObj
 	tmpUpdate = DeploymentConfigContainerResourcesReconciler(desired, existing)
 	update = update || tmpUpdate
 
+	tmpUpdate = DeploymentConfigAffinityReconciler(desired, existing)
```
eguzki, line 47: Why not Install Only?

miguelsorianod: It could be install only too I guess.
The parameters are inherently configurable so I decided to implement it configurable.

miguelsorianod:
I've added fields and implementations for both non-db and db DCs.

As originally commented in issue `THREESCALE-5053` there are some elements that do not fit in any specific preexisting section:

system-memcached. I've added two fields in the main 'System' section:
system.memcachedAffinity
system.memcachedTolerations

system-sphinx. I've added a new section called 'sphinx' in 'system'
system.sphinx.affinity
system.sphinx.tolerations

zync-database. I've added two fields in the main 'Zync' section:
zync.databaseAffinity
zync.databaseTolerations

The reasons why sphinx has its own section and the others don't are:
* I considered the other ones 'databases' and not sphinx. To be honest I think it could be argued whether sphinx is a database or not or whether potentially could be external or not. It's in sort of a middleground between application and a database.
* I'm thinking on memcached and zync-database potentially being external (zync-database has been requested by some people in the past) so I didn't want to create a new section for fields that might only apply in an 'embedded-only' context. If the PR https://github.com/3scale/3scale-operator/pull/383 related to increased database granularity was performed I'd just have created sections for them too and add them inside an 'embedded' subsection but it's not the case as the changes there have been at least temporarily delayed so I decided to add them directly in the main section and deprecate them in the future if we end up adding granularity for embedded/external DBs.
* zync-database and system-memcached already have fields in their corresponding main-level component (postgresqlimage field and memcachedImage field)

I also have some doubts regarding the naming of memcached and zync postgresql. If you look at the Jira issue `THREESCALE-5053` I already expressed my concern:

memcached:  maybe a more generic term name than memcached that is not limited to memcached implementation in case it changes in the future (for example being in redis instead of memcached)? It's true though that we already have memcached-named fields (memcachedImage)
sphinx: maybe a more generic term name than sphinx that is not limited to sphinx implementation in case it changes in the future?
zync database: maybe call it databaseSpec.postgresSQL? maybe postgreSQL? will it potentially change or have possible different dbs? We already have a postgreSQLImage field but might be a good idea to just name it databaseXXX (although we would end up having different namings between the image and the affinity configs then).

Finally, keep in mind that depending on what changes are performed it might additional code migrations (due to possible deprecations of fields, changes in current schema, ...)

Please let's review the current implementation before continuing.

Thank you.

eguzki:
For unittest, you can "initialize" your apimanager objects calling `SetDefaults`. Then the context will be closer to production code running.

deploy/crds/apps.3scale.net_apimanagers_crd.yaml
```diff
@@ -41,19 +41,1353 @@ spec:
                   type: boolean
                 productionSpec:
                   properties:
+                    affinity:
```
eguzki, line 44: Definitely this CRD file is not human readable anymore. It is already too big. 

We have to trust on tooling

pkg/3scale/amp/operator/system_mysql_reconciler.go
```diff
@@ -25,7 +25,7 @@ func (r *SystemMySQLReconciler) Reconcile() (reconcile.Result, error) {
 	}
 
 	// DC
-	err = r.ReconcileDeploymentConfig(systemMySQL.DeploymentConfig(), reconcilers.CreateOnlyMutator)
```
miguelsorianod, line 28: Using reconcilers.CreateOnlyMutator seemed originally wrong, as  SystemPostgreSQL Reconciler was using `DeploymentConfigResourcesMutator` instead of `reconcilers.CreateOnlyMutator`.

In new reconciler used in this case I've made it reconcile resources too

miguelsorianod:
New changes made:
* Updated mutators used for DCs that weren't using the Generic Mutator
* Added unit tests
* Updated documentation

Ready for review @eguzki

pkg/3scale/amp/operator/system_mysql_reconciler.go
eguzki: :+1:

eguzki:
Overall LGTM.

I miss some documentation example. Updating API reference is not enough.

Something like, let's configure: 
* node affinity A for apicast, backend listener and backend redis
* node affinity B for the remaining components

I guess this is going to be useful eventually.

miguelsorianod:
> Overall LGTM.
> 
> I miss some documentation example. Updating API reference is not enough.
> 
> Something like, let's configure:
> 
>     node affinity A for apicast, backend listener and backend redis
>     node affinity B for the remaining components
> 
> I guess this is going to be useful eventually.

Being advanced use cases I considered there was not much point on having specific examples for it but I've added an example anyway :+1:

PR #385, opened 2020-06-11
Apicast grafana dashboards

roivaz:
This PR adds templates for two apicast dashboards:

* A main app dashboard with apicast process information and global http metrics
* A "services" dashboard that allows users to select a system service and review http statistics for that specific service. For this dashboard to show data, the apicast "APICAST_EXTENDED_METRICS" environment variable needs to be set to "true"


eguzki:
Looks very good. 

It needs some integration I will do for you.

Let's try this and we can iterate further when we check dashboards deployed by the operator.

good job :medal_military:

PR #386, opened 2020-06-11
Add configurable Storage Class for remaining PVCs

miguelsorianod:
Add configurable storage class for the following database PVCs:
* system-redis
* backend-redis
* system-database (MySQL and PostgreSQL)

Remarkable points:
* Even if the corresponding storageclass name is changed by a user once already set, it will not be fully reconciled (as it happens with System's FileStorage PVC)
* Those sections/fields only take effect when highAvailability.enabled is not true, as they are embedded-database-only specific fields
* Operator-only functionality


miguelsorianod:
@eguzki ready for a review before continuing implementation.

I've committed proposal of names and structures, taking into account that PR https://github.com/3scale/3scale-operator/pull/383 is not going to happen or at least is not going to happen soon

pkg/apis/apps/v1alpha1/apimanager_types.go
```diff
@@ -172,13 +172,20 @@ type BackendSpec struct {
...
+	RedisPersistentVolumeClaimSpec *BackendRedisPersistentVolumeClaimSpec `json:"redisPersistentVolumeClaim,omitempty"`
+	// +optional
 	ListenerSpec *BackendListenerSpec `json:"listenerSpec,omitempty"`
 	// +optional
 	WorkerSpec *BackendWorkerSpec `json:"workerSpec,omitempty"`
 	// +optional
 	CronSpec *BackendCronSpec `json:"cronSpec,omitempty"`
 }
 
+type BackendRedisPersistentVolumeClaimSpec struct {
```
eguzki, line 184: why not reusable `PersistentVolumeClaimSpec`?

pkg/apis/apps/v1alpha1/apimanager_types.go
```diff
@@ -172,13 +172,20 @@ type BackendSpec struct {
 	// +optional
 	RedisImage *string `json:"redisImage,omitempty"`
 	// +optional
+	RedisPersistentVolumeClaimSpec *BackendRedisPersistentVolumeClaimSpec `json:"redisPersistentVolumeClaim,omitempty"`
```
eguzki, line 175: what about `json:"persistentVolumeClaim,omitempty"` ??

eguzki:
Left some comments

pkg/apis/apps/v1alpha1/apimanager_types.go
miguelsorianod: I don't want to give it a "generic" name being in a broad section. If it were to be inside a "redis" section I'd do it but I didn't want to being there. It's the same case as with 'redisImage'. I didn't choose 'image' for 'redisImage' as it was in a shared broad section

pkg/apis/apps/v1alpha1/apimanager_types.go
miguelsorianod: See comment below

miguelsorianod:
Implementation should be done.

I've added more unit tests and documentation.

Just to clarify, this is an operator-only functionality.

The PR at the moment has been opened against 'master' branch.

Ready for review @eguzki

PR #391, opened 2020-06-18
Increase pvc size

eguzki:
Increase PVC resource request:
* system shared storage: 10Gi
* Redis system: 10Gi
* Redis backend: 10Gi
* System mysql: 10Gi
* System postgresql: 10Gi

### DIFF
--- a/pkg/3scale/amp/operator/system_reconciler.go
+++ b/pkg/3scale/amp/operator/system_reconciler.go
@@ -28,7 +28,7 @@ func NewSystemReconciler(baseAPIManagerLogicReconciler *BaseAPIManagerLogicRecon
 }
 
 func (r *SystemReconciler) reconcileFileStorage(system *component.System) error {
-	if r.apiManager.IsSystemPostgreSQLEnabled() {
+	if r.apiManager.IsSystemS3Enabled() {
 		return r.validateS3StorageProvidedConfiguration()
 	}
 
--- a/pkg/3scale/amp/operator/system_reconciler.go
+++ b/pkg/3scale/amp/operator/system_reconciler.go
@@ -28,14 +28,16 @@ func NewSystemReconciler(baseAPIManagerLogicReconciler *BaseAPIManagerLogicRecon
 }
 
 func (r *SystemReconciler) reconcileFileStorage(system *component.System) error {
+	if r.apiManager.IsSystemPostgreSQLEnabled() {
+		return r.validateS3StorageProvidedConfiguration()
+	}
+
 	if r.apiManager.Spec.System.FileStorageSpec != nil {
-		if r.apiManager.Spec.System.FileStorageSpec.S3 != nil {
-			return r.validateS3StorageProvidedConfiguration()
-		}
 		if r.apiManager.Spec.System.FileStorageSpec.DeprecatedS3 != nil {
 			r.Logger().Info("Warning: deprecated amazonSimpleStorageService field in CR being used. Ignoring it... Please use simpleStorageService")
 		}
 	}
+
 	// System RWX PVC, i.e. shared storage
 	return r.ReconcilePersistentVolumeClaim(system.SharedStorage(), reconcilers.CreateOnlyMutator)
 }
--- a/pkg/3scale/amp/operator/upgrade.go
+++ b/pkg/3scale/amp/operator/upgrade.go
@@ -16,6 +16,7 @@ import (
 	appsv1 "github.com/openshift/api/apps/v1"
 	imagev1 "github.com/openshift/api/image/v1"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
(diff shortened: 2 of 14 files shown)


eguzki:
Warning for the upgrading path: 

`only dynamically provisioned pvc can be resized and the storageclass that provisions the pvc must support resize`

Otherwise upgrade path might fail. Should the operator provide upgrade path? Can be checked if pvc has been provisioned dinamically AND the storageclass suppports resize?

pkg/3scale/amp/operator/system_reconciler.go
```diff
@@ -28,14 +28,16 @@ func NewSystemReconciler(baseAPIManagerLogicReconciler *BaseAPIManagerLogicRecon
 }
 
 func (r *SystemReconciler) reconcileFileStorage(system *component.System) error {
+	if r.apiManager.IsSystemPostgreSQLEnabled() {
```
miguelsorianod, line 31: I think this is wrong. Here we should be checking whether S3 is being used or not. PostgreSQL check (system-database) doesn't seem to be correct here

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -727,6 +736,72 @@ func (u *UpgradeApiManager) ensurePodTemplateLabels(desired *appsv1.DeploymentCo
...
+func (u *UpgradeApiManager) upgradePVCSizes() (reconcile.Result, error) {
+	if !u.apiManager.IsExternalDatabaseEnabled() {
+		if u.apiManager.IsSystemPostgreSQLEnabled() {
+			err := u.updatePVCSize(component.SystemPostgreSQLPVCName, resource.MustParse("10Gi"))
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+		}
+
+		if u.apiManager.IsSystemMysqlEnabled() {
```
miguelsorianod, line 748: I think this would not be totally correct.

IsSystemMysqlEnabled is:

```
func (apimanager *APIManager) IsSystemMysqlEnabled() bool {
	return !apimanager.IsExternalDatabaseEnabled() &&
		apimanager.Spec.System.DatabaseSpec != nil &&
		apimanager.Spec.System.DatabaseSpec.MySQL != nil
}
```

This checks whether System MySQL is explicitely enabled. There's another case where MySQL is enabled, which is the default case (where system mysql usage is implicit). In the default case no spec.system.databasespec.mysql is set

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -727,6 +736,72 @@ func (u *UpgradeApiManager) ensurePodTemplateLabels(desired *appsv1.DeploymentCo
...
+	err := u.Client().Get(context.TODO(), namespacedName, existing)
+	// NotFound also regarded as error, as pvc is expected to exist
+	if err != nil {
+		return err
+	}
+
+	// v1.ResourceStorage must exist, otherwise PVC is regarded as invalid
+	existingSize := existing.Spec.Resources.Requests[v1.ResourceStorage]
+
+	if existingSize.Cmp(desiredSize) != 0 {
```
miguelsorianod, line 793: I seem to remember there were some comparison methods that checked unexported fields, and that made detecting differences due to different internal representations of the same value depending on the serialized value. Have we checked this is working as expected in all cases?

That's why helper.CmpResources was created in its day for other kind of resources. I don't know whether CmpResources takes into account ResourceStorage though

miguelsorianod:
> Warning for the upgrading path:
> 
> only dynamically provisioned pvc can be resized and the storageclass that provisions the pvc must support resize
> 
> Otherwise upgrade path might fail. Should the operator provide upgrade path? Can be checked if pvc has been provisioned dinamically AND the storageclass suppports resize?

That's a concerning issue yes...

I'm also concerned about increasing x10 storage requirements of a 3scale installation.

Let's talk about this a little bit through other channels.

pkg/3scale/amp/operator/system_reconciler.go
eguzki: good catch

pkg/3scale/amp/operator/upgrade.go
eguzki: Right.

I fixed updating the `IsSystemMysqlEnabled` method to 
```
func (apimanager *APIManager) IsSystemMysqlEnabled() bool {                
    // Mysql is the default one                                            
    // If it happens to be more DB's, they should also be included         
    return !apimanager.IsExternalDatabaseEnabled() &&                      
        !apimanager.IsSystemPostgreSQLEnabled()                            
}                                                                          
```

pkg/3scale/amp/operator/upgrade.go
eguzki: CmpResources does not take into account ResourceStorage. Already checked it. 

The upgrade logic uses `Quantity` type comparator, which I expect to be working

```
func main() {
	desired := resource.MustParse("10Gi")
	existing := resource.MustParse("1Gi")
	fmt.Println("desired == existing", existing.Cmp(desired) == 0)
	existing = resource.MustParse("10Gi")
	fmt.Println("desired == existing", existing.Cmp(desired) == 0)
}
```

```
desired == existing false
desired == existing true
```

eguzki:
Closing this issue as:
* the upgrading path may fail and there is no way to prevent it
* Initial required storage size can be up to 40Gb, which can be too much

A better way to go is: 
* allow "install only" configuration of storage size
* allow linking PV provisioned by the customer by name (thus no storage size is needed)

PR #394, opened 2020-06-19
Add optional comment tag to system's appspec and sidekiqspec

miguelsorianod:
It does not change existing behaviour due to omitempty automatically
implies optional but it's still recommended the optional comment
tag to be added


miguelsorianod:
No CRD does not change due to omitempty automatically implies +optional

PR #397, opened 2020-06-22
Doc backendapi crds

eguzki:
Documentation of Bakend CRD's.


deploy/crds/capabilities.3scale.net_products_crd.yaml
```diff
@@ -80,7 +80,6 @@ spec:
                           type: string
                         value:
                           description: Limit Value
-                          format: int64
```
miguelsorianod, line 83: I guess this change is because of the incompatibility with OpenAPI validation. An alternative to allow big numbers would be using a string and parsing the value (it was done this way with another similar attribute with other Capabilities kind).
Are the limits potentially big? If so we should consider it

deploy/olm-catalog/3scale-operator-master/0.0.0/3scale-operator-master.v0.0.0.clusterserviceversion.yaml
```diff
@@ -103,6 +103,48 @@ metadata:
...
           }
+        },
+        {
+          "apiVersion": "capabilities.3scale.net/v1beta1",
+          "kind": "Account",
+          "metadata": {
+            "name": "example-account"
+          },
+          "spec": {
+            "size": 3
```
miguelsorianod, line 114: I guess at this point in this PR this is expected because we are waiting to have all changes applied before regenerating the CSV. Just let's not forget about updating the CRDs files and this correctly, as right now it's a generic CRD value (spec.size)

deploy/olm-catalog/3scale-operator-master/0.0.0/3scale-operator-master.v0.0.0.clusterserviceversion.yaml
```diff
@@ -267,7 +326,7 @@ spec:
                   value: centos/postgresql-10-centos7
                 - name: ZYNC_POSTGRESQL_IMAGE
                   value: centos/postgresql-10-centos7
-                image: quay.io/3scale/3scale-operator:master
```
miguelsorianod, line 270: This is performing an unwanted change :( we should revert this before merging

deploy/crds/capabilities.3scale.net_products_crd.yaml
eguzki: This is  because 3scale go client objects define value as `int`. https://github.com/3scale/3scale-porta-go-client/blob/86274104320b597755af595f27be8771c944a579/client/types.go#L576

We could change in 3scale porta go client and in the CRD, but in 64bit systems, int range is  [-2^63, 2^63-1]. Really big.

There is no need to do so.

eguzki:
ready for a new review @miguelsorianod

PR #403, opened 2020-06-25
Product and Bakend CRD's Mapping Rule object increment field required

eguzki:
3scale API does not allow creating mapping rules without `increment` value. It does not set *default* value. Thus, 3scale operator should not set default value and `increment` value  will be required in the CR.


eguzki:

PR #405, opened 2020-06-26
Monitoring upgrade

eguzki:
Upgrade path for monitoring resources.

Added required changes to support monitoring only with operator.

### DIFF
--- a/pkg/3scale/amp/component/backend.go
+++ b/pkg/3scale/amp/component/backend.go
@@ -405,7 +405,7 @@ func (backend *Backend) buildBackendWorkerEnv() []v1.EnvVar {
 		helper.EnvVarFromSecret("CONFIG_EVENTS_HOOK_SHARED_SECRET", "system-events-hook", "PASSWORD"),
 	)
 
-	if backend.Options.WorkerMetrics != nil {
+	if backend.Options.WorkerMetrics {
 		result = append(result,
 			v1.EnvVar{Name: "CONFIG_WORKER_PROMETHEUS_METRICS_PORT", Value: BackendWorkerMetricsPortStr},
 			v1.EnvVar{Name: "CONFIG_WORKER_PROMETHEUS_METRICS_ENABLED", Value: "true"},
@@ -430,7 +430,7 @@ func (backend *Backend) buildBackendListenerEnv() []v1.EnvVar {
 		helper.EnvVarFromSecret("CONFIG_INTERNAL_API_PASSWORD", BackendSecretInternalApiSecretName, BackendSecretInternalApiPasswordFieldName),
 	)
 
-	if backend.Options.ListenerMetrics != nil {
+	if backend.Options.ListenerMetrics {
 		result = append(result,
 			v1.EnvVar{Name: "CONFIG_LISTENER_PROMETHEUS_METRICS_PORT", Value: BackendListenerMetricsPortStr},
 			v1.EnvVar{Name: "CONFIG_LISTENER_PROMETHEUS_METRICS_ENABLED", Value: "true"},
--- a/pkg/3scale/amp/component/backend.go
+++ b/pkg/3scale/amp/component/backend.go
@@ -398,9 +398,15 @@ func (backend *Backend) buildBackendWorkerEnv() []v1.EnvVar {
 	result = append(result,
 		helper.EnvVarFromSecret("CONFIG_EVENTS_HOOK", "system-events-hook", "URL"),
 		helper.EnvVarFromSecret("CONFIG_EVENTS_HOOK_SHARED_SECRET", "system-events-hook", "PASSWORD"),
-		v1.EnvVar{Name: "CONFIG_WORKER_PROMETHEUS_METRICS_PORT", Value: BackendWorkerMetricsPortStr},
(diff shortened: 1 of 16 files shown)


pkg/3scale/amp/component/backend.go
```diff
@@ -14,6 +14,11 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+const (
+	BackendListenerName = "backend-listener"
+	BackendWorkerName   = "backend-worker"
```
miguelsorianod, line 19: I'm curious why has this not been done for backend-cron?

pkg/3scale/amp/component/backend.go
```diff
@@ -398,9 +403,15 @@ func (backend *Backend) buildBackendWorkerEnv() []v1.EnvVar {
 	result = append(result,
 		helper.EnvVarFromSecret("CONFIG_EVENTS_HOOK", "system-events-hook", "URL"),
 		helper.EnvVarFromSecret("CONFIG_EVENTS_HOOK_SHARED_SECRET", "system-events-hook", "PASSWORD"),
-		v1.EnvVar{Name: "CONFIG_WORKER_PROMETHEUS_METRICS_PORT", Value: BackendWorkerMetricsPortStr},
-		v1.EnvVar{Name: "CONFIG_WORKER_PROMETHEUS_METRICS_ENABLED", Value: "true"},
 	)
+
+	if backend.Options.WorkerMetrics != nil {
```
miguelsorianod, line 408: Shouldn't we be checkind that is not null AND that is set to true? :thinking:

pkg/3scale/amp/operator/apicast_options_provider.go
```diff
@@ -35,7 +35,7 @@ func (a *ApicastOptionsProvider) GetApicastOptions() (*component.ApicastOptions,
 	a.apicastOptions.ImageTag = product.ThreescaleRelease
 	a.apicastOptions.OpenSSLVerify = strconv.FormatBool(*a.apimanager.Spec.Apicast.OpenSSLVerify)
 	a.apicastOptions.ResponseCodes = strconv.FormatBool(*a.apimanager.Spec.Apicast.IncludeResponseCodes)
-	boolTmp := a.apimanager.IsMonitoringEnabled()
+	boolTmp := true
```
miguelsorianod, line 38: Does this change have any impact? Why has this been changed to always be true and not dependant on the monitoring cr flag?

pkg/3scale/amp/component/backend.go
eguzki: Backend cron DC name is not used in upgrade path.

Should be done? I think so. Required, not at all.

pkg/3scale/amp/operator/backend_options_provider.go
```diff
@@ -58,6 +58,10 @@ func (o *OperatorBackendOptionsProvider) GetBackendOptions() (*component.Backend
 	o.backendOptions.ListenerMonitoringLabels = o.listenerMonitoringLabels()
 	o.backendOptions.WorkerMonitoringLabels = o.workerMonitoringLabels()
 
+	boolTmp := true
```
miguelsorianod, line 61: Why are we enabling this unconditionally? shouldn't be based on the monitoring cr flag?

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -36,6 +36,9 @@ func NewUpgradeApiManager(b *reconcilers.BaseReconciler, apiManager *appsv1alpha
 }
 
 func (u *UpgradeApiManager) Upgrade() (reconcile.Result, error) {
+	// TODO Upgrade should be grouped by component and not by feature.
```
miguelsorianod, line 39: I think this totally depends on what you are doing.
Sometimes it seems it's desirable to upgrade by component and sometimes it seems it is desirable to be done by feature, depending on whether you want to do a specific set of changes before another etc...

pkg/3scale/amp/component/backend.go
eguzki: That field type is `*bool`. But it is not intended to hold true or false. The real information is on pointer existence. As it is not reconciliable. It is enabled for operator based deployment and disabled for template based deployment.

Changing it to `bool`. It is confusing.

pkg/3scale/amp/operator/apicast_options_provider.go
eguzki: For simplicity, it will be always enabled for operator deployments and disabled for template deployments.

pkg/3scale/amp/operator/backend_options_provider.go
eguzki: see comments above

pkg/3scale/amp/component/backend.go
miguelsorianod: I agree. I was just curios about why was not done :+1:

pkg/3scale/amp/component/backend.go
miguelsorianod: :+1:

eguzki:
Metrics exposed by components are always enabled. 

Monitoring resources (services, servicemonitors, grafana dashboards, prometheus rules) will only be deployed when that flag is enabled. And when the flag is back to false, all resources will be deleted.

PR #414, opened 2020-07-07
Simplify monitoring 

eguzki:
* monitoring-key=middleware only in grafanadashboards
* replace servicemonitors with podmonitors
  * PodMonitors require `metrics` port to be exposed in the DC or in the Service (OCP 4.2)

Added upgrading path

### DIFF
--- a/pkg/3scale/amp/component/backend.go
+++ b/pkg/3scale/amp/component/backend.go
@@ -126,6 +126,9 @@ func (backend *Backend) WorkerDeploymentConfig() *appsv1.DeploymentConfig {
 							Env:             backend.buildBackendWorkerEnv(),
 							Resources:       backend.Options.WorkerResourceRequirements,
 							ImagePullPolicy: v1.PullIfNotPresent,
+							Ports: []v1.ContainerPort{
+								v1.ContainerPort{Name: "metrics", ContainerPort: BackendWorkerMetricsPort, Protocol: v1.ProtocolTCP},
+							},
 						},
 					},
 					ServiceAccountName: "amp"}},
@@ -264,6 +267,7 @@ func (backend *Backend) ListenerDeploymentConfig() *appsv1.DeploymentConfig {
 								v1.ContainerPort{HostPort: 0,
 									ContainerPort: 3000,
 									Protocol:      v1.ProtocolTCP},
+								v1.ContainerPort{Name: "metrics", ContainerPort: BackendListenerMetricsPort, Protocol: v1.ProtocolTCP},
 							},
 							Env:       backend.buildBackendListenerEnv(),
 							Resources: backend.Options.ListenerResourceRequirements,
--- a/pkg/3scale/amp/component/backend.go
+++ b/pkg/3scale/amp/component/backend.go
@@ -126,9 +126,7 @@ func (backend *Backend) WorkerDeploymentConfig() *appsv1.DeploymentConfig {
 							Env:             backend.buildBackendWorkerEnv(),
 							Resources:       backend.Options.WorkerResourceRequirements,
 							ImagePullPolicy: v1.PullIfNotPresent,
-							Ports: []v1.ContainerPort{
-								v1.ContainerPort{Name: "metrics", ContainerPort: BackendWorkerMetricsPort, Protocol: v1.ProtocolTCP},
-							},
+							Ports:           backend.workerPorts(),
 						},
 					},
 					ServiceAccountName: "amp"}},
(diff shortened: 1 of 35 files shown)


eguzki:
cc @slopezz

pkg/3scale/amp/component/system.go
```diff
@@ -794,13 +802,14 @@ func (system *System) SidekiqDeploymentConfig() *appsv1.DeploymentConfig {
...
 						v1.Container{
-							Name:            "system-sidekiq",
+							Name:            SystemSidekiqName,
 							Image:           "amp-system:latest",
 							Args:            []string{"rake", "sidekiq:worker", "RAILS_MAX_THREADS=25"},
 							Env:             system.buildSystemBaseEnv(),
 							Resources:       *system.Options.SidekiqContainerResourceRequirements,
 							VolumeMounts:    system.sidekiqContainerVolumeMounts(),
 							ImagePullPolicy: v1.PullIfNotPresent,
+							Ports:           system.sideKiqPorts(),
```
miguelsorianod, line 812: Does Ports = nil has the same effect than Ports = v1.Ports{} ?
We are applying that change in this case when Sidekiq metrics aren't enabled compared to sidekiq deployments in previous 3scale-operator versions.
Maybe define ports in sideKiqPorts() method as nil by default so we don't change it?

pkg/3scale/amp/component/backend.go
```diff
@@ -538,3 +535,25 @@ func (backend *Backend) ListenerPodDisruptionBudget() *v1beta1.PodDisruptionBudg
...
+
+	if backend.Options.ListenerMetrics {
+		ports = append(ports, v1.ContainerPort{Name: "metrics", ContainerPort: BackendListenerMetricsPort, Protocol: v1.ProtocolTCP})
+	}
+
+	return ports
+}
+
+func (backend *Backend) workerPorts() []v1.ContainerPort {
+	ports := []v1.ContainerPort{}
```
miguelsorianod, line 552: Same comment as https://github.com/3scale/3scale-operator/pull/414#discussion_r451721168

pkg/3scale/amp/component/system.go
eguzki: For 3scale operator based deployments, ports will be set. Even if monitoring is not enabled. Same for apicast, backend. That is why upgrade path has been implemented also. If upgrade was not implemented, one 3scale after being upgraded and monitoring enabled, ports will not be set and monitoring would fail.

For template based deployments, there is no change. Basically, because it is optional. So, template generation remains the same.

`Ports []ContainerPort `json:"ports,omitempty"`

eguzki:
comments answered

pkg/3scale/amp/component/system.go
miguelsorianod: > For 3scale operator based deployments, ports will be set. Even if monitoring is not enabled

That doesn't happen with sidekiq and backend's worker. When the components' option does not have the metrics enabled in those cases there are no additional ports at all defined, and assuming that operator/templates will appropriately or serialize/deserialize in a way a component option to a value or another seems does not seem a correct approach to me as design of the component should be independent of that information. It's just a component with options from its point of view.

In my opinion if we were to be 100% strictly correct we should not set the ports empty if there are no ports defined, as it is a change that we're doing compared to what we were doing before (leave nil as default vs leaving empty now) and component design should be independent of operator/templates from that point of view. 

In any case, if the behaviour doesn't change we don't need to be 100% strictly correct so good enough for me :+1: . Just leaving my thoughts on this.

eguzki: > That doesn't happen with sidekiq and backend's worker.

Yes it happens. Sidekiq and backend worker ports will be set even if monitoring is disabled. Check code.

eguzki: Changing to return nil if there are no ports defined (template case only)

eguzki:
Added new commit based on your comments. @miguelsorianod

pkg/3scale/amp/component/system.go
miguelsorianod: > Yes it happens. Sidekiq and backend worker ports will be set even if monitoring is disabled. Check code.

No. When metrics are not enabled in those cases no ports are set, there would be no ports elements at all, just an empty ports variable which was the previous behavior in this PR, or nil being the new behavior that you've changed and the way I was pointing I thought/think it was the more accurate solution as it's the same as it was being done before this PR. That was the difference between sidekiq and backend worker compared to backend listener where at least 1 port element is set.

Maybe the confusion has been language and we were talking about different things. To me with "no ports are set" I meant no elements are set on the ports variable.

eguzki: Yes. Check code.

When metrics are not enabled in those cases ports ARE set.

eguzki: to something not empty and not nil

miguelsorianod: My understanding of the following method is that when metrics are not enabled ports are set to nil (and before with the previous implementation in the PR they were set to empty, which It was what I was pointing I'd rather be nil instead):

https://github.com/3scale/3scale-operator/blob/968b874d3dede547cc1692b9c1201a863db2e5fb/pkg/3scale/amp/component/backend.go#L551-L559

https://github.com/3scale/3scale-operator/blob/968b874d3dede547cc1692b9c1201a863db2e5fb/pkg/3scale/amp/component/backend.go#L129

Same applies to sidekiq.

I'm not taking into account upgrades, operator-specific scenario or templates-specific scenario, as I'm evaluating it from a perspective of the component itself with the options as I mentioned before.

PR #416, opened 2020-07-07
Add oracle's specific custom configuration

miguelsorianod:
In order to allow external Oracle database an additional environment variable is needed to provide Oracle's SYSTEM's (oracle's administrative user) password.


miguelsorianod:
Looking at the oracle documentation in the templates repository I've seen the `ORACLE_SYSTEM_PASSWORD` is defined in pre-hook and post-hook pods of the system-app DeploymentConfig (https://github.com/3scale/3scale-amp-openshift-templates/tree/master/amp/system-oracle)

The current implementation of this PR adds the `ORACLE_SYSTEM_PASSWORD` environment variable as an optional environment variable of those pods that references the system-database secret.
The idea is that the user would pre-create the system-database secret with that environment variable (and also the `DATABASE_URL` one as always) Notice how that optional environment variable secret reference is added on ALL templates contents (standard, s3, postgresql, ha...) too with the current implementation.
From the operator point of view we could check whether that environment variable is set or not and conditionally add that it or not to the prehook/posthook pods and in the templates never do it (with an option configuration parameter in system's component probably) but I'm not sure whether it's worth it / desirable so for the moment I haven't done that. @eguzki let me know your thoughts about this approach.

I'm a little bit surprised about `ORACLE_SYSTEM_PASSWORD` being sent to system-app post-hook pod, as there are no other environment variable sent there, not even the `DATABASE_URL` environment variable. Is that environment variable really needed in the post-hook pod  @hallelujah?

hallelujah:
@miguelsorianod  the `ORACLE_SYSTEM_PASSWORD` is needed on the pre-hook and on the init containers when checking if database is ready. The post-hook does not check anything thus it is not needed there

miguelsorianod:
Thank you hery! 

I've updated the PR to only include `ORACLE_SYSTEM_PASSWORD` in the pre-hook pod

eguzki:
It only make sense for the operator, so I would create the env var conditionally only for the operator. I guess, upgrade producedure is not required either (not for the operator and not for the templates).

miguelsorianod:
> It only make sense for the operator, so I would create the env var conditionally only for the operator. I guess, upgrade producedure is not required either (not for the operator and not for the templates).

I've added an option to System component so we can enable oracle-related optional settings only on operator side. Now templates content does not change.
upgrade procedure should not be required, as it's an optional environment variable and the user would be the one setting the value in the secret only in case is desired, in the initial deployment stage (not for existing installations)

PR #420, opened 2020-07-14
fixed THREESCALE-5631

leathersole:
replaced version 0.4.1 with 0.4.2


eguzki:
That is correct, but does not have any effect. Manifest files that are published are located in another repository.

http://pkgs.devel.redhat.com/cgit/containers/3scale-operator-prod-operator-metadata/

PR #425, opened 2020-07-14
capabilities: return error when provider account not found

eguzki:


eguzki:
not related to the changes of this PR. 

Fixed here: https://github.com/3scale/3scale-operator/pull/426

eguzki:
rebased, tests should be working now

PR #429, opened 2020-07-15
[Backport 427] leave blank com.redhat.component-version metering label when too long

eguzki:


thomasmaas:
So I would just always leave it blank

PR #432, opened 2020-07-17
Capabilities provider account status

eguzki:
Add provider account host url in status for `Backend` and `Product` CRD's.

As example, check `providerAccountHost` field in status object
```
kind: Backend
metadata:
  name: backend-a
  namespace: eguz
spec:
  ...
status:
  backendId: 61535
  providerAccountHost: https://3scale-supertest-admin.3scale.net
```

### DIFF
--- a/pkg/controller/backend/backend_controller.go
+++ b/pkg/controller/backend/backend_controller.go
@@ -201,37 +201,6 @@ func (r *ReconcileBackend) reconcile(backendResource *capabilitiesv1beta1.Backen
 	return reconcile.Result{}, nil
 }
 
-func (r *ReconcileBackend) reconcileSpec(backendResource *capabilitiesv1beta1.Backend) (*helper.BackendAPIEntity, error) {
-	err := r.validateSpec(backendResource)
-	if err != nil {
-		// Do not wrap error
-		return nil, err
-	}
-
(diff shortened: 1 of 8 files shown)


eguzki:
ping @miguelsorianod

pkg/controller/backend/backend_controller.go
```diff
@@ -150,13 +150,41 @@ func (r *ReconcileBackend) reconcile(backendResource *capabilitiesv1beta1.Backen
...
+			// On Validation error, no need to retry as spec is not valid and needs to be changed
+			logger.Info("ERROR", "spec validation error", err)
+			r.EventRecorder().Eventf(backendResource, corev1.EventTypeWarning, "Invalid Backend Spec", "%v", err)
+			return reconcile.Result{}, nil
+		}
+
+		return reconcile.Result{}, err
+	}
+
+	providerAccount, err := helper.LookupProviderAccount(r.Client(), backendResource.Namespace, backendResource.Spec.ProviderAccountRef, r.Logger())
```
miguelsorianod, line 165: Code from reconcileSpec has been copied here and slightly modified.
I think we should remove reconcileSpec or update it to the new content and call it from here. Same would apply in product controller

eguzki: removed unused method

PR #437, opened 2020-07-22
Update APIManagerBackup PVC requirements documentation

miguelsorianod:
This PR updates APIManagerBackup documentation to clarify PV pre-binding usage.
    
When providing a pre-existing PersistentVolume to APIManagerBackup
resources still have to be defined as per K8s PVCs behavior.
This commit updates the APIManagerBackup reference documentation
and deployment examples to specify and clarify this behavior which
might be surprising but that is how PVCs specification in K8s
requires and behaves.

Details about this behavior in: https://issues.redhat.com/browse/THREESCALE-5677


eguzki:
Looking good

PR #443, opened 2020-07-30
Capabilities: default provider account from local namespace 3scale deployment

eguzki:
If provider account is not referenced directly in the CR *and* if there is no default secret with provider_account credentials, then this PR implements the last option before reporting error, look up for the default provider account (3scale-admin) in the current namespace 3scale deployment.

Some refactor was included to create `helper` package in the `controllers` package to avoid `pkg.helper` dependency on `components` package. Moved the account provider look up method to the new package


pkg/controller/helper/lookup_provider_account.go
```diff
@@ -0,0 +1,131 @@
...
+	err := cl.List(context.TODO(), apimanagerList, listOps...)
+	if err != nil {
+		return nil, fmt.Errorf("providerAccountFromLocal3scaleSource: %w", err)
+	}
+
+	if len(apimanagerList.Items) == 0 {
+		return nil, nil
+	}
+
+	apimanager := apimanagerList.Items[0]
```
miguelsorianod, line 114: Would it be worth it to check that there are no more than 1 APIManager? It shouldn't but I wonder whether checking that would be appropriate.

miguelsorianod (approved):
lgtm :+1:

pkg/controller/helper/lookup_provider_account.go
eguzki: enhancement!

If there are more than 1,  raise an error?

PR #444, opened 2020-07-31
capabilities: reconciliate product on backend metric deletion

eguzki:
When backend metric or method is deleted, Application plan limits and pricing rules with references to it have to be deleted also, as it happens automatically in the System UI.

Fixes [issue](https://issues.redhat.com/browse/THREESCALE-5534) about product and backend consistency when there are references between them.

### DIFF
--- a/pkg/controller/helper/product_list.go
+++ b/pkg/controller/helper/product_list.go
@@ -1,0 +1,50 @@
+package helper
+
+import (
+	"context"
+	"fmt"
+
+	capabilitiesv1beta1 "github.com/3scale/3scale-operator/pkg/apis/capabilities/v1beta1"
+
+	"github.com/go-logr/logr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	controllerclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ProductList returns a list of product custom resources where all elements:
+// - Sync state (ensure remote product exist and in sync)
+// - Same 3scale provider Account
+func ProductList(ns string, cl client.Client, providerAccount *ProviderAccount, logger logr.Logger) ([]capabilitiesv1beta1.Product, error) {
+	productList := &capabilitiesv1beta1.ProductList{}
+	opts := []controllerclient.ListOption{
+		controllerclient.InNamespace(ns),
+	}
+	err := cl.List(context.TODO(), productList, opts...)
+	logger.V(1).Info("Get list of Product resources.", "Err", err)
+	if err != nil {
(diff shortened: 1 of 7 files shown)


eguzki:
ping @miguelsorianod

miguelsorianod:
I was waiting for your input here.

After we talked about this you were going to review this approach (let it fail vs try to fix it, etc...). Have you reviewed it?

eguzki:
You are the one in the asignees section :) This is ready to be reviewed.

miguelsorianod:
? That does not change what was talked  and the action points of the other day. I was expecting you to review it and give me feedback based on that.
If it helps you I've re-assigned it to you. Let me know your views on that and why one decision has been taken in front of the other

eguzki:
This PR implements a "Product CR Spec" update from the Backend Controller to make it valid from a invalid state. This adds robustness to the operator, kind of healing capability when some spec becomes invalid due to an external action. This same behavior has been seen by 3scale internal System. When backend metric or method is deleted in the dashboard, the application plan limits and pricing rules with references to it are automatically deleted also.

The other approach is, do not react and do not "heal" invalid CR. If backend metric is deleted, the product spec will be invalid, as existing backend reference no longer exists. The operator would report as events, CR status and internal log and manual action would be required. 

Both approaches can be valid. Since 3scale UI "heals" application plans limits and pricing rules automatically, for consistency shake, I would add this feature to the operator.

pkg/controller/helper/product_list.go
```diff
@@ -0,0 +1,50 @@
...
+	validProducts := make([]capabilitiesv1beta1.Product, 0)
+	for idx := range productList.Items {
+		// Filter by synchronized
+		if !productList.Items[idx].IsSynced() {
+			continue
+		}
+
+		productProviderAccount, err := LookupProviderAccount(cl, ns, productList.Items[idx].Spec.ProviderAccountRef, logger)
+		if err != nil {
+			return nil, fmt.Errorf("BackendList: %w", err)
```
miguelsorianod, line 38: I think this log should contain 'ProductList' right?

miguelsorianod (changes requested):
Thanks for the detailed information and insight @eguzki :+1: 

I'm fine with this approach. Both approaches seem acceptable solutions to me with some pros and cons for each one.

I left a minor comment. All other changes look good to me :+1:

pkg/controller/helper/product_list.go
eguzki: :+1:

miguelsorianod (approved):
:+1:

PR #446, opened 2020-08-05
Fix regex validation in Product type for fields that specify USD

miguelsorianod:
The original regular expression, that I understand was intended to validate a floating point number with optionally two decimal numbers exactly was not validating that specification.

This PR fixes the regular expression to validate it.

This applies to pricePerUnit, setupFee and costMonth fields

### DIFF
--- a/deploy/crds/capabilities.3scale.net_products_crd.yaml
+++ b/deploy/crds/capabilities.3scale.net_products_crd.yaml
@@ -42,7 +42,7 @@ spec:
                     type: boolean
                   costMonth:
                     description: Cost per Month (USD)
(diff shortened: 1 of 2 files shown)


miguelsorianod:
With the new regex the following values specified in costMonth (and also for the other fixed fields) have the following validation results:

```
"25.34" -> validation pass
"2" -> validation pass
"0.43" -> validation pass
"2.3" -> validation failure
"28.342" -> validation failure
"25a34" -> validation failure
"ab.cd" -> validation failure
"2.cd" -> validation failure
"ab.34 -> validation failure
```

eguzki (approved):

PR #449, opened 2020-08-11
Create and Reconcile the 'position' attribute in Product and Backend MappingRules

miguelsorianod:
This PRs makes the 'position' attribute configurable in Product and Backend MappingRules.

The 'position' attribute allows a user to define the processing priority order for a set of MappingRules.

This implementation approach has added this attribute as a field in the CR for the MappingRules.

The implementation has some advantages and drawbacks:
* Advantage: The user can define the desired position number arbitrarily. Position numbers between mappingrules do not necessarily have to be confitguous. For example, you could have 3 mappingRules with the positions 3,6,9 defined respectively
* Disadvantage: When a MappingRule is added or removed all the existing MappingRule positions in System are reordered. For example, if there are 3 MappingRules with positions 1,2,7 respectively and a new MappingRule is added with position set to 6, the original MappingRule with position 7 is changed by system to position 8. This causes an inconsistency between what's defined in the CR and what's effective in System, and when reconciliation happens in that case it would always detect differences. The implication of this is that if we take this approach we are moving the responsibility of reordering the existing MappingRules position to the user when a MappingRule is added or removed.

An alternative implementation is  provided in the pull request #452 to evaluate which one we prefer/is more desirable

### DIFF
--- a/pkg/controller/backend/3scale_reconciler.go
+++ b/pkg/controller/backend/3scale_reconciler.go
@@ -475,6 +475,16 @@ func (t *ThreescaleReconciler) reconcileMatchedMappingRules(matchedList []mappin
 			params["delta"] = strconv.Itoa(data.spec.Increment)
 		}
 
+		//
+		// Reconcile position. Only reconcile position when it has been set
+		// in the CR spec. If it has not been set on the spec we assume the desired
+		// position is the one set by the 3scale API and we do not modify
+		// the spec with the value from the API either.
+		//
+		if data.spec.Position != nil && *data.spec.Position != data.item.Position {
+			params["position"] = strconv.FormatInt(int64(*data.spec.Position), 10)
+		}
+
 		if len(params) > 0 {
 			err := t.backendAPIEntity.UpdateMappingRule(data.item.ID, params)
 			if err != nil {
@@ -503,6 +513,11 @@ func (t *ThreescaleReconciler) createNewMappingRules(desiredList []capabilitiesv
 			"http_method": spec.HTTPMethod,
 			"metric_id":   strconv.FormatInt(metricID, 10),
 			"delta":       strconv.Itoa(spec.Increment),
+		}
+
+		// Only set params position when position is set in CR's spec
+		if spec.Position != nil {
+			params["position"] = strconv.FormatInt(int64(*spec.Position), 10)
 		}
 
 		err = t.backendAPIEntity.CreateMappingRule(params)
--- a/pkg/controller/product/mapping_rules.go
+++ b/pkg/controller/product/mapping_rules.go
@@ -131,6 +131,16 @@ func (t *ThreescaleReconciler) reconcileMatchedMappingRules(matchedList []mappin
 			params["delta"] = strconv.Itoa(data.spec.Increment)
 		}
 
+		//
+		// Reconcile position. Only reconcile position when it has been set
+		// in the CR spec. If it has not been set on the spec we assume the desired
+		// position is the one set by the 3scale API and we do not modify
+		// the spec with the value from the API either.
+		//
+		if data.spec.Position != nil && *data.spec.Position != data.item.Position {
+			params["position"] = strconv.FormatInt(int64(*data.spec.Position), 10)
+		}
+
 		if len(params) > 0 {
 			err := t.productEntity.UpdateMappingRule(data.item.ID, params)
 			if err != nil {
@@ -159,6 +169,11 @@ func (t *ThreescaleReconciler) createNewMappingRules(desiredList []capabilitiesv
 			"http_method": spec.HTTPMethod,
 			"metric_id":   strconv.FormatInt(metricID, 10),
 			"delta":       strconv.Itoa(spec.Increment),
+		}
+
+		// Only set params position when position is set in CR's spec
+		if spec.Position != nil {
+			params["position"] = strconv.FormatInt(int64(*spec.Position), 10)
 		}
 
 		err = t.productEntity.CreateMappingRule(params)


eguzki:
It is not clear to me which is the best approach. On the one hand, allowing **position** field to be part of the CRD seems more consistent with the API. The first implication is that the operator code is simpler. On the other hand, though, several reconciliation loops will be run because this unexpected 3scale behavior of updating internally position values when adding/deleting MRs. 

Question: those reconciliation extra runs end up to a steady state or updating MR values leads to endless conflicts between desired state and existing state? I would say only few more runs. 

IMO, if 3scale API ends up in a steady state, I would run for this approach, even if it requires some extra reconciliation loops. It seems more consistent with the 3scale API (and also with System UI having a position field) and more flexible than position value auto-calculation. 

Besides, position seems to be an optional attribute of the MR. With the auto computed approach, we are enforcing position always.

eguzki:
BTW, the behavior you described about internal position update value (which I have reproduced easily in UI) seems to be a bug to me. @hallelujah or @guicassolato (or anybody else from your team), can you confirm this is the expected behavior? (quoted here for you)

```
When a MappingRule is added or removed all the existing MappingRule positions in System are reordered. For example, if there are 3 MappingRules with positions 1,2,7 respectively and a new MappingRule is added with position set to 6, the original MappingRule with position 7 is changed by system to position 8.
```

eguzki:
@miguelsorianod we might add a validation if position is allowed. Position values can not be repeated. Otherwise endless reconciliation loop would be triggered. 

Let's say we have this MR list:
```
mappingRules:                
  - httpMethod: GET          
    pattern: "/a1"         
    increment: 1             
    metricMethodRef: hits    
    position: 1
  - httpMethod: GET          
    pattern: "/a2"         
    increment: 1             
    metricMethodRef: hits
    position: 1    <- repeated position value
```

The operator would create first MR successfully. It would try to create the second MR and the API would not fail. Instead, it would assign position 1 to the MR */a2* and update position value  of MR `/a1` to 2. As this is not the desired value, reconciliation loop would happen again and again.

miguelsorianod:
> Question: those reconciliation extra runs end up to a steady state or updating MR values leads to endless conflicts between desired state and existing state? I would say only few more runs.

I think it leads to a steady state. I've done the following test:
Create three mapping rules with explicit positions 1,5.
Then add a new mapping rule with explicit position 4.
Once that is done I can see in the logs the new metric being created, and then I also can see in the logs that an UpdateMappingRule happens, which I understand happens because the original one with position 5 was updated by System API to position 6 and the operators sees that what's in the CR (5) does not match with what's in System (6) so it updates it to the value in the CR reverting to 5.

I have not tried all sorts of combinations though so there could be some that were more troublesome.

> The operator would create first MR successfully. It would try to create the second MR and the API would not fail. Instead, it would assign position 1 to the MR /a2 and update position value of MR /a1 to 2. As this is not the desired value, reconciliation loop would happen again and again.

Yes that was one of my worries about allowing the position attribute and one of the reasons I implemented the approach in the other PR, as we would stop worrying about all of this kind of issues (temporarily inaccurate states, possible errors etc.). 

> Besides, position seems to be an optional attribute of the MR. With the auto computed approach, we are enforcing position always.

My thought was/is that at the end, even if the position is not explicitly assigned the MappingRules have a position set anyway so why not taking advantage of the fact that we are using an array (where the order matters). It seemed simpler and we don't lose any real functionality.

> IMO, if 3scale API ends up in a steady state, I would run for this approach, even if it requires some extra reconciliation loops. It seems more consistent with the 3scale API (and also with System UI having a position field) and more flexible than position value auto-calculation.

What flexibility are you referring to? I think flexibility in this case does not add any real value. Only thing you would lose in flexibility I can think of is the ability to leave gaps between positions, but that does not affect the functionality, it's just for organization purposes and it can be achieved by simply placing the element in between other existing elements with the other approach. It seems to me a limitation of the UI that we would solve by using the array positions.

> @miguelsorianod we might add a validation if position is allowed. Position values can not be repeated. Otherwise endless reconciliation loop would be triggered.

Checking for duplicates basically would require adding more logic, checking it and reconciling status and marking it appropriately leaving it in an inconsistent state if it has been incorrectly defined, adding complexity which was one of the justifications written here about going to this approach instead of the other one.

What are we really gaining by allowing an explicit Position attribute? I'm having a hard time seeing the value of that apart from "we are translating the API as a CR field" and several edge cases / non-ideal scenarios by doing this.

eguzki:
I think we have done a good analysis and end up in a "dead end". We need anyone from @3scale/system contribute to this and drop their opinion about what is the best approach. Basically, a) have `position` attribute as optional field or b) enforce `position` value based on the location in the array.

hallelujah:
> BTW, the behavior you described about internal position update value (which I have reproduced easily in UI) seems to be a bug to me. @hallelujah or @guicassolato (or anybody else from your team), can you confirm this is the expected behavior? (quoted here for you)
> 
> ```
> When a MappingRule is added or removed all the existing MappingRule positions in System are reordered. For example, if there are 3 MappingRules with positions 1,2,7 respectively and a new MappingRule is added with position set to 6, the original MappingRule with position 7 is changed by system to position 8.
> ```

It is not a bug, it is a feature, mapping rules are ordered queue.
We need strict position number for APIcast to apply the correct rules in order. Two rules cannot have the same position.

thomasmaas:
@hallelujah position attribute is not used by apicast, the actual position in the json config is what apicast uses.

eguzki:
> @hallelujah position attribute is not used by apicast (https://github.com/3scale/APIcast/blob/master/gateway/src/apicast/configuration.lua#L27), the actual position in the json config is what apicast uses. So to me the solution proposed in #452 makes more sense.

That would mean that the 3scale operator would _shadow_ `position` attribute of the 3scale API and delegate order to the location in the array of the CR. I think it is ok for now as `position` does not play any other role. In general, I tend to respect backend API's, but it is ok for this case.

@miguelsorianod if you agree, close this issue and I will review the other PR.

hallelujah:
> > @hallelujah position attribute is not used by apicast (https://github.com/3scale/APIcast/blob/master/gateway/src/apicast/configuration.lua#L27), the actual position in the json config is what apicast uses. So to me the solution proposed in #452 makes more sense.
> 
> That would mean that the 3scale operator would _shadow_ `position` attribute of the 3scale API and delegate order to the location in the array of the CR. I think it is ok for now as `position` does not play any other role. In general, I tend to respect backend API's, but it is ok for this case.
> 
> @miguelsorianod if you agree, close this issue and I will review the other PR.

Yes the position is not used by APIcast, but the ordering should be strict. So we cannot have 2 same positions.

miguelsorianod:
> @miguelsorianod if you agree, close this issue and I will review the other PR.

I agree.

PR #454, opened 2020-08-20
Configurable resource requirements for APIManager components

miguelsorianod:
This PR adds configurable resource requirements for APIManager components.

### Summary of the implemented behaviour

* Configurable resource requirements per component has been added
* When resource requirements are not specified at component level the global defaults for the component are used for it (if enabled)
* When resource requirements are specified at component level, they take over preference over the global default and they follow a REPLACE behaviour (instead of merge behaviour) with the defaults.

### Detailed implemented behaviour

* global-level defaults for resource limits and requests for the components of 3scale are provided. This global-level defaults can be enabled or disabled through a CR attribute that already existed before this PR, named 'resourceRequirementsEnabled'.
* component-level *resources* attribute is provided at component level (for example spec.backend.listenerSpec.resources attribute):
 * If it is not set by the user or set to nil and the global-level CR attribute is not set to true then no resource limits and/or requests are enforced
 * If it is not set by the user or set to nil and the global-level CR attribute is set to true then the defaults for that given component are set
 * If it is set by the user to a non-nil value then the resources requests and limits set are the ones specified by the user with a REPLACE behaviour, not a merge behavior. Global defaults are not used for that component then, independently of the value of the global-level CR attribute. In this case, omitting a limit, request or cpu,memory inside a limit o request is equivalent to not enforcing it for that component. This is the same behavior that ResourceLimits have in Kubernetes.

#### Possible use case scenarios
1. The user wants to have default limits and requests for all components enabled and wants to modify them for a specific DeploymentConfig
  - The user is able to do that by setting the global resourceRequirementsEnabled attribute to true (default behavior) and then specifying the desired
    values on the corresponding *resources* attribute provided at component level in the APIManager CR
2. The user wants to have default limits and requests for all components disabled and wants to set them with custom values for a specific DeploymentConfig
  - The user is able to do that by setting the global resourceRequirementsEnabled attribute to false and then specifying the desired values on the
    corresponding *resources* attribute provided at component level in the APIManager CR
  - Limitation: The user has no way to specify not having a default limit and having a custom request for that same component or viceversa
3. The user wants to have default limits and requests for all components enabled and wants to disable them for a specific DeploymentConfig
  - The user is able to do that by setting the global resourceRequirementsEnabled attribute to true (default behavior) and then setting the corresponding
    *resources* attribute provided at component level in the APIManager CR as the empty map value `{}`

#### Unsupported scenarios

Basically the ones that involve a potentially desired merge behavior between global defaults and custom values:
1. On a given component, having a global default resource request and a custom resource limit for that same component or viceversa
2. On a given component, not having a global default resource request and having a custom resource limit for the same component or viceversa
3. On a given component, having a global default CPU and having custom Memory, or having global default Memory and having custom CPU for resource requirements
4. On a given component, having a global default CPU and having custom Memory, or having global default Memory and using custom CPU  for resource limits

Allowing those unsupported scenarios would require us to stop using the v1.ResourceRequirements type provided by Kubernetes in our CR and provide our own type where we can differentiate between setting resource requests to nil or empty, as the Kubernetes implementation treats those values with the same behavior, which is not enforcing them in that case.
The alternative implementation:
* Increases complexity
* May not be intuitive for users as it differs from how v1.ResourceRequirements field works (nil, vs empty, ...)
* Removes flexibility. Right now by reusing v1.ResourceRequirements other possible additional resources can be specified potentially, we would not be limited to CPU and requests. We would not be able to take advantage on changes/improvements on that type
* Allows more configuration combinations between defaults and custom resources at both requests/limits level and cpu/memory inside requests/limits
* Opens up new additional questions about how to specify cpus and memory and other potential resources (provide a map as kubernetes does, attribute fields? use Quantity attribute? etc...)

- [x] Documentation

### DIFF
--- a/deploy/crds/apps.3scale.net_apimanagers_crd.yaml
+++ b/deploy/crds/apps.3scale.net_apimanagers_crd.yaml
@@ -670,6 +670,26 @@ spec:
                     replicas:
                       format: int64
                       type: integer
+                    resources:
+                      description: ResourceRequirements describes the compute resource
+                        requirements.
+                      properties:
+                        limits:
+                          additionalProperties:
+                            type: string
+                          description: 'Limits describes the maximum amount of compute
+                            resources allowed. More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/'
+                          type: object
+                        requests:
+                          additionalProperties:
+                            type: string
+                          description: 'Requests describes the minimum amount of compute
+                            resources required. If Requests is omitted for a container,
+                            it defaults to Limits if that is explicitly specified,
+                            otherwise to an implementation-defined value. More info:
+                            https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/'
+                          type: object
+                      type: object
                     tolerations:
                       items:
                         description: The pod this Toleration is attached to tolerates
@@ -1347,6 +1367,26 @@ spec:
                     replicas:
                       format: int64
                       type: integer
+                    resources:
+                      description: ResourceRequirements describes the compute resource
+                        requirements.
+                      properties:
+                        limits:
+                          additionalProperties:
+                            type: string
+                          description: 'Limits describes the maximum amount of compute
+                            resources allowed. More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/'
+                          type: object
+                        requests:
+                          additionalProperties:
+                            type: string
+                          description: 'Requests describes the minimum amount of compute
+                            resources required. If Requests is omitted for a container,
+                            it defaults to Limits if that is explicitly specified,
+                            otherwise to an implementation-defined value. More info:
+                            https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/'
+                          type: object
+                      type: object
                     tolerations:
                       items:
                         description: The pod this Toleration is attached to tolerates
--- a/deploy/crds/apps.3scale.net_apimanagers_crd.yaml
+++ b/deploy/crds/apps.3scale.net_apimanagers_crd.yaml
@@ -10353,6 +10353,26 @@ spec:
                     replicas:
                       format: int64
                       type: integer
+                    resources:
+                      description: ResourceRequirements describes the compute resource
+                        requirements.
+                      properties:
+                        limits:
+                          additionalProperties:
+                            type: string
+                          description: 'Limits describes the maximum amount of compute
+                            resources allowed. More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/'
+                          type: object
+                        requests:
+                          additionalProperties:
+                            type: string
+                          description: 'Requests describes the minimum amount of compute
+                            resources required. If Requests is omitted for a container,
+                            it defaults to Limits if that is explicitly specified,
+                            otherwise to an implementation-defined value. More info:
+                            https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/'
+                          type: object
+                      type: object
                     tolerations:
                       items:
                         description: The pod this Toleration is attached to tolerates
@@ -10992,6 +11012,25 @@ spec:
                             - topologyKey
                             type: object
                           type: array
+                      type: object
+                  type: object
+                databaseResources:
+                  description: ResourceRequirements describes the compute resource
+                    requirements.
+                  properties:
+                    limits:
+                      additionalProperties:
+                        type: string
+                      description: 'Limits describes the maximum amount of compute
+                        resources allowed. More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/'
+                      type: object
+                    requests:
+                      additionalProperties:
+                        type: string
+                      description: 'Requests describes the minimum amount of compute
+                        resources required. If Requests is omitted for a container,
+                        it defaults to Limits if that is explicitly specified, otherwise
+                        to an implementation-defined value. More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/'
                       type: object
                   type: object
                 databaseTolerations:
@@ -11668,6 +11707,26 @@ spec:
                     replicas:
                       format: int64
                       type: integer
+                    resources:
+                      description: ResourceRequirements describes the compute resource
+                        requirements.
+                      properties:
+                        limits:
+                          additionalProperties:
+                            type: string
+                          description: 'Limits describes the maximum amount of compute
+                            resources allowed. More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/'
+                          type: object
+                        requests:
+                          additionalProperties:
+                            type: string
+                          description: 'Requests describes the minimum amount of compute
+                            resources required. If Requests is omitted for a container,
+                            it defaults to Limits if that is explicitly specified,
+                            otherwise to an implementation-defined value. More info:
+                            https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/'
+                          type: object
+                      type: object
                     tolerations:
                       items:
                         description: The pod this Toleration is attached to tolerates
--- a/deploy/crds/apps.3scale.net_apimanagers_crd.yaml
+++ b/deploy/crds/apps.3scale.net_apimanagers_crd.yaml
@@ -4829,6 +4829,66 @@ spec:
                               type: array
                           type: object
                       type: object
+                    developerContainerResources:
+                      description: ResourceRequirements describes the compute resource
+                        requirements.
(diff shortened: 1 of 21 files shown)


eguzki:
```
Configurable resource requirements per component has been added
```

:+1: per component is  per container.

```
When resource requirements are not specified at component level the global defaults for the component are used for it (if enabled)
```

Good. Intuitive and common sense.    

```
When resource requirements are specified at component level, they take over preference over the global default and they follow a REPLACE behaviour (instead of merge behaviour) with the defaults.
```
Agree. REPLACE behavior is easier and enough. Merge behavior (with defaults) can be confusing and requires not standard data structure type in the CRD to define resource requirements. REPLACE behavior requires, though, to know existing defaults. A user may want to update cpu limit and leave cpu requests, memory limits and memory requests "as defaults". So the user needs to create structure with desired cpu limit and copy other values from defaults.  For example:

```
resources:
      requests:
        memory: "64Mi" <- copied from defaults
        cpu: "250m" <- copied from defaults
      limits:
        memory: "128Mi" <- copied from defaults
        cpu: "500m" <- desired cpu value
```

For that, default values should be documented.

deploy/crds/apps.3scale.net_apimanagers_crd.yaml
```diff
@@ -670,6 +670,26 @@ spec:
...
+                    resources:
+                      description: ResourceRequirements describes the compute resource
+                        requirements.
+                      properties:
+                        limits:
+                          additionalProperties:
+                            type: string
+                          description: 'Limits describes the maximum amount of compute
+                            resources allowed. More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/'
+                          type: object
```
eguzki, line 682: `cpu` and `memory` is not defined here?

pkg/3scale/amp/operator/apicast_options_provider.go
```diff
@@ -60,6 +60,18 @@ func (a *ApicastOptionsProvider) setResourceRequirementsOptions() {
 		a.apicastOptions.ProductionResourceRequirements = v1.ResourceRequirements{}
 		a.apicastOptions.StagingResourceRequirements = v1.ResourceRequirements{}
 	}
+
+	// DeploymentConfig-level ResourceRequirements CR fields have priority over
+	// spec.resourceRequirementsEnabled, overwriting that setting when they are
+	// defined
+	if a.apimanager.Spec.Apicast.ProductionSpec.Resources != nil {
```
eguzki, line 67: `a.apimanager.Spec.Apicast.ProductionSpec` is not nil always?

pkg/3scale/amp/operator/backend_options_provider.go
```diff
@@ -156,6 +156,19 @@ func (o *OperatorBackendOptionsProvider) setResourceRequirementsOptions() {
 		o.backendOptions.WorkerResourceRequirements = v1.ResourceRequirements{}
 		o.backendOptions.CronResourceRequirements = v1.ResourceRequirements{}
 	}
+
+	// DeploymentConfig-level ResourceRequirements CR fields have priority over
+	// spec.resourceRequirementsEnabled, overwriting that setting when they are
+	// defined
+	if o.apimanager.Spec.Backend.ListenerSpec.Resources != nil {
```
eguzki, line 163: `o.apimanager.Spec.Backend.ListenerSpec` not nil always?

pkg/3scale/amp/operator/zync_options_provider.go
```diff
@@ -117,6 +117,19 @@ func (z *ZyncOptionsProvider) setResourceRequirementsOptions() {
...
 	}
+
+	// DeploymentConfig-level ResourceRequirements CR fields have priority over
+	// spec.resourceRequirementsEnabled, overwriting that setting when they are
+	// defined
+	if z.apimanager.Spec.Zync.AppSpec.Resources != nil {
+		z.zyncOptions.ContainerResourceRequirements = *z.apimanager.Spec.Zync.AppSpec.Resources
+	}
+	if z.apimanager.Spec.Zync.QueSpec.Resources != nil {
+		z.zyncOptions.ContainerResourceRequirements = *z.apimanager.Spec.Zync.QueSpec.Resources
```
eguzki, line 128: should be `z.zyncOptions.QueContainerResourceRequirements`

eguzki (changes requested):
Overall LGTM.

Left some comments. I also miss some option provider tests.

pkg/3scale/amp/operator/zync_options_provider.go
miguelsorianod: :+1:

pkg/3scale/amp/operator/backend_options_provider.go
miguelsorianod: Yes.

We initialize it in the `setDefaults` method for the CR

pkg/3scale/amp/operator/apicast_options_provider.go
miguelsorianod: Yes.

We initialize it in the `setDefaults` method for the CR

deploy/crds/apps.3scale.net_apimanagers_crd.yaml
miguelsorianod: Nope.

The k8s value for limits/requests is:
```
type ResourceRequirements struct {
	// Limits describes the maximum amount of compute resources allowed.
	// More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/
	// +optional
	Limits ResourceList `json:"limits,omitempty" protobuf:"bytes,1,rep,name=limits,casttype=ResourceList,castkey=ResourceName"`
	// Requests describes the minimum amount of compute resources required.
	// If Requests is omitted for a container, it defaults to Limits if that is explicitly specified,
	// otherwise to an implementation-defined value.
	// More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/
	// +optional
	Requests ResourceList `json:"requests,omitempty" protobuf:"bytes,2,rep,name=requests,casttype=ResourceList,castkey=ResourceName"`
}
```

Being ResourceList:
```
type ResourceList map[ResourceName]resource.Quantity
```

miguelsorianod:
Remanining tests added.

Documentation pending.

miguelsorianod:
Added documentation

doc/apimanager-reference.md
```diff
@@ -25,7 +25,7 @@ This resource is the resource used to deploy a 3scale API Management solution.
...
 | ImageStreamTagImportInsecure | `imageStreamTagImportInsecure` | bool | No | `false` | Set to true if the server may bypass certificate verification or connect directly over HTTP during image import |
-| ResourceRequirementsEnabled | `resourceRequirementsEnabled` | bool | No | `true` | When true, 3Scale API management solution is deployed with the optimal resource requirements and limits. Setting this to false removes those resource requirements. ***Warning*** Only set it to false for development and evaluation environments |
+| ResourceRequirementsEnabled | `resourceRequirementsEnabled` | bool | No | `true` | When true, 3Scale API management solution is deployed with the optimal resource requirements and limits. Setting this to false removes those resource requirements. ***Warning*** Only set it to false for development and evaluation environments. When set to `true`, default compute resources are set for the APIManager components. See [Default APIManager's components compute resources](#Default-APIManager's-components-compute-resources) to see the default assigned values |
```
eguzki, line 28: The anchor link to default values is not correct

eguzki (changes requested):

eguzki (approved):

eguzki:
fix the broken link and merge.

doc/apimanager-reference.md
```diff
@@ -432,3 +450,163 @@ existing secret names.
...
+
+### Default APIManager's components compute resources
+
+When APIManager's `spec.resourceRequirementsEnabled` attribute is set to
+`true`, default compute resources are set for the APIManager components.
+
+The specific compute resources default values that are set for the
+APIManager components are the following ones:
+
+#### system-app's system-master container
```
eguzki, line 463: Up to you to change or leave like this. What about table format? it could be more readable
| **Component** | **CPU request**| **CPU limit** | **Memory request** | **Memory request** |
| --- | --- | --- | --- | --- |
| system-master | 50m | 1000m | 600Mi | 800Mi |

doc/apimanager-reference.md
miguelsorianod: In my local markdown editor seems to work correctly. Apparently in GitHub works differently.
I've changed it

doc/apimanager-reference.md
miguelsorianod: Table format looks better, more compact. Changed :+1:

Issue #453, opened 2020-08-18
[system] Add PROMETHEUS_EXPORTER_PORT environment


hallelujah:
We are adding more prometheus exporter endpoint to System component.

- system-app
- system-sidekiq

Each container of each pod should have different `PROMETHEUS_EXPORTER_PORT` value

Container | Port | Service
---------------|---------|-----
system-master |  9394 | web server
system-provider |  9395 | web server
system-developer |  9396 | web server
system-sidekiq | 9394 | sidekiq

The endpoint are accessible from `http://{server}:{port}/metrics` where `server` is normally `0.0.0.0` locally

eguzki:
For SystemApp pods containers, just let us know the port numbers and we will configure make all the arrangements to have prometheus scraping metrics. 

For systemSidekiq, Currently there are exposed metrics at port 9394 https://github.com/3scale/3scale-operator/blob/df0c0e962f5fd42926d52d316090d574e2ba4ae5/pkg/3scale/amp/component/system.go#L99 If you change it or add a new metric service, let us know the port number to do all the integration with the prometheus.

hallelujah:
@eguzki I updated the table, you can take them as the reference

miguelsorianod:
Hi @hallelujah,

When you say `Each container of each pod should have different PROMETHEUS_EXPORTER_PORT value` do you mean we should set/send the environment variable `PROMETHEUS_EXPORTER_PORT` to the container?

miguelsorianod:
What are the endpoints for the metrics in the system-app containers? `/metrics` ?

miguelsorianod:
@hallelujah,

As @eguzki said, For system-sidekiq we are already using SystemSidekiqMetricsPort = 9394
Can you make that the default for Sidekiq in system code?

hallelujah:
Closing as those port are defaults if the environment variable is not set see https://github.com/3scale/porta/pull/2093

PR #458, opened 2020-08-31
Add system-app prometheus monitoring

miguelsorianod:
This PR adds Prometheus Monitoring for system-app containers.

The expected environment variables and port numberings for system-app containers Prometheus functionality has been defined in https://github.com/3scale/3scale-operator/issues/453. This PR uses thoses values.

- [x] Upgrade
- [x] Testing upgrade

### DIFF
--- a/pkg/3scale/amp/component/system.go
+++ b/pkg/3scale/amp/component/system.go
@@ -97,14 +97,12 @@ const (
 )
 
 const (
-	SystemSidekiqMetricsPort = 9394
-)
-
-const (
-	SystemAppPrometheusExporterPortEnvVarName = "PROMETHEUS_EXPORTER_PORT"
-	SystemAppMasterContainerPrometheusPort    = 9395
-	SystemAppProviderContainerPrometheusPort  = 9396
-	SystemAppDeveloperContainerPrometheusPort = 9394
+	SystemAppPrometheusExporterPortEnvVarName     = "PROMETHEUS_EXPORTER_PORT"
+	SystemSidekiqPrometheusExporterPortEnvVarName = "PROMETHEUS_EXPORTER_PORT"
+	SystemSidekiqMetricsPort                      = 9394
+	SystemAppMasterContainerPrometheusPort        = 9395
+	SystemAppProviderContainerPrometheusPort      = 9396
+	SystemAppDeveloperContainerPrometheusPort     = 9394
 )
 
 type System struct {
@@ -270,6 +268,15 @@ func (system *System) buildAppDeveloperContainerEnv() []v1.EnvVar {
 	result := system.buildSystemBaseEnv()
 	if system.Options.AppMetrics {
 		result = append(result, helper.EnvVarFromValue(SystemAppPrometheusExporterPortEnvVarName, strconv.Itoa(SystemAppDeveloperContainerPrometheusPort)))
+	}
+
+	return result
+}
+
+func (system *System) buildSystemSidekiqContainerEnv() []v1.EnvVar {
+	result := system.buildSystemBaseEnv()
+	if system.Options.SideKiqMetrics {
+		result = append(result, helper.EnvVarFromValue(SystemSidekiqPrometheusExporterPortEnvVarName, strconv.Itoa(SystemSidekiqMetricsPort)))
 	}
 
 	return result
@@ -583,16 +590,10 @@ func (system *System) AppDeploymentConfig() *appsv1.DeploymentConfig {
 					Volumes:     system.appPodVolumes(),
 					Containers: []v1.Container{
 						v1.Container{
-							Name:  "system-master",
-							Image: "amp-system:latest",
-							Args:  []string{"env", "TENANT_MODE=master", "PORT=3002", "container-entrypoint", "bundle", "exec", "unicorn", "-c", "config/unicorn.rb"},
-							Ports: []v1.ContainerPort{
-								v1.ContainerPort{
-									Name:          "master",
-									HostPort:      0,
-									ContainerPort: 3002,
-									Protocol:      v1.ProtocolTCP},
-							},
+							Name:         "system-master",
+							Image:        "amp-system:latest",
+							Args:         []string{"env", "TENANT_MODE=master", "PORT=3002", "container-entrypoint", "bundle", "exec", "unicorn", "-c", "config/unicorn.rb"},
+							Ports:        system.appMasterPorts(),
 							Env:          system.buildAppMasterContainerEnv(),
 							Resources:    *system.Options.AppMasterContainerResourceRequirements,
 							VolumeMounts: system.appMasterContainerVolumeMounts(),
@@ -632,16 +633,10 @@ func (system *System) AppDeploymentConfig() *appsv1.DeploymentConfig {
 							StdinOnce:       false,
 							TTY:             false,
 						}, v1.Container{
-							Name:  "system-provider",
-							Image: "amp-system:latest",
-							Args:  []string{"env", "TENANT_MODE=provider", "PORT=3000", "container-entrypoint", "bundle", "exec", "unicorn", "-c", "config/unicorn.rb"},
-							Ports: []v1.ContainerPort{
-								v1.ContainerPort{
-									Name:          "provider",
-									HostPort:      0,
-									ContainerPort: 3000,
-									Protocol:      v1.ProtocolTCP},
-							},
+							Name:         "system-provider",
+							Image:        "amp-system:latest",
+							Args:         []string{"env", "TENANT_MODE=provider", "PORT=3000", "container-entrypoint", "bundle", "exec", "unicorn", "-c", "config/unicorn.rb"},
+							Ports:        system.appProviderPorts(),
 							Env:          system.buildAppProviderContainerEnv(),
 							Resources:    *system.Options.AppProviderContainerResourceRequirements,
 							VolumeMounts: system.appProviderContainerVolumeMounts(),
@@ -681,16 +676,10 @@ func (system *System) AppDeploymentConfig() *appsv1.DeploymentConfig {
 							StdinOnce:       false,
 							TTY:             false,
 						}, v1.Container{
-							Name:  "system-developer",
-							Image: "amp-system:latest",
-							Args:  []string{"env", "PORT=3001", "container-entrypoint", "bundle", "exec", "unicorn", "-c", "config/unicorn.rb"},
-							Ports: []v1.ContainerPort{
-								v1.ContainerPort{
-									Name:          "developer",
-									HostPort:      0,
-									ContainerPort: 3001,
-									Protocol:      v1.ProtocolTCP},
-							},
+							Name:         "system-developer",
+							Image:        "amp-system:latest",
+							Args:         []string{"env", "PORT=3001", "container-entrypoint", "bundle", "exec", "unicorn", "-c", "config/unicorn.rb"},
+							Ports:        system.appDeveloperPorts(),
 							Env:          system.buildAppDeveloperContainerEnv(),
 							Resources:    *system.Options.AppDeveloperContainerResourceRequirements,
 							VolumeMounts: system.appDeveloperContainerVolumeMounts(),
@@ -851,7 +840,7 @@ func (system *System) SidekiqDeploymentConfig() *appsv1.DeploymentConfig {
 							Name:            SystemSidekiqName,
 							Image:           "amp-system:latest",
 							Args:            []string{"rake", "sidekiq:worker", "RAILS_MAX_THREADS=25"},
-							Env:             system.buildSystemBaseEnv(),
+							Env:             system.buildSystemSidekiqContainerEnv(),
 							Resources:       *system.Options.SidekiqContainerResourceRequirements,
 							VolumeMounts:    system.sidekiqContainerVolumeMounts(),
 							ImagePullPolicy: v1.PullIfNotPresent,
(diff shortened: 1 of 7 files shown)


pkg/3scale/amp/component/system_monitoring.go
```diff
@@ -30,6 +30,75 @@ func (system *System) SystemSidekiqPodMonitor() *monitoringv1.PodMonitor {
...
+
+func (system *System) SystemAppProviderContainerPodMonitor() *monitoringv1.PodMonitor {
+	return &monitoringv1.PodMonitor{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   "system-app-master-container",
+			Labels: system.Options.CommonSidekiqLabels,
+		},
+		Spec: monitoringv1.PodMonitorSpec{
+			PodMetricsEndpoints: []monitoringv1.PodMetricsEndpoint{{
+				Port:   "masterMetrics",
```
miguelsorianod, line 72: Notice how here the port is 'masterMetrics' and not just 'metrics'. In system component I've assigned different `ContainerPort` names for each container in system-app. The reason for that is that it seems that PodMonitor is at PodLevel and the way to differentiate between endpoints in the pod is by the Port name or by path or by scheme.
So, to differentiate between the container ports in system-app I've had to set unique names for them.

What do you think?

pkg/3scale/amp/component/system_options.go
```diff
@@ -93,6 +93,7 @@ type SystemOptions struct {
 	MemcachedLabels          map[string]string `validate:"required"`
 	SMTPLabels               map[string]string `validate:"required"`
 	SideKiqMetrics           bool
+	AppMetrics               bool
```
miguelsorianod, line 96: Created AppMetrics option to be able to enable this only on operator side and not on templates side, following same approach as SidekiqMetrics

pkg/3scale/amp/component/system.go
```diff
@@ -241,6 +249,24 @@ func (system *System) buildSystemBaseEnv() []v1.EnvVar {
 	return result
 }
 
+func (system *System) buildAppMasterContainerEnv() []v1.EnvVar {
+	result := system.buildSystemBaseEnv()
+	result = append(result, helper.EnvVarFromValue(SystemAppPrometheusExporterPortEnvVarName, strconv.Itoa(SystemAppMasterContainerPrometheusPort)))
```
miguelsorianod, line 254: I'm not sure whether the definition of this environment variable should be conditional to the appMetrics boolean. We could do it but because there are several metrics ports in the same pod between containers leaving the default might cause port collision between containers in the templates approach if we don't include the environment variables?

On the other hand, kind of weird not including envvars conditionally but doing it for the containerports.

What do you think?

pkg/3scale/amp/component/system_monitoring.go
```diff
@@ -30,6 +30,75 @@ func (system *System) SystemSidekiqPodMonitor() *monitoringv1.PodMonitor {
...
 }
 
+func (system *System) SystemAppPodMonitor() *monitoringv1.PodMonitor {
+	return &monitoringv1.PodMonitor{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   "system-app",
+			Labels: system.Options.CommonSidekiqLabels,
+		},
+		Spec: monitoringv1.PodMonitorSpec{
+			PodMetricsEndpoints: []monitoringv1.PodMetricsEndpoint{
```
miguelsorianod, line 40: Is there any difference between doing this or creating 3 PodMonitors with each monitor including only PodMetricsEndpoints to a specific container?

Keep in mind the selector of PodMonitor is of Pod Labels (containers do not have labels on their own if I'm not wrong)

pkg/3scale/amp/component/system_monitoring.go
```diff
@@ -30,6 +30,37 @@ func (system *System) SystemSidekiqPodMonitor() *monitoringv1.PodMonitor {
...
+				{
+					Port:   "masterMetrics",
```
miguelsorianod, line 42: Notice how here the port is 'masterMetrics' and not just 'metrics'. In system component I've assigned different ContainerPort names for each container in system-app. The reason for that is that it seems that PodMonitor is at PodLevel and the way to differentiate between endpoints in the pod is by the Port name or by path or by scheme.
So, to differentiate between the container ports in system-app I've had to set unique names for them.

What do you think?

eguzki:
What about adding env var for sidekiq as well to define port number? The  upgrade process might not need as it the "upgrade" will be executed by the reconciliation logic.

eguzki:
This is about scraping metrics. But there are no new prometheus rules or new dashboards, thus those "new" metrics are not used at all. This PR could wait for contribution from System guys with dashboard and prometheus rules.

miguelsorianod:
> This is about scraping metrics. But there are no new prometheus rules or new dashboards, thus those "new" metrics are not used at all. This PR could wait for contribution from System guys with dashboard and prometheus rules.

This is what this PR is about. Addition of PrometheusRules and GrafanaDashboards will be done in another PR when System  team has defined and created them.

> What about adding env var for sidekiq as well to define port number? The upgrade process might not need as it the "upgrade" will be executed by the reconciliation logic.

I can do that, but that still doesn't solve what I commented in https://github.com/3scale/3scale-operator/pull/458#discussion_r480071804. Even if there are no new prometheus rules or dashboards and the new metrics "are not used", those processes will be running on the templates approach

eguzki:
> I can do that, but that still doesn't solve what I commented in #458 (comment). Even if there are no new prometheus rules or dashboards and the new metrics "are not used", those processes will be running on the templates approach

That is another good reason to make them conditional. For templates always no and for operator if enabled in CR.

miguelsorianod:
> That is another good reason to make them conditional. For templates always no and for operator if enabled in CR.

See this comment: https://github.com/3scale/3scale-operator/pull/458#discussion_r480071804
That's precisely my concern. By disabling it in templates I fear ports between container might collision

eguzki:
I said early:
>If there are conflicts between containers in the same pods, we will detect it very quickly and fix it either by changing defaults or setting env vars unconditionally.

@hallelujah can you make sure default metric ports for system-master, system-developer, system-provider containers do not conflict? Should be part of https://github.com/3scale/porta/pull/2093 . The main reason is to avoid changes in templates (which are being deprecated but still we need to support them). For the operator, we will be using env vars to set them.


Defaults can be: 

Container | Port | Service
---------------|---------|-----
system-master | TBD / 9394 | web server
system-provider | TBD / 9395 | web server
system-developer | TBD / 9396 | web server
system-sidekiq | 9359 | sidekiq

hallelujah:
@eguzki if you do not set the `PROMETHEUS_EXPORTER_PORT` in each container they will be set by default to 9394.
So they will conflict and make the application crash. It is needed to set them per container

eguzki:
@hallelujah then that obliges us to make changes in templates as well. We were going to add `PROMETHEUS_EXPORTER_PORT` anyway to the operator, but since templates are kind of deprecated, we are said to keep template changes as few as possible. 

For you it would be change few constants. For us it means changes in templates deployment, besides upgrade procedure from 2.9 to 2.10. 

Can you change the default port numbers easily so they are different for each container?

eguzki:
Upgrade is needed. If we do not upgrade, when deploying 2.10 images metric ports would conflict because currently `PROMETHEUS_EXPORTER_PORT` is not set.

eguzki:
What about computing the default value of the metric ports as 
``` 6000 + UNICORN_PORT```? That way we make sure no conflict happens. Defaults would be used in templates (but nobody would scrape metrics, though). For operator, we will be using `PROMETHEUS_EXPORTER_PORT`.

**Comment by hallelujah** [comment]
To summarize:

- sidekiq already has http://localhost:9394/metrics and it won't change. Leaving PROMETHEUS_EXPORTER_PORT as empty will default to it
- system-app containers will have new http://localhost:9394/metrics http://localhost:9395/metrics http://localhost:9396/metrics
Those will be the defaults if the `PROMETHEUS_EXPORTER_PORT` is not assigned

**Comment by miguelsorianod** [comment]
I updated the PR to remove the environment variables setting in templates scenario (to avoid needing an upgrade procedure there)
I removed the "Position" attribute from the Backend and Product CRs (in MappingRuleSpec) that for some reason it was already defined but not used anywhere.

**Review by eguzki — CHANGES_REQUESTED** [review]

File: pkg/3scale/amp/component/system_monitoring.go (hunk 4 above)
Comment (miguelsorianod) [reply pkg/3scale/amp/component/system_monitoring.go:40]: :+1:

File: pkg/3scale/amp/component/system_monitoring.go (hunk 5 above)
Comment (miguelsorianod) [reply pkg/3scale/amp/component/system_monitoring.go:42]: :+1:

File: pkg/3scale/amp/component/system.go (hunk 3 above)
Comment (miguelsorianod) [reply pkg/3scale/amp/component/system.go]: We need to write an ugrade procedure as they are not reconciled automatically

**Comment by miguelsorianod** [comment]
I have added upgrade procedure to the operator part and made some minor changes.

@eguzki please review whenever you can. Upgrade will be tested once we agree on the changes :+1:

**Comment by eguzki** [comment]
Looking good so far

**Comment by miguelsorianod** [comment]
When doing the upgrade tests the following errors have been detected:

```
2020-09-14T13:40:01.619+0200	ERROR	controller-runtime.controller	Reconciler error	{"controller": "apimanager-controller", "request": "msoriano-test/example-apimanager", "error": "Upgrading monitoring settings: DeploymentConfig.apps.openshift.io \"system-app\" is invalid: [spec.template.spec.containers[0].ports[1].name: Invalid value: \"masterMetrics\": must contain only alpha-numeric characters (a-z, 0-9), and hyphens (-), spec.template.spec.containers[1].ports[1].name: Invalid value: \"providerMetrics\": must contain only alpha-numeric characters (a-z, 0-9), and hyphens (-), spec.template.spec.containers[2].ports[1].name: Invalid value: \"developerMetrics\": must be no more than 15 characters, spec.template.spec.containers[2].ports[1].name: Invalid value: \"developerMetrics\": must contain only alpha-numeric characters (a-z, 0-9), and hyphens (-)]"}


2020-09-14T13:55:24.333+0200	ERROR	controller_apimanager	Error upgrading APIManager	{"Request.Namespace": "msoriano-test", "Request.Name": "example-apimanager", "error": "Upgrading monitoring settings: DeploymentConfig.apps.openshift.io \"system-app\" is invalid: [spec.template.spec.containers[1].ports[1].name: Invalid value: \"provider-metrics\": must be no more than 15 characters, spec.template.spec.containers[2].ports[1].name: Invalid value: \"developer-metrics\": must be no more than 15 characters]"}
```

Apparently container port names can only be lowercase (and digits and hyphens) and they can't be longer tan 15 characters.
I've renamed the container port names to to:
```
master-metrics
prov-metrics
dev-metrics
```

**Comment by miguelsorianod** [comment]
Upgrade has been tested:
* prometheus envvars are added to system-app and system-sidekiq pods
* metrics ports have are added in system-app containers
* system-app PodMonitor is created

**Comment by miguelsorianod** [comment]
When testing the upgrade I've found an interesting behaviour:

When making requests to the different metric endpoints in system-app pod:
curl 127.0.0.1:9394/metrics
curl 127.0.0.1:9395/metrics
curl 127.0.0.1:9396/metrics

The results are the same between all endpoints. I think metrics are being aggregated and all ports metrics endpoints return the same aggregated result.

**Comment by miguelsorianod** [comment]
Ready for review @eguzki

**Comment by hallelujah** [comment]
> When testing the upgrade I've found an interesting behaviour:
> 
> When making requests to the different metric endpoints in system-app pod:
> curl 127.0.0.1:9394/metrics
> curl 127.0.0.1:9395/metrics
> curl 127.0.0.1:9396/metrics
> 
> The results are the same between all endpoints. I think metrics are being aggregated and all ports metrics endpoints return the same aggregated result.
> 
> If that's the case, I guess we can leave the PodMonitor as it is (referencing the different ports) for redundancy purposes, or just arbitrarily choose only one of the ports for it.

Yes for now they are aggregated due to the sidekiq-prometheus gem 

We plan on removing this strange behaviour by using yabeda. There is another temporary endpoint you can query:

curl 127.0.0.1:9394/yabeda-metrics
curl 127.0.0.1:9395/yabeda-metrics
curl 127.0.0.1:9396/yabeda-metrics

Once everything is satisfying we are going to remove sidekiq-prometheus, we will keep you in touch
Remember sidekiq is

**Comment by hallelujah** [comment]
:warning: Due to an issue with current templates (developer container does not have TENANT_MODE)I had to change the ports so now they are:

- developer: http://localhost:9394
- provider: http://localhost:9395
- master: http://localhost:9396

https://github.com/3scale/porta/pull/2123/commits/3daab2119b8ff28ce51c24b3ffdcf550ca5f748d

**Review by eguzki — APPROVED** [review]

File: pkg/3scale/amp/component/system.go [hunk 6]
```diff
@@ -92,11 +93,24 @@ const (
...
 )
 
 const (
-	SystemSidekiqMetricsPort = 9394
+	SystemAppPrometheusExporterPortEnvVarName     = "PROMETHEUS_EXPORTER_PORT"
+	SystemSidekiqPrometheusExporterPortEnvVarName = "PROMETHEUS_EXPORTER_PORT"
+	SystemSidekiqMetricsPort                      = 9394
+	SystemAppMasterContainerPrometheusPort        = 9394
+	SystemAppProviderContainerPrometheusPort      = 9395
+	SystemAppDeveloperContainerPrometheusPort     = 9396
```
Martouta, line 110: AFAIK these ports don't match what we have in the System code 😕 
https://github.com/3scale/porta/blob/9c175735a7ada82e33fc90cab4a8b762c8a7ad9d/lib/prometheus_exporter_port.rb#L5-L10

Martouta (changes requested):
The ports seem to be wrong. Ask @hallelujah in case of doubt.
https://github.com/3scale/3scale-operator/pull/458/files#r490694337

hallelujah:
> The ports seem to be wrong. Ask @hallelujah in case of doubt.
> https://github.com/3scale/3scale-operator/pull/458/files#r490694337

https://github.com/3scale/3scale-operator/pull/458#issuecomment-694301232

pkg/3scale/amp/component/system.go
eguzki: Do not have to match. System code defines default ports if not set. We have decided to set them in order to know exactly port numbers. As we do for other components as well.

Is it an issue? Do those port numbers get in conflict with other services in the same pod?

Martouta: @hallelujah should say. I do not know about this, but I only remember that when we reviewed his PR in Porta/System, he told us that those ports numbers came from an agreement made with you guys. And later those changed for an issue. So... if they changed in System, why not here? if they do not need to match, why did we have to agree in the port numbers in the first place?

eguzki: Because we need those default ports to be different for templates. In templates ports are not set, monitoring only enabled in operator based deployments.

miguelsorianod: I think it's better to have them match though, so between templates and operator the used ports are effectively the same.

I'll change it to match the new values if there's no objection.

miguelsorianod: Updated to the new values

miguelsorianod:
> curl 127.0.0.1:9394/yabeda-metrics
> curl 127.0.0.1:9395/yabeda-metrics
> curl 127.0.0.1:9396/yabeda-metrics
> 
> Once everything is satisfying we are going to remove sidekiq-prometheus, we will keep you in touch
> Remember sidekiq is

For the moment should we be setting /metrics instead of /yabeda-metrics then? It's not clear to me what should we be using for 3scale 2.10

hallelujah:
Never use the `/yabeda-metrics` it is experimental and future metrics will lie there. At some point we will replace `/metrics` by the content of `/yabeda-metrics`

miguelsorianod:
Thanks for the clarification @hallelujah

eguzki (approved):

PR #461, opened 2020-09-03
operator deployment resource requests

eguzki:
Jira Issue: https://issues.redhat.com/browse/THREESCALE-5860

What
----------

Currently, no requests or limits for CPU and memory are set for the 3scale operator.
Current QoS class

Best Effort

How
------

Requests and limits for CPU and memory should be set in the operator deployment which will be used for the CSV generation.


miguelsorianod:
What was the conclusion about limits?

I didn't send this PR as a decision on that didn't seem taken seeing the comments where it is being discussed now and the feedback we got

eguzki:
There is no hard opinion about limits. The little benefit of making the operator restart when there are memory leaks is not seen in other operators. 

We can always add limits later if there is a good reason to do so.

miguelsorianod (approved):

miguelsorianod:
I updated the PR to remove the environment variables setting in templates scenario (to avoid needing an upgrade procedure there)
I removed the "Position" attribute from the Backend and Product CRs (in MappingRuleSpec) that for some reason it was already defined but not used anywhere.

PR #462, opened 2020-09-04
Add ConsoleLink Creation/Remove

myeung18:
Signed-off-by: Marco Yeung <myeung@redhat.com>

Added ConsoleLink Creation/Remove and upgraded client-go/RHsyseng libraries for OCP4.3 or above


myeung18:
This new PR is to replace ##448.

deploy/cluster_role_binding.yaml
```diff
@@ -0,0 +1,11 @@
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: 3scale-operator
+subjects:
+  - kind: ServiceAccount
+    name: 3scale-operator
```
eguzki, line 7: I wonder what happens when `namespace` is not defined. Potentially, there could be one `3scale-operator` service account on each namespace.

deploy/cluster_role.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
eguzki, line 2: I guess `ClusterRole` is required because `consolinks` are kind of goblal resources. Is this true? Is it not enough to add this to the existing `3scale-operator` namespaced role?

go.mod
```diff
@@ -59,4 +59,6 @@ replace (
 
 replace github.com/docker/docker => github.com/moby/moby v0.7.3-0.20190826074503-38ab9da00309 // Required by Helm
 
-replace github.com/openshift/api => github.com/openshift/api v0.0.0-20190924102528-32369d4db2ad // Required until https://github.com/operator-framework/operator-lifecycle-manager/pull/1241 is resolved
+replace github.com/openshift/api => github.com/openshift/api v0.0.0-20200527184302-a843dc3262a0 // Required until https://github.com/operator-framework/operator-lifecycle-manager/pull/1241 is resolved
+
+replace github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20191125132246-f6563a70e19a
```
eguzki, line 64: Is this a hard requirement to manage console link resources? current openshift go client version does not support consolelinks?

pkg/apis/addtoscheme_apps_v1alpha1.go
```diff
@@ -2,9 +2,15 @@ package apis
...
+	routev1 "github.com/openshift/api/route/v1"
+	consolev1 "github.com/openshift/api/console/v1"
+
 )
 
 func init() {
 	// Register the types with the Scheme so the components can map objects to GroupVersionKinds and back
-	AddToSchemes = append(AddToSchemes, v1alpha1.SchemeBuilder.AddToScheme)
+	AddToSchemes = append(AddToSchemes,
+		routev1.Install,
```
eguzki, line 13: `routev1` is being registered in https://github.com/3scale/3scale-operator/blob/c941a204e884f9df6eb1cccc33928db37a3ca568/cmd/manager/main.go#L116

Any good reason to do here again?

Same goes for `consolev1`. Should be registered in `cmd/manager/main.go` or here?

pkg/controller/apimanager/apimanager_controller.go
```diff
@@ -101,6 +105,11 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	err = c.Watch(&source.Kind{Type: &routev1.Route{}}, &handler.EnqueueRequestForObject{})
```
eguzki, line 108: As watched routes are not managed by the apimanager controller, I think it makes sense to create a separate controller for this logic. Note that this apimanager controller always tries to read APIManager type out of reconcile request object (https://github.com/3scale/3scale-operator/blob/c941a204e884f9df6eb1cccc33928db37a3ca568/pkg/controller/apimanager/apimanager_controller.go#L143)

pkg/controller/apimanager/apimanager_controller.go
```diff
@@ -263,6 +282,15 @@ func (r *ReconcileAPIManager) reconcileAPIManagerLogic(cr *appsv1alpha1.APIManag
 		}
 	}
 
+	reader := read.New(r.Client()).WithNamespace(cr.Namespace).WithOwnerObject(cr)
```
eguzki, line 285: What is the purpose of these few lines of code?

pkg/controller/apimanager/apimanager_controller.go
```diff
@@ -216,10 +225,20 @@ func (r *ReconcileAPIManager) apiManagerInstance(namespacedName types.Namespaced
...
+					Name: namespacedName.Name,
+					Namespace: namespacedName.Namespace,
+				}
+				helper.RemoveConsoleLink(r.Context(), r.Client(), instance)
+			}
 			return nil, nil
 		}
 		return nil, err
 	}
+	if err := helper.ConsoleLinkSupported(); err == nil {
```
eguzki, line 239: I do not think apimanager controller is the one appropriate for this logic. But if it was, this would not be the right place. It would be inside the `reconcileAPIManagerLogic` method

pkg/helper/webconsole.go
```diff
@@ -0,0 +1,112 @@
...
+
+func getRoute(ctx context.Context, c client.Client, apimanager *appsv1alpha1.APIManager) *routev1.Route {
+	reader := read.New(c).WithNamespace(apimanager.Namespace)
+	deployedRoutes, err := reader.List(&routev1.RouteList{})
+	if err != nil {
+		return nil
+	}
+	for _, route := range deployedRoutes {	//look for 3scale admin routes, capture route object
+		realr := route.(*routev1.Route)
+		if strings.Compare(realr.Spec.To.Name, "system-provider") == 0 {
```
eguzki, line 59: We should start with `system-master` route. There could exist N routes with `system-provider` as value in the `To` field.

pkg/helper/webconsole.go
```diff
@@ -0,0 +1,112 @@
...
+func ConsoleLinkSupported() error {
+	gvk := schema.GroupVersionKind{Group: "console.openshift.io", Version: "v1", Kind: "ConsoleLink"}
+	return kubernetes.CustomResourceDefinitionExists(gvk)
+}
+
+// CreateConsoleLink creates a ConsoleLink object if it doesn't already exist
+func CreateConsoleLink(ctx context.Context, c client.Client, apimanager *appsv1alpha1.APIManager) {
+	route := getRoute(ctx, c, apimanager)
+	if route != nil {
+		consoleLinkName := fmt.Sprintf("%s-%s", apimanager.ObjectMeta.Name, apimanager.Namespace)
```
eguzki, line 35: Try to create `ReconcileConsoleLink` method of the `BaseAPIManagerLogicReconciler`  type in `base_apimanager_logic_reconciler.go`. You should have everything you need and the implementation should be easy.

eguzki (changes requested):

deploy/cluster_role_binding.yaml
myeung18: The yaml won't create if namespace is missed, now added back.

deploy/cluster_role.yaml
myeung18: The consolelinks is clusterwide resource, we must have a ClusterRole.

go.mod
myeung18: Consolelink exists in this existing version, but it will need further investigation to confirm if it works just like the one in the newer version in this new client-go.
However, the operator-util we used to verify consolelink before add/update it does request this new client-go, as operator-util is meant to support all console enhancement like consoleyamlsamples, login page etc.

pkg/controller/apimanager/apimanager_controller.go
myeung18: Further investigated, they are redundant and removed.

pkg/helper/webconsole.go
myeung18: I see the pattern 3scale is using to do the reconciliation, so adding the ReconcileConsoleLink method in that base reconciler, do you mean consolelink creation (as well as the DeleteConsoleLink method) should use this pattern as well?  That means, the desired and/or original Consolelink objects should be constructed at the right time, and eventually are passed to BaseReconciler.ReconcileResource() to determine the creation or update?  After that, the webconsole.go could be removed?

pkg/helper/webconsole.go
myeung18: modified

pkg/apis/addtoscheme_apps_v1alpha1.go
myeung18: yes, agreed. The code moved to main.go.

eguzki:
Waiting for your input. Let us know.

pkg/helper/webconsole.go
myeung18: new method created

pkg/controller/apimanager/apimanager_controller.go
myeung18: updated

pkg/controller/apimanager/apimanager_controller.go
myeung18: I will move all the Route watch, ConsoleLink creation and deletion related codes to a new Controller: webconsole_controller.go. In there, I will figure out a new way for ConsoleLink creation/deletion instead of using the existence of apimanager to determine ConsoleLink's creation/deletion.  Does this sound right to go?

myeung18:
> Waiting for your input. Let us know

sorry of the delay, now I'm back.

pkg/controller/apimanager/apimanager_controller.go
myeung18: added a new controller

myeung18:
> Waiting for your input. Let us know.

ready for further review

myeung18:
@eguzki   I just did a rebase, built and tested in my lab, 3scale worked ok.
I do see those ci test failures, but they seem to appear on other PRs as well. 
Is there anything else I should on this PR?

miguelsorianod:
Hi @myeung18,

Yesterday we applied some changes to fix some issues that we had in our CI tests (https://github.com/3scale/3scale-operator/pull/480). Could you rebase again from master so the tests are run with the new changes?

Thank you.

pkg/helper/webconsole.go
```diff
@@ -0,0 +1,120 @@
...
+				logger.Info("Console link has been created:", consoleLinkName)
+			}
+		} else if err == nil && consoleLink != nil {
+			updateConsoleLink(ctx, route, consoleLink, c)
+		}
+	}
+}
+func getRoute(ctx context.Context, c client.Client, req *reconcile.Request) *routev1.Route {
+	// here the loop tries the best to collect a route, the reconcile event may comes
+	// before the route resources get created/updated
```
miguelsorianod, line 57: I'm not sure why this loop is needed.

Why would the reconcile event come before the route gets created/updated?
In the controller code there's a watch for Route resources. An event will be triggered when a route is created/updated/deleted

pkg/helper/webconsole.go
```diff
@@ -0,0 +1,120 @@
...
+	reader := read.New(c).WithNamespace(req.Namespace)
+	for i := 1; i < 30; i++ {
+		time.Sleep(time.Duration(100) * time.Microsecond)
+		deployedRoutes, err := reader.List(&routev1.RouteList{})
+		if err == nil {
+			for _, route := range deployedRoutes { //look for 3scale admin routes, capture route object
```
miguelsorianod, line 63: Outdated comment

pkg/helper/webconsole.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+	}
+}
+
+func getConsoleLink(consoleLinkName string, route *routev1.Route, req *reconcile.Request) *consolev1.ConsoleLink {
+	return &consolev1.ConsoleLink{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: consoleLinkName,
+			Labels: map[string]string{
+				"3scale.net/name": req.Name,
```
miguelsorianod, line 91: Would `3scale.net/route-name` make more sense than `3scale.net/name`?

pkg/helper/webconsole.go
```diff
@@ -0,0 +1,120 @@
...
+	if url != consoleLink.Spec.Href || linkTxt != consoleLink.Spec.Text {
+		consoleLink.Spec.Href = url
+		consoleLink.Spec.Text = linkTxt
+		if err := c.Update(ctx, consoleLink); err != nil {
+			logger.Error(err, "failed to reconcile Console Link", consoleLink)
+		}
+	}
+}
+
+func getConsoleLink(consoleLinkName string, route *routev1.Route, req *reconcile.Request) *consolev1.ConsoleLink {
```
miguelsorianod, line 86: Could I get a captured picture of the ConsoleLink in the UI to see how it looks once created?

pkg/helper/webconsole.go
```diff
@@ -0,0 +1,120 @@
...
+// ReconcileConsoleLink creates a ConsoleLink object if it doesn't already exist
+func ReconcileConsoleLink(ctx context.Context, c client.Client, req *reconcile.Request) {
+	route := getRoute(ctx, c, req)
+	if route == nil {
+		removeConsoleLink(ctx, c, req)
+	} else {
+		consoleLinkName := fmt.Sprintf("%s-%s", ConsoleLinkNamePrefix, req.Namespace)
+		consoleLink := &consolev1.ConsoleLink{}
+		err := c.Get(ctx, types.NamespacedName{Name: consoleLinkName}, consoleLink)
+		if err != nil && apierrors.IsNotFound(err) {
```
miguelsorianod, line 43: What happens when error is nil and apierrors is not the IsNotFound error?
It seems we ignore the error in that case. I think when that happens we should log the error and perform a requeue (with a delay if you prefer)

myeung18: - Then if the link exists, it will be updated.
- added logging message.  `else` part

miguelsorianod: For what I see in the current implementation if there's an error we do not do anything. So if there's some temporary error when doing a create/update/delete the action would be simply lost.
I think we should retry in those cases instead of just logging because otherwise the ConsoleLink would never be created/reconciled (unless there's an external event on the route that triggers the reconciliation loop again)

pkg/helper/webconsole.go
myeung18: 3sale_link

myeung18:
> @eguzki I just did a rebase, built and tested in my lab, 3scale worked ok.
> I do see those ci test failures, but they seem to appear on other PRs as well.
> Is there anything else I should on this PR?

pkg/helper/webconsole.go
myeung18: modified

pkg/helper/webconsole.go
myeung18: changed

pkg/helper/webconsole.go
myeung18: I can't reproduce the case I saw before.  
I removed the loop.

eguzki:
I just hit re-run. If it keeps failing, it needs to be debugged. The issue is that 3scale did not deploy as expected. Maybe the expectation is what needs to be worked out. We will check.

pkg/helper/webconsole.go
myeung18: if the error situation is not temporary, and it persists, will that cause nonstop requeue? Do we need additional logic prevent this from happening?

myeung18:
> Hi @myeung18,
> 
> Yesterday we applied some changes to fix some issues that we had in our CI tests (#480). Could you rebase again from master so the tests are run with the new changes?
> 
> Thank you.

rebased, all changes are ready for review

pkg/helper/webconsole.go
```diff
@@ -0,0 +1,121 @@
...
+			consoleLink = getConsoleLink(consoleLinkName, route, req)
+			if err := c.Create(ctx, consoleLink); err != nil {
```
miguelsorianod, line 45: If creation fails we should also retry apart from logging

pkg/helper/webconsole.go
```diff
@@ -0,0 +1,121 @@
...
+				logger.Error(err, "Console link is not created.")
+			} else {
+				logger.Info("Console link has been created:", consoleLinkName)
+			}
+		} else if err == nil && consoleLink != nil {
+			updateConsoleLink(ctx, route, consoleLink, c)
```
miguelsorianod, line 51: If update fails we should also retry apart from logging

pkg/helper/webconsole.go
miguelsorianod: It will cause a nonstop requeue indeed. A delay requeue can be added if desired(with RequeueAfter), and in any case when lots of events are sent Kubernetes performs aggregation of them and also applies an exponential backoff retry time when it detects lots of them being sent in a rapid succession

pkg/helper/webconsole.go
```diff
@@ -0,0 +1,121 @@
...
+func ConsoleLinkSupported() error {
+	gvk := schema.GroupVersionKind{Group: "console.openshift.io", Version: "v1", Kind: "ConsoleLink"}
+	return kubernetes.CustomResourceDefinitionExists(gvk)
+}
+
+// ReconcileConsoleLink creates a ConsoleLink object if it doesn't already exist
+func ReconcileConsoleLink(ctx context.Context, c client.Client, req *reconcile.Request) error {
+	route := getRoute(ctx, c, req)
+	if route == nil {
+		removeConsoleLink(ctx, c, req)
```
miguelsorianod, line 38: Here if it fails to remove we should also requeue. There's an exception in this case: If it fails to remove because it does not exist anymore then we don't need to requeue

pkg/helper/webconsole.go
miguelsorianod: Thanks for the capture :+1: . I see that it is an entry at project-level indeed.
I'm wondering whether we should be more specific in  displayed name of the link to make it clear that the url is the master portal url (as there are other urls: admin portal, developer portal, ...).

For example... Maybe:
"APIManager - 3scale Master Portal"
or
"APIManager - 3scale System Master Portal"
or
"APIManager - 3scale System Master"

What do you think @eguzki?

pkg/helper/webconsole.go
miguelsorianod: :+1:

pkg/helper/webconsole.go
miguelsorianod: :+1:

pkg/helper/webconsole.go
miguelsorianod: :+1:

pkg/helper/webconsole.go
eguzki: `APIManager - 3scale Master Portal` seems to be a better option.

eguzki:
Good, all checks passed.

pkg/helper/webconsole.go
myeung18: added requeue

pkg/helper/webconsole.go
myeung18: added requeue

pkg/helper/webconsole.go
myeung18: added requeue

pkg/helper/webconsole.go
myeung18: I made a manual test which makes the deletion will never succeed. I dont see any backoff. I need to kill the app and operator to stop the events. I think this should never happen in normal situation, just for your info..

pkg/helper/webconsole.go
myeung18: changed

pkg/controller/webconsole/webconsole_controller.go
```diff
@@ -0,0 +1,94 @@
...
+	*reconcilers.BaseReconciler
+}
+
+//Reconcile reads the state of the Routes and makes changes to the corresponding Consolelinks
+func (r *ReconcileWebConsole) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	logger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	logger.Info("Reconciling ReconcileWebConsole")
+
+	if err := helper.ConsoleLinkSupported(); err == nil {
+		if err = helper.ReconcileConsoleLink(r.Context(), r.Client(), &request); err != nil {
```
miguelsorianod, line 88: It would be nice to log the error here

pkg/helper/webconsole.go
```diff
@@ -0,0 +1,144 @@
...
+// ConsoleLinkSupported checks if a ConsoleLink CRD exists
+func ConsoleLinkSupported() error {
+	gvk := schema.GroupVersionKind{Group: "console.openshift.io", Version: "v1", Kind: "ConsoleLink"}
+	return kubernetes.CustomResourceDefinitionExists(gvk)
+}
+
+// ReconcileConsoleLink creates/deletes/updates a ConsoleLink based on the Route
+func ReconcileConsoleLink(ctx context.Context, c client.Client, req *reconcile.Request) error {
+	route := getRoute(ctx, c, req)
+	if route == nil {
```
miguelsorianod, line 37: Probably a good idea to check the error and return the error to retry if it fails (the exception is if the error is not found of course)

pkg/helper/webconsole.go
```diff
@@ -0,0 +1,144 @@
...
+	return nil
+}
+
+// removeConsoleLink removes a ConsoleLink object if it exists
+// error is returned if the deletion is failed.
+func removeConsoleLink(ctx context.Context, c client.Client, req *reconcile.Request) error {
+	consoleLink := &consolev1.ConsoleLink{}
+	consoleLinkName := fmt.Sprintf("%s-%s", ConsoleLinkNamePrefix, req.Namespace)
+	err := c.Get(ctx, types.NamespacedName{Name: consoleLinkName}, consoleLink)
+	if err == nil && consoleLink != nil {
```
miguelsorianod, line 96: This checking seems odd.

If err is nil ConsoleLink shouldn't be nil by definition right? :thinking:

miguelsorianod: Additionally, if the error is not nil and it is not the isnotfound error then the error should be returned so it can be retried

pkg/helper/webconsole.go
miguelsorianod: The events would never stop in that case. They should be incrementally taking longer to be received due to the exponential backoff when lots of them are received in a quick succession. Maybe you are not reproducing the exponential backoff because you've added a delay in the requeue that might be big enough to not be considered "event spamming" by the framework :thinking:

eguzki:
Created PR https://github.com/myeung18/3scale-operator/pull/1 for some refactoring

myeung18:
refactors are applied and tested.

miguelsorianod:
Looks good to me :+1:

miguelsorianod (approved):

eguzki:
Congratulations @myeung18 for the very good job. :medal_military: 

Thank you very much for the contribution

eguzki (approved):

PR #467, opened 2020-09-14
Resource pvc system storage and databases

eguzki:
Closes https://issues.redhat.com/browse/THREESCALE-3652

There is no way to define PVC size in the CR using the Operator deployment for any of the storage backends required by 3scale.

* system storage PVC

```
kind: APIManager
metadata:
  name: apimanager1
spec:
  wildcardDomain: example.com
  system:
    fileStorage:
      persistentVolumeClaim:
        resources: 
          requests: 2Gi
```

* MySQL PVC
```
kind: APIManager
metadata:
  name: apimanager1
spec:
  wildcardDomain: example.com
  system:
    database:
      mysql:
        persistentVolumeClaim:
          resources: 
            requests: 2Gi
```
* PostgreSQL PVC
```
kind: APIManager
metadata:
  name: apimanager1
spec:
  wildcardDomain: example.com
  system:
    database:
      postgresql:
        persistentVolumeClaim:
          resources: 
            requests: 2Gi
```

Note: Backend Redis PVC and System Redis PVC's not updated.


eguzki:
First obvious question, should we add resources and volumeName to redis PVC's (system and backend) as well?  They are supposed to be used only for backup and if correctly configured (rotation), should never be too big.

pkg/3scale/amp/operator/system_mysql_options_provider.go
```diff
@@ -159,11 +159,22 @@ func (s *SystemMysqlOptionsProvider) setResourceRequirementsOptions() {
...
 		s.apimanager.Spec.System.DatabaseSpec.MySQL.PersistentVolumeClaimSpec != nil {
+
 		s.mysqlOptions.PVCStorageClass = s.apimanager.Spec.System.DatabaseSpec.MySQL.PersistentVolumeClaimSpec.StorageClassName
+		volumeName = s.apimanager.Spec.System.DatabaseSpec.MySQL.PersistentVolumeClaimSpec.VolumeName
+		if s.apimanager.Spec.System.DatabaseSpec.MySQL.PersistentVolumeClaimSpec.Resources != nil {
+			storageRequests = s.apimanager.Spec.System.DatabaseSpec.MySQL.PersistentVolumeClaimSpec.Resources.Requests
+		}
 	}
+
+	s.mysqlOptions.PVCVolumeName = volumeName
```
miguelsorianod, line 176: Can we move this one inside the conditional? Just assigning the value it has there, without the need to create the auxiliar variable either.

pkg/3scale/amp/operator/system_postgresql_options_provider.go
```diff
@@ -145,11 +145,22 @@ func (s *SystemPostgresqlOptionsProvider) setResourceRequirementsOptions() {
...
 		s.apimanager.Spec.System.DatabaseSpec.PostgreSQL.PersistentVolumeClaimSpec != nil {
+
 		s.options.PVCStorageClass = s.apimanager.Spec.System.DatabaseSpec.PostgreSQL.PersistentVolumeClaimSpec.StorageClassName
+		volumeName = s.apimanager.Spec.System.DatabaseSpec.PostgreSQL.PersistentVolumeClaimSpec.VolumeName
+		if s.apimanager.Spec.System.DatabaseSpec.PostgreSQL.PersistentVolumeClaimSpec.Resources != nil {
+			storageRequests = s.apimanager.Spec.System.DatabaseSpec.PostgreSQL.PersistentVolumeClaimSpec.Resources.Requests
+		}
 	}
+
+	s.options.PVCVolumeName = volumeName
```
miguelsorianod, line 162: Same comment as https://github.com/3scale/3scale-operator/pull/467/files?file-filters%5B%5D=.go&file-filters%5B%5D=.md#r490158143

pkg/apis/apps/v1alpha1/apimanager_types.go
```diff
@@ -342,6 +342,13 @@ type SystemRedisPersistentVolumeClaimSpec struct {
 type SystemPVCSpec struct {
 	// +optional
 	StorageClassName *string `json:"storageClassName,omitempty"`
+	// Resources represents the minimum resources the volume should have.
+	// Ignored when VolumeName field is set
```
miguelsorianod, line 346: About this...

Unfortunately, although Resources is ignored when VolumeName field is set, it seems that it is still mandatory at the PVC level even if you define VolumeName in the PVC
Fortunately I think in this implementation when StorageClassName is not set we still set defaults at PVC level right? If that's the case then this would be correct I think

doc/apimanager-reference.md
```diff
@@ -152,6 +152,8 @@ Only one of the fields can be chosen. If no field is specified then PVC is used.
 | **Field** | **json/yaml field**| **Type** | **Required** | **Default value** | **Description** |
 | --- | --- | --- | --- | --- | --- |
 | StorageClassName | `storageClassName` | string | No | nil | The Storage Class to be used by the PVC |
+| Resources | `resources` | [PersistentVolumeClaimResourcesSpec](apimanagerbackup-reference.md#PersistentVolumeClaimResourcesSpec) | No | nil | The minimum resources the volume should have |
```
miguelsorianod, line 155: Referencing APIManagerBackup seems a little bit weird. Descriptions there are in the context of backup too. For example in the linked url "Size of the PersistentVolumeClaim where the backup is to be performed. Set enough size to contain all data that is backed up."

Can we duplicate the content here and adapt it? Or maybe linking to the K8s resources API type documentation and describing that is limited to some specific fields inside of it

miguelsorianod (changes requested):
I left some comments.

Implementation-wise looks correct to me.

I have some comments about offering this:

* Updates on the resources fields can only be done if the underlying storage used accepts
dynamic resizing. If we are going to offer this I think we should make this super clear (at doc level at least)
* I'm wondering whether for the case of DBs it is a good idea to offer the ability to set the size. By allowing that we are kind of encouraging people to do it when they have higher DB size needs, which might incentive users to use this when they need that for production environments, which is not what we want. In that case users should use external databases.

pkg/3scale/amp/operator/system_mysql_options_provider.go
eguzki: what is the benefit? The way it is  now, `s.mysqlOptions.PVCVolumeName` is always assigned. The way you propose, it is only assigned when the conditional is true. Both would work, assuming no other method modifies that variable, but for me it is much clearer like this.

pkg/apis/apps/v1alpha1/apimanager_types.go
eguzki: Indeed, resources are always set.

doc/apimanager-reference.md
eguzki: You mean to duplicate also the [type](https://github.com/3scale/3scale-operator/blob/ec9187a45d0de68f2c943e73940a09310bd450f3/pkg/apis/apps/v1alpha1/apimanagerbackup_types.go#L53) 
```
type PersistentVolumeClaimResources struct {
	// Storage Resource requests to be used on the PersistentVolumeClaim.
	// To learn more about resource requests see:
	// https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
	Requests resource.Quantity `json:"requests"` // Should this be a string or a resoure.Quantity? it seems it is serialized as a string
}
```

Or maybe take it to some `commons` and reuse from any CRD.

pkg/3scale/amp/operator/system_mysql_options_provider.go
miguelsorianod: To me it is more clear and less verbose the other way around. No need to define another variable where the zero-value does it already, as it's often done in Go. You only set it in case MySQL PVC section is defined, as with all the other parameters.

In any case, as you prefer, it's a matter of style.

doc/apimanager-reference.md
miguelsorianod: I meant at documentation level only

pkg/3scale/amp/operator/system_mysql_options_provider.go
eguzki: Let it be like this, then. This method `setPersistentVolumeClaimOptions` will decide the final value of `s.mysqlOptions.PVCVolumeName` without assuming its current value is zero value (nil).

eguzki:
Regarding this comment:
> I'm wondering whether for the case of DBs it is a good idea to offer the ability to set the size. By allowing that we are kind of encouraging people to do it when they have higher DB size needs, which might incentive users to use this when they need that for production environments, which is not what we want. In that case users should use external databases.

It has been requested by a customer case: https://issues.redhat.com/browse/THREESCALE-3652. Adding features does not mean to encourage to use them for all scenarios and cases. In that sense, I think the 3scale's message for the customers is clear: for production ready deployment, use external databases.

pkg/apis/apps/v1alpha1/apimanager_types.go
miguelsorianod: :+1:

doc/apimanager-reference.md

miguelsorianod:
After discussion about this, let's go ahead with the changes, making very clear the implications of them in our documentation.

Can you add into the documentation the following information?
* Make clear that the parameters are not updateable when the underlying assigned PersistentVolume does not support resizing (in reference and maybe in the user guide)
* Make clear that for production environment external databases should be used and users should not rely on using internal DBs with increased PVC sizes

doc/apimanager-reference.md
```diff
@@ -259,6 +264,8 @@ that should be set on it.
 | **Field** | **json/yaml field**| **Type** | **Required** | **Default value** | **Description** |
 | --- | --- | --- | --- | --- | --- |
 | StorageClassName | `storageClassName` | string | No | nil | The Storage Class to be used by the PVC |
+| Resources | `resources` | [PersistentVolumeClaimResourcesSpec](#PersistentVolumeClaimResourcesSpec) | No | nil | The minimum resources the volume should have |
```
miguelsorianod, line 267: Can we clarify that this element does not really take effect when VolumeName is provided?

doc/apimanager-reference.md
```diff
@@ -259,6 +264,8 @@ that should be set on it.
...
+| VolumeName | `volumeName` | string | No | nil | The binding reference to the PersistentVolume backing this claim |
```
miguelsorianod, line 268: Can we clarify that this is used to provide a pre-existing PV instead of letting the operator create its own one?

doc/apimanager-reference.md
```diff
@@ -358,6 +365,12 @@ an `APIManager` status field should never be modified by the user.
 | --- | --- | --- | --- |
 | No fields for the moment | | | |
 
+## PersistentVolumeClaimResourcesSpec
+
+| **json/yaml field**| **Type** | **Required** | **Default value** | **Description** |
+| --- | --- | --- | --- | --- |
+| `requests` | [v1 Quantity](https://v1-17.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#quantity-resource-core) | Yes | N/A | Size of the PersistentVolumeClaim where the backup is to be performed. Set enough size to contain all [data that is backed up](#data-that-is-backed-up).
```
miguelsorianod, line 372: The description here should be slightly changed as it does not deal with the backup in this case

doc/apimanager-reference.md
```diff
@@ -198,6 +199,8 @@ Only one of the fields can be chosen. If no field is specified then PVC is used.
 | **Field** | **json/yaml field**| **Type** | **Required** | **Default value** | **Description** |
 | --- | --- | --- | --- | --- | --- |
 | StorageClassName | `storageClassName` | string | No | nil | The Storage Class to be used by the PVC |
+| Resources | `resources` | [PersistentVolumeClaimResourcesSpec](#PersistentVolumeClaimResourcesSpec) | No | nil | The minimum resources the volume should have. Resources will not take any effect when VolumeName is provided. |
```
miguelsorianod, line 202: :+1: about not taking effect when VolumeName is provided.

Could we document here also that the parameter is not updateable when the underlying PV is not resizable? (see comment https://github.com/3scale/3scale-operator/pull/467#issuecomment-700603059)

Can we also add somewhere (maybe in the highAvailability attribute) the information about internal DBs not being intended for production environments?

miguelsorianod (approved):

PR #474, opened 2020-09-23
product controller: fix backend index build

### LINKED ISSUE #472: Failed on multiple backend declaration
Cannot declare multiple backendUsages in Product definition as the next doc: https://github.com/3scale/3scale-operator/blob/master/doc/operator-capabilities.md#product-backend-usages

When i declare 2 or more backends in the same Product, it failed with:

reconcile product spec: Task failed SyncBackendUsage: Error sync product [backendX] backendusages: product [backendX] update backendusage: error calling 3scale system - reason: {"backend_api_id":["has already been taken"]} - code: 422

it looks like try to asign the same ID to all the backends in backendUsages

eguzki:
Fixes #472

### DIFF
--- a/pkg/controller/helper/backendapi_remote_index.go
+++ b/pkg/controller/helper/backendapi_remote_index.go
@@ -22,8 +22,8 @@ func NewBackendAPIRemoteIndex(client *threescaleapi.ThreeScaleClient, logger log
 	backendIDIndex := map[int64]*BackendAPIEntity{}
 	backendSystemNameIndex := map[string]*BackendAPIEntity{}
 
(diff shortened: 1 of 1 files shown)


miguelsorianod (approved):

miguelsorianod (approved):
I understand the issue was that a copy of the backend was being sent to NewBackendAPIEntity, and that class performs some kind of modification of Backend's attributes, so it was modifying the copy.
By sending a reference to the original Backend the change to the attribute is effective outside.

eguzki:
Never use pointers to loop variables as I did.

PR #486, opened 2020-10-02
reconcile system database secret when external databases enabled

eguzki:
When external databases are enabled, system database secret is not reconciled, hence ower reference is not set.

The implementation required some refactoring:

* Backend redis secret is reconciled in redis_reconciled.go and highavailability_reconciler.go. It has been removed from backend_reconciled.go
* System redis secret is reconciled in redis_reconciled.go and highavailability_reconciler.go. It has been removed from system_reconciled.go
* highavailabiilty_reconciler.go now reconciles three secrets: system redis, backend redis and system database secrets..

*TODO*
- [x] new HA reconciler unittests
- [x] redis option provider unittests update with new redis params

### DIFF
--- a/pkg/3scale/amp/component/highavailability_options.go
+++ b/pkg/3scale/amp/component/highavailability_options.go
@@ -6,7 +6,7 @@ type HighAvailabilityOptions struct {
 	BackendRedisQueuesEndpoint          string `validate:"required"`
 	BackendRedisQueuesSentinelHosts     string
 	BackendRedisQueuesSentinelRole      string
-	BackendRedisStorageEndpoint         string
+	BackendRedisStorageEndpoint         string `validate:"required"`
 	BackendRedisStorageSentinelHosts    string
 	BackendRedisStorageSentinelRole     string
 	SystemDatabaseURL                   string `validate:"required"`
--- a/pkg/3scale/amp/component/highavailability_options.go
+++ b/pkg/3scale/amp/component/highavailability_options.go
@@ -3,20 +3,25 @@ package component
 import "github.com/go-playground/validator/v10"
 
 type HighAvailabilityOptions struct {
-	AppLabel                            string `validate:"required"`
 	BackendRedisQueuesEndpoint          string `validate:"required"`
-	BackendRedisQueuesSentinelHosts     string `validate:"required"`
-	BackendRedisQueuesSentinelRole      string `validate:"required"`
-	BackendRedisStorageEndpoint         string `validate:"required"`
-	BackendRedisStorageSentinelHosts    string `validate:"required"`
-	BackendRedisStorageSentinelRole     string `validate:"required"`
+	BackendRedisQueuesSentinelHosts     string
+	BackendRedisQueuesSentinelRole      string
+	BackendRedisStorageEndpoint         string
+	BackendRedisStorageSentinelHosts    string
+	BackendRedisStorageSentinelRole     string
 	SystemDatabaseURL                   string `validate:"required"`
 	SystemRedisURL                      string `validate:"required"`
-	SystemRedisSentinelsHosts           string `validate:"required"`
-	SystemRedisSentinelsRole            string `validate:"required"`
+	SystemRedisSentinelsHosts           string
+	SystemRedisSentinelsRole            string
+	SystemRedisNamespace                string
 	SystemMessageBusRedisURL            string `validate:"required"`
-	SystemMessageBusRedisSentinelsHosts string `validate:"required"`
-	SystemMessageBusRedisSentinelsRole  string `validate:"required"`
+	SystemMessageBusRedisSentinelsHosts string
+	SystemMessageBusRedisSentinelsRole  string
+	SystemMessageBusRedisNamespace      string
+
+	BackendRedisLabels   map[string]string `validate:"required"`
+	SystemRedisLabels    map[string]string `validate:"required"`
+	SystemDatabaseLabels map[string]string `validate:"required"`
 }
 
 func NewHighAvailabilityOptions() *HighAvailabilityOptions {
--- a/pkg/3scale/amp/operator/highavailability_options_provider.go
+++ b/pkg/3scale/amp/operator/highavailability_options_provider.go
@@ -190,7 +190,7 @@ func (h *HighAvailabilityOptionsProvider) setSystemRedisOptions() error {
 	}
 
 	for _, option := range casesWithDefault {
-		val, err := h.secretSource.FieldValue(option.secretName, option.secretField, option.defValue)
+		val, err := h.secretSource.FieldValueFromRequiredSecret(option.secretName, option.secretField, option.defValue)
 		if err != nil {
 			return err
 		}
--- a/pkg/3scale/amp/operator/highavailability_options_provider.go
+++ b/pkg/3scale/amp/operator/highavailability_options_provider.go
@@ -108,7 +108,7 @@ func (h *HighAvailabilityOptionsProvider) setBackendRedisOptions() error {
 	}
 
 	for _, option := range casesWithDefault {
-		val, err := h.secretSource.FieldValue(option.secretName, option.secretField, option.defValue)
+		val, err := h.secretSource.FieldValueFromRequiredSecret(option.secretName, option.secretField, option.defValue)
 		if err != nil {
 			return err
 		}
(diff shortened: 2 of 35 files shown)


pkg/3scale/amp/component/highavailability_options.go
```diff
@@ -3,20 +3,25 @@ package component
...
-	AppLabel                            string `validate:"required"`
 	BackendRedisQueuesEndpoint          string `validate:"required"`
-	BackendRedisQueuesSentinelHosts     string `validate:"required"`
-	BackendRedisQueuesSentinelRole      string `validate:"required"`
-	BackendRedisStorageEndpoint         string `validate:"required"`
-	BackendRedisStorageSentinelHosts    string `validate:"required"`
-	BackendRedisStorageSentinelRole     string `validate:"required"`
+	BackendRedisQueuesSentinelHosts     string
+	BackendRedisQueuesSentinelRole      string
+	BackendRedisStorageEndpoint         string
```
miguelsorianod, line 9: I think `BackendRedisStorageEndpoint` might be better to be required because you always have to provide it

pkg/3scale/amp/operator/highavailability_options_provider.go
```diff
@@ -106,6 +144,61 @@ func (h *HighAvailabilityOptionsProvider) setSystemRedisOptions() error {
...
+		{
+			&h.options.SystemMessageBusRedisNamespace,
+			component.SystemSecretSystemRedisSecretName,
+			component.SystemSecretSystemRedisMessageBusRedisNamespace,
+			component.DefaultSystemMessageBusRedisNamespace(),
+		},
+	}
+
+	for _, option := range casesWithDefault {
+		val, err := h.secretSource.FieldValue(option.secretName, option.secretField, option.defValue)
```
miguelsorianod, line 193: Shouldn't it be `FieldValueFromRequiredSecret` ?

pkg/3scale/amp/operator/highavailability_options_provider.go
```diff
@@ -77,6 +74,47 @@ func (h *HighAvailabilityOptionsProvider) setBackendRedisOptions() error {
...
+		{
+			&h.options.BackendRedisQueuesSentinelRole,
+			component.BackendSecretBackendRedisSecretName,
+			component.BackendSecretBackendRedisQueuesSentinelRoleFieldName,
+			component.DefaultBackendQueuesSentinelRole(),
+		},
+	}
+
+	for _, option := range casesWithDefault {
+		val, err := h.secretSource.FieldValue(option.secretName, option.secretField, option.defValue)
```
miguelsorianod, line 111: Shouldn't it be `FieldValueFromRequiredSecret` ?

pkg/3scale/amp/operator/redis_options_provider.go
```diff
@@ -51,13 +58,127 @@ func (r *RedisOptionsProvider) GetRedisOptions() (*component.RedisOptions, error
 
 	r.setPersistentVolumeClaimOptions()
 
-	err := r.options.Validate()
+	// Should the operator be reading redis secrets?
```
miguelsorianod, line 61: I think so.

Even if the DBs are deployed by the operator we allow the users to precreate the secret to allow non-default values (For example, using a specific password that the user wants to define instead of an auto-generated one)

pkg/3scale/amp/operator/redis_options_provider.go
```diff
@@ -51,13 +58,127 @@ func (r *RedisOptionsProvider) GetRedisOptions() (*component.RedisOptions, error
...
+	// When HA is disabled, do we support external redis?
```
miguelsorianod, line 62: We do not support external redis when HA (external mode) is disabled.
Users could precreate a secret in that case with urls pointing to external DBs but it's not a supported scenario. Should we document this edge-case? (until now we are assuming the user is responsible enough to not do that when internal dbs are deployed)

pkg/3scale/amp/operator/redis_options_provider.go
```diff
@@ -51,13 +58,127 @@ func (r *RedisOptionsProvider) GetRedisOptions() (*component.RedisOptions, error
...
+	// If answer is true, why does the operator deploy redis?
+	// If the answer is no, then it would be sufficient to set default URL's (internal redis url)
```
miguelsorianod, line 64: See my comment https://github.com/3scale/3scale-operator/pull/486/files#r500176502.

Even if the answer is no I think we should reconcile other fields to allow non-default values for fields of the secret.
This applies to system redis and to system database too.

pkg/3scale/amp/operator/redis_options_provider.go
```diff
@@ -51,13 +58,127 @@ func (r *RedisOptionsProvider) GetRedisOptions() (*component.RedisOptions, error
...
+		{
+			&r.options.SystemMessageBusRedisNamespace,
+			component.SystemSecretSystemRedisSecretName,
+			component.SystemSecretSystemRedisMessageBusRedisNamespace,
+			component.DefaultSystemMessageBusRedisNamespace(),
+		},
+	}
+
+	for _, option := range cases {
+		val, err := r.secretSource.FieldValue(option.secretName, option.secretField, option.defValue)
```
miguelsorianod, line 172: Same comment as https://github.com/3scale/3scale-operator/pull/486/files#r500173988

miguelsorianod (changes requested):
Design looks good to me :+1: 

I left some comments

pkg/3scale/amp/operator/redis_options_provider.go
eguzki: But then we are allowing the users to point to external redis instance (both backend and system) or even broken (non existing) redis instance, hence breaking the deployment. 

If the operator deploys redis internally, it should make sure backend and system are configured to use that redis. For that, "desired" url for the secret should be internal url. In case the customer accidentally updates that secret, the operator would reconcile to the internal redis, making the system more robust.

Are we going to allow custom password, allowing also broken deployment, when redis is deployed internally?

pkg/3scale/amp/operator/redis_options_provider.go
eguzki: My reasoning also applies for system database.

pkg/3scale/amp/operator/redis_options_provider.go
eguzki: When redis is deployed internally, sentinel configuration and namespace conf does not have any sense. 

This change may belong to another PR, but I have my doubts about the current implementation, hence bringing up the discussion.

pkg/3scale/amp/operator/redis_options_provider.go
miguelsorianod: When redis is deployed internally using a custom pre-creatd password should not break the deployent.

pkg/3scale/amp/operator/redis_options_provider.go
eguzki: This is also related to the previous conv about the necessity of reading the secret. 

In this case, we are reconciling the secret and it may not exist (in the beginning). If not existing, default values are read and the secret will be created.

pkg/3scale/amp/operator/redis_options_provider.go
miguelsorianod: As I commented in https://github.com/3scale/3scale-operator/pull/486#discussion_r500177300, until now we are assuming the user is responisble enough to not configure parameters that it shouldn't (like pointing to another DB when redis are internal), as we are doing with other configurations like resource requirements (we assume they place values that are enough), affinity/tolerations (we assume we configure them in compatible nodes architectures), and so on.

pkg/3scale/amp/operator/redis_options_provider.go
miguelsorianod: sentinel configuration does not make sense indeed.
namespace conf I think be configured without issues.
Some secret attributes make sense when redis DBs are internal and some don't as commented in https://github.com/3scale/3scale-operator/pull/486#discussion_r500177300

pkg/3scale/amp/operator/redis_options_provider.go
miguelsorianod: You are right.

In this case redis options provider is used only for internal redis so the secret can be created automatically by the operator if it has not been pre-created by the user.

pkg/3scale/amp/operator/redis_options_provider.go
eguzki: leaving the comment as a reminder note for a future enhancement

miguelsorianod (approved):

PR #489, opened 2020-10-08
Allow Zync Database to be configured as an external database on the operator

miguelsorianod:
This PR intends to allow Zync Database to be configured externally.

Currently it does so by adding a field `.spec.highAvailability.externalZyncDatabaseEnabled`, whose value (`true` or `false`) only takes effect when `spec.highAvailability.enabled` is set to `true`. This implies that with this implementation the only way to set an external zync database is by enabling the other ones to be external too. We can discuss whether we want that to happen or be more granular, taking into account the implications.
The reason why behavior of the existing `spec.highAvailability.Enabled` has not been changed to include external zync database is because doing that would break current users deployments (thus why another different new optional parameter is needed)

When the zync database is configured externally:
* The `zync-database-postgresql` ImageStream is not created anymore
* The `zync-database` DeploymentConfig is not created anymore
* The `zync-database` Service is not created anymore

This is an operator-only functionality.

The possibility of splitting the internal zync database specific components into its own component has been evaluated but to simplify the number of changes needed and diminish the impact on templates side that approach has not been followed.

Let's iterate from here.

- [x] Add unit tests
- [x] Add documentation


eguzki:
Looking good so far

miguelsorianod:
Added unit tests and documentation.

Ready for a review @eguzki

eguzki (approved):

PR #496, opened 2020-10-19
Openapi CRD

### LINKED ISSUE #119: OpenAPI Support & Discovery
## OpenAPI Support

We should be able to relate one API with its OpenAPI Document, so we can create several objects and extract some useful information: 

* Base Path (Path routing?) : extracted from the annotation: `discovery.3scale.net/path: /v1/api`
* Metrics: Extracted from OperationIDs or, if empty, by joining the "Path+Verb".
* Mapping Rules: Based on path , http verb, Operation ID and an increment of 1.

Example path definition extracted from [Example OpenAPI Document](https://github.com/OAI/OpenAPI-Specification/blob/master/examples/v3.0/api-with-examples.yaml): 

```yaml
/v2:  <-- HTTP Path
    get:   <-- Verb
      operationId: getVersionDetailsv2 <-- Metric
      summary: Show API version details <-- Metric Description? 
```

## Options
Let's see some of the options, on how to relate the API object with the OpenAPI Document: 

### 1. Using the "Service Discovery Spec"

The operator should be able to detect if there is an "Openshift Service" with the same name as the value of `privateBaseURL`, then, read that "Openshift Service" and look for the label: `discovery.3scale.net: "true"`, if that label is set, then proceed to extract all the possible annotations defined in the "[Service Discovery Spec](https://docs.google.com/document/d/1QsunYWA9jy1VRjkIeWFUwEScSXsmaAYmuwEF1Zs5SoU/edit#)" as:

```
    discovery.3scale.net/scheme: http
    discovery.3scale.net/port: 80
    discovery.3scale.net/path: /v1/api
    discovery.3scale.net/description-path: http://example.svc:8080/doc/swagger.json
```

This option doesn't require any change to the OpenAPI objects. 

### 2. Extending the API Object with: `OpenAPIRef`

This proposal extends the actual API Object. Adding a new `OpenAPIRef` to the integration Methods that apply (Apicast, Istio integration):

```yaml
      openAPIRef:
           URL: http://example.svc:8080/doc/swagger.json
           ConfigMapRef:
             name: MyOpenAPIDocument
           Document: |
             openapi: "3.0.0"
             info:
               title: Simple API overview
               version: 2.0.0
             paths:
               /:
                 get:
                   operationId: listVersionsv2
                   summary: List API versions
                   responses:
                       '200':
                           description: |-
                                200 response
                          content:
                              application/json:
                  (..)
```

This supports:
* **URL**: A remote URL from where to fetch the OpenAPI Document
* **ConfigMapRef**: A k8s ObjectReference that points to a ConfigMap that contains the OpenAPI Document (defaulting to openapi.yaml or openapi.json)
* **Document**: Inline OpenAPI Document.

Giving the user several ways of specifying the source of the OpenAPI Document.

Example: 
```yaml
kind: API
metadata:
  labels:
    environment: staging
  name: api01
spec:
  planSelector:
    matchLabels:
      api: api01
  description: api01
  integrationMethod:
    apicastOnPrem:
      apiTestGetRequest: /
      authenticationSettings:
        credentials:
          apiKey:
            authParameterName: user-key
            credentialsLocation: headers
        errors:
          authenticationFailed:
            contentType: text/plain; charset=us-ascii
            responseBody: Authentication failed
            responseCode: 403
          authenticationMissing:
            contentType: text/plain; charset=us-ascii
            responseBody: Authentication Missing
            responseCode: 403
        hostHeader: ""
        secretToken: MySecretTokenBetweenApicastAndMyBackend_1237120312
      mappingRulesSelector:
        matchLabels:
          api: api01
      privateBaseURL: https://echo-api.3scale.net:443
      productionPublicBaseURL: https://api.testing.com:443
      stagingPublicBaseURL: https://api.testing.com:443
      openAPIRef:
           Document: |
             openapi: "3.0.0"
             info:
               title: Simple API overview
               version: 2.0.0
             paths:
               /:
                 get:
                   operationId: listVersionsv2
                   summary: List API versions
                   responses:
                       '200':
                           description: |-
                                200 response
                          content:
                              application/json:
                  (..)
  metricSelector:
    matchLabels:
      api: api01
```

### 3. A combination of both approaches, allowing to override the discovered OpenAPI Doc.

This third option, combines previous options: discover if the target `privateBaseURL` is an openshift service, and if it contains the label  `discovery.3scale.net: "true"` check for the OpenAPI Document, but also check if the user has specified another OpenAPI Document by using the `openAPIRef` extension of the API object.

The user specified `openAPIRef` takes precedence over the auto-discovery of the `privateBaseURL` service, so `openAPIRef` **overrides** the values of the service annotations. 

### Proposal

I think the combination of both options is the best way to go as this gives the user more freedom on how to import the OpenAPI document.

We should implement option number 2 first and then, option number 1.

eguzki:
CRD (and controller) holding OAS3 reference (configmap or URL) as source of truth for 3scale Product and 3scale Backend configuration.

TODO:
- [x] Doc
- [ ] ~Unittests~
- [x] Custom private base URL
- [x] Openapi Status with relevant info
- [x] Prefix matching option
- [x] private api security options
- [x] OAS from URL

Fixes #119


eguzki:
Ready for initial review. 

Missing only small things and unittests

eguzki:
Unittests left as tech debt as coming soon upgrade of the operator-sdk version would require lots of refactors

deploy/crds/capabilities.3scale.net_openapis_crd.yaml
```diff
@@ -0,0 +1,204 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: openapis.capabilities.3scale.net
+spec:
+  group: capabilities.3scale.net
+  names:
+    kind: Openapi
```
miguelsorianod, line 8: I'd prefer naming our kind OpenAPI to match with the conventions with known accronyms and what we are doing with other kinds (`APIManager`, APIcast,...).

miguelsorianod: If we change this keep in mind that the struct names (Go data types) should be changed too

pkg/apis/capabilities/v1beta1/openapi_types.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+// OpenapiSpec defines the desired state of Openapi
+type OpenapiSpec struct {
+	// OpenAPIRef Reference to the OpenAPI Specification
+	OpenAPIRef OpenAPIRefSpec `json:"openapiRef"`
+
+	// ProviderAccountRef references account provider credentials
+	// +optional
+	ProviderAccountRef *corev1.LocalObjectReference `json:"providerAccountRef,omitempty"`
```
miguelsorianod, line 55: Are this other fields defined because are not supported natively in OpenAPI spec?

pkg/apis/capabilities/v1beta1/openapi_types.go
```diff
@@ -0,0 +1,193 @@
...
+	// OpenapiFailedConditionType indicates that an error occurred during reconcilliation.
+	// The operator will retry.
+	OpenapiFailedConditionType common.ConditionType = "Failed"
+)
+
+// OpenAPIRefSpec Reference to the OpenAPI Specification
+type OpenAPIRefSpec struct {
+	// ConfigMapRef ConfigMap that contains the OpenAPI Document
+	// +optional
+	ConfigMapRef *corev1.ObjectReference `json:"configMapRef,omitempty"`
```
miguelsorianod, line 40: I wonder whether we should be using `ObjectReference`. Wouldn't a `LocalObjectReference` be enough?
`ObjectReference` has lots of fields that depend on the user creating them and forces us to unintuitive changes and checkings throughout our code (see setDefaults methods and potentially other checkings in the code)

pkg/controller/openapi/backend_reconciler.go
```diff
@@ -0,0 +1,206 @@
...
+	"github.com/google/go-cmp/cmp"
+
+	"github.com/getkin/kin-openapi/openapi3"
+	"github.com/go-logr/logr"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+type BackendReconciler struct {
```
miguelsorianod, line 24: When we move to the new operator-sdk format we should rename this. The reason for that is that under the new format all controlers under a given API Group live in the same directory, which means the variable namespace is shared (see https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/project-v3-multigroup/controllers/ship for an example)

This means that `BackendReconciler` type would be in the same package as the other controllers and that might be confusing when the Backend controller is in the same place. Can we go ahead of that so we don't have to rename it during the migration?

pkg/controller/openapi/product_reconciler.go
```diff
@@ -0,0 +1,360 @@
...
+	"k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+var (
+	// LastSlashRegexp matches the last slash
+	LastSlashRegexp = regexp.MustCompile(`/$`)
+)
+
+type ProductReconciler struct {
```
miguelsorianod, line 30: Same comment as https://github.com/3scale/3scale-operator/pull/496/files?file-filters%5B%5D=.dia&file-filters%5B%5D=.go&file-filters%5B%5D=.png&file-filters%5B%5D=.yml#r513648085

pkg/controller/openapi/status_reconciler.go
```diff
@@ -0,0 +1,191 @@
...
+	"github.com/3scale/3scale-operator/pkg/reconcilers"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type StatusReconciler struct {
```
miguelsorianod, line 18: Same comment as https://github.com/3scale/3scale-operator/pull/496/files?file-filters%5B%5D=.dia&file-filters%5B%5D=.go&file-filters%5B%5D=.png&file-filters%5B%5D=.yml#r513648085

pkg/controller/openapi/product_reconciler.go
```diff
@@ -0,0 +1,360 @@
...
+		return p.openapiCR.Spec.ProductSystemName
+	}
+
+	return helper.SystemNameFromOpenAPITitle(p.openapiObj)
+}
+
+func (p *ProductReconciler) desiredObjName() string {
+	// DNS1123 Label compliant name. Due to UIDs are 36 characters of length this
+	// means that the maximum prefix lenght that can be provided is of 26
+	// characters. If the generated name is not DNS1123 compliant an error is
```
miguelsorianod, line 194: Is this comment obsolete? the `desiredObjName` does not return an error

pkg/apis/capabilities/v1beta1/openapi_types.go
eguzki: Yes, these other fields are 3scale specific configuration beyond OpenAPI spec. Taken from 3scale toolbox. 

However, 3scale product full configuration is not yet covered. Application plans and other settings are still missing.

pkg/apis/capabilities/v1beta1/openapi_types.go
eguzki: The idea was not to limit to the local namespace itself the location of the configmap. One customer may have a configmap in some other namespace and providing custom access to `3scale-operator` service account, the operator would be able to access config maps from other namespaces than local namespace. 

Maybe we should only define `Name` and `Namespace` attributes.

pkg/controller/openapi/product_reconciler.go
eguzki: the function does not return error, however it could return some name not compliant with DNS1123. 

The comments is there as a reminder that extra processing could be needed.

eguzki:
ready for new review @miguelsorianod

pkg/apis/capabilities/v1beta1/openapi_types.go
miguelsorianod: :+1:

pkg/apis/capabilities/v1beta1/openapi_types.go
```diff
@@ -0,0 +1,193 @@
...
+	OpenAPIRef OpenAPIRefSpec `json:"openapiRef"`
+
+	// ProviderAccountRef references account provider credentials
+	// +optional
+	ProviderAccountRef *corev1.LocalObjectReference `json:"providerAccountRef,omitempty"`
+
+	// ProductionPublicBaseURL Custom public production URL
+	// +kubebuilder:validation:Pattern=`^https?:\/\/.*$`
+	// +optional
+	ProductionPublicBaseURL string `json:"productionPublicBaseURL,omitempty"`
```
miguelsorianod, line 60: Is this parameter optional? If so, I think it might be better to define it as a pointer to string to follow Kubernetes conventions https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required.

I'm aware that in this case (and probably others) the zero-value is "forbidden" because it might imply it is unset itself, but they still insist on using pointers (see some of the arguments in the API conventions guide for that too).
I also think would also be more consistent with what we are doing in other places.

Same would apply to other attributes I've seen of this CR.

doc/operator-application-capabilities.md
```diff
@@ -629,6 +636,116 @@ oc create secret generic threescale-provider-account --from-literal=adminURL=htt
...
+### Features
+
+* [OpenAPI 3.0.2](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md) specification
+  * Just a single top level security requirement supported. Operation level security requirements not supported.
+  * Only first `server.url` element in servers list parsed as private url. Path and operation level `servers` elements not supported.
+  * Supported security schemes: `apiKey`.
+* OpenAPI spec document can be read from:
+  * ConfigMap
+  * URL. Supported schemes are http and https
+* When the `spec.productionPublicBaseURL` or the `spec.stagingPublicBaseURL` (or both) fields are provided, implicitly the customer is asking for "APIcast self-managed" deploment mode. Otherwise, default deployment mode will be set, that is, "APIcast 3scale managed".
```
miguelsorianod, line 650: typo : `deploment mode` -> `deployment mode`

pkg/apis/capabilities/v1beta1/openapi_types.go
```diff
@@ -0,0 +1,193 @@
...
+	// +optional
+	PrefixMatching bool `json:"prefixMatching,omitempty"`
+
+	// PrivateAPIHostHeader Custom host header sent by the API gateway to the private API
+	// +optional
+	PrivateAPIHostHeader string `json:"privateAPIHostHeader,omitempty"`
+
+	// PrivateAPISecretToken Custom secret token sent by the API gateway to the private API
+	// +optional
+	PrivateAPISecretToken string `json:"privateAPISecretToken,omitempty"`
```
miguelsorianod, line 85: Shouldn't this be a secret reference?

doc/operator-application-capabilities.md
```diff
@@ -629,6 +636,116 @@ oc create secret generic threescale-provider-account --from-literal=adminURL=htt
...
+## OpenAPI custom resource
+
+### Features
+
+* [OpenAPI 3.0.2](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md) specification
+  * Just a single top level security requirement supported. Operation level security requirements not supported.
+  * Only first `server.url` element in servers list parsed as private url. Path and operation level `servers` elements not supported.
+  * Supported security schemes: `apiKey`.
+* OpenAPI spec document can be read from:
+  * ConfigMap
```
miguelsorianod, line 648: I was wondering...

Can the OpenAPI spec contain some kind of sensitive information? If so, we should use a Secret instead of a ConfigMap. If not, a ConfigMap is already ok.

doc/operator-application-capabilities.md
```diff
@@ -629,6 +636,116 @@ oc create secret generic threescale-provider-account --from-literal=adminURL=htt
...
+  * URL. Supported schemes are http and https
```
miguelsorianod, line 649: Can we document accepted formats of the openapi spec (for both url and configmap)? Is it json, yaml, or both?
(same info in openapi crd doc would be useful too)

deploy/crds/capabilities.3scale.net_v1beta1_openapi_cr_url.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: capabilities.3scale.net/v1beta1
+kind: OpenAPI
+metadata:
+  name: example-openapi
```
miguelsorianod, line 4: Can we change the name of this sample or the configmap one?
The reason I'm asking this is due to in the new operator-sdk version you cannot have two samples with the same name

pkg/apis/capabilities/v1beta1/openapi_types.go
eguzki: This is a 3scale token, so called "Secret Token". Nothing to do with k8s secrets
Screenshot_2020-10-29 Services - Settings Red Hat 3scale API Management

pkg/helper/openapi_utils.go
```diff
@@ -0,0 +1,173 @@
...
+func SystemNameFromOpenAPITitle(obj *openapi3.Swagger) string {
+	openapiTitle := obj.Info.Title
+	openapiTitleToLower := strings.ToLower(openapiTitle)
+	return NonWordCharRegexp.ReplaceAllString(openapiTitleToLower, "_")
+}
+
+func K8sNameFromOpenAPITitle(obj *openapi3.Swagger) string {
+	openapiTitle := obj.Info.Title
+	openapiTitleToLower := strings.ToLower(openapiTitle)
+	return NonAlphanumRegexp.ReplaceAllString(openapiTitleToLower, "")
```
miguelsorianod, line 34: I'm probably nitpicking here but if K8s names have character length limitation. Maybe we should document somewhere that the title in the openapi spec has to have a max length.
K8s max length seems to be 253 characters which is quite high (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/) but the names often end up automatically being part of a label for some objects, in which case they effectively have a maximum length of 63 characters which is easily reachable.
I found myself in that situation when working on backup/restore. I don't how likely is that to happen here though so letting it up to you.

pkg/apis/capabilities/v1beta1/openapi_types.go
miguelsorianod: But is that token sensitive information? If so, it should be in a secret.

doc/operator-application-capabilities.md
eguzki: I thought about that too.

For the OpenID auth, 3scale required issuer data which contains some secret token. For the toolbox, the customers requested this issuer to be provided in the command line option instead of the OpenAPI spec, to keep OpenAPI public. So, I guess the customers want to keep OpenAPI spec public without any sensitive data. I do not know if this is the case for all cases. 

I will update to secret, just in case.

pkg/helper/openapi_utils.go
eguzki: That comment already exists in functions calling this function. Those functions complete the final k8s name and therefore the comment was done there.

miguelsorianod: Sorry, I meant on the user-facing side.
How likely is that it will reach this limit and thus an error creating objects will happen?

pkg/controller/openapi/openapi_controller.go
```diff
@@ -0,0 +1,399 @@
...
+
+	if len(openapiConfigMapObj.Data) < 1 {
+		fieldErrors = append(fieldErrors, field.Invalid(configMapRefFldPath, resource.Spec.OpenAPIRef.ConfigMapRef, "ConfigMap was empty"))
+		return nil, &helper.SpecFieldError{
+			ErrorType:      helper.InvalidError,
+			FieldErrorList: fieldErrors,
+		}
+	}
+
+	// Get arbitrary key value
```
miguelsorianod, line 302: Is the intended approach to get the first attribute found in the ConfigMap, independently of the attribute name?

pkg/controller/openapi/openapi_controller.go
```diff
@@ -0,0 +1,399 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to primary resource OpenAPI
+	err = c.Watch(&source.Kind{Type: &capabilitiesv1beta1.OpenAPI{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return err
+	}
+
```
miguelsorianod, line 88: We've added Backends and Products as owned objects of the API objects.
Should we receive events to re-reconcile them when changes happen to them and thus add watches here?

pkg/controller/openapi/openapi_product_reconciler.go
```diff
@@ -0,0 +1,360 @@
...
+	updated := helper.EnsureObjectMeta(existing, desired)
+
+	// OwnerRefenrence
+	updatedTmp, err := p.EnsureOwnerReference(p.openapiCR, existing)
+	if err != nil {
+		return false, err
+	}
+	updated = updated || updatedTmp
+
+	// Maybe too rough compare method?
```
miguelsorianod, line 167: That's indeed a tough question...

Let's leave it like this at the moment and see how people use it and what kind of interactions happen between controllers.

pkg/controller/openapi/openapi_status_reconciler.go
```diff
@@ -0,0 +1,191 @@
...
+
+	newStatus.ProviderAccountHost = s.providerAccountHost
+
+	productResourceName, err := s.getManagedProduct()
+	if err != nil {
+		return nil, err
+	}
+	newStatus.ProductResourceName = productResourceName
+
+	backendResourceNames, err := s.getManagedBackends()
```
miguelsorianod, line 88: Can multiple backends belong to a single OpenAPI resource?
looking at the OpenAPIBackendReconciler code it seems only a Backend is created :thinking:

pkg/controller/openapi/openapi_controller.go
```diff
@@ -0,0 +1,399 @@
...
+	productReconciler := NewOpenAPIProductReconciler(r.BaseReconciler, openapiCR, openapiObj, providerAccount, logger)
+	_, err = productReconciler.Reconcile()
+	if err != nil {
+		statusReconciler := NewOpenAPIStatusReconciler(r.BaseReconciler, openapiCR, providerAccount.AdminURLStr, err, false)
+		return statusReconciler, reconcile.Result{}, err
+	}
+
+	productSynced, err := r.checkProductSynced(openapiCR)
```
miguelsorianod, line 216: Why checkProductSynced is performed and not for backend?

miguelsorianod (changes requested):
I left some feedback and some questions to better understand the motivation of some implementation.

Looks good :+1:

pkg/apis/capabilities/v1beta1/openapi_types.go
eguzki: ok, let it be a secret.

eguzki: @miguelsorianod this field exists in `products.capabilities.3scale.net`. So it should also be updated to a secret. That would break backwards compat. I see your concern, but the requested change does not belong to this PR. 

I have opened issue for that: https://github.com/3scale/3scale-operator/issues/507

pkg/helper/openapi_utils.go
eguzki: The code verifies if it is valid DNS1123 Label. However, this is a k8s name. The validation is wrong. 

Can you elaborate more on your statement `but the names often end up automatically being part of a label for some objects`.

pkg/controller/openapi/openapi_controller.go
eguzki: Not the first one, but an arbitrary key. 

It is explained in the doc and reference doc. If the secret contains several fields (files), only one is read.

pkg/controller/openapi/openapi_controller.go
eguzki: They are already being watched by another controller of the same operator. It would need to be analyzed how the operator behaves.And it is not expected to be updated manually. I would not add watches now. But indeed, could be interesting.

pkg/controller/openapi/openapi_controller.go
eguzki: The product adds the backend as backend usage. The product will not be in sync until the backend usage is in sync. The product controller makes sure the backend usage's items are valid Backend CRs and are sync'ed.

eguzki: Adding the comment for future reference

pkg/apis/capabilities/v1beta1/openapi_types.go
miguelsorianod: Changing it in here would not break backwards compatibility no?
It would break backwards compatibility if we were to change it in `Product`, but not here.
We should be able to send it to `Product` as it expects without breaking compatibility

pkg/helper/openapi_utils.go
miguelsorianod: In some K8s objects a label is created with the same value as the resource name which effectively means that the K8s name cannot be longer than the label limit. I think the case where I detected that  was in `Job` objects. That's why the validation was done more strict in that case.

pkg/controller/openapi/openapi_controller.go
miguelsorianod: I mean the first one processed by the code. `Data` is a map, there's no order, so basically it will randomly read one entry among the entries if there is more than one of them.

pkg/controller/openapi/openapi_controller.go
miguelsorianod: I hope this not causes confusion/missunderstandings because if users change Products/Backends (try to do it) they will potentially get the impression they can because the OpenAPI controller will not react to those changes in any way. It might happen that they do a change in time N and those changes are reverted in time N+M when OpenAPI controller receives an event, where M could be hours/days...

pkg/controller/openapi/openapi_status_reconciler.go
miguelsorianod: What request are you referring to? Do you mean in the OpenAPI spec document?

pkg/controller/openapi/openapi_controller.go
miguelsorianod: :+1:

doc/operator-application-capabilities.md
```diff
@@ -735,9 +873,9 @@ Refer to [Tenant CRD Reference](tenant-reference.md) documentation for more info
 * Deletion of a [Product CR](product-reference.md) is not reconciled. Existing Product in 3scale will not be deleted. [THREESCALE-5539](https://issues.redhat.com/browse/THREESCALE-5539)
 * [Product CRD](product-reference.md) Single sign on (SSO) authentication for the admin and developers portal
 * [Product CRD](product-reference.md) OpenID Connect authentication [THREESCALE-5537](https://issues.redhat.com/browse/THREESCALE-5537)
+* [Product CRD](product-reference.md) Policy chaing management [THREESCALE-6235](https://issues.redhat.com/browse/THREESCALE-6235)
```
miguelsorianod, line 876: typo: `chaing` -> `chain`

pkg/apis/capabilities/v1beta1/openapi_types.go
eguzki: Does not make sense having the info in a secret for the openapi CR and open field in the product CR. This issue has to be tacked separately.

pkg/helper/openapi_utils.go
eguzki: But that does not apply to products and backends created by openapi CR

I will change validation

pkg/controller/openapi/openapi_controller.go
eguzki: Check doc. The implementation is consistent with the doc. Is the doc incorrect? Should the secret be processed otherwise?

pkg/controller/openapi/openapi_status_reconciler.go
eguzki: https://issues.redhat.com/browse/THREESCALE-6197

miguelsorianod: :+1:

pkg/apis/capabilities/v1beta1/openapi_types.go
miguelsorianod: I don't see why it does not make sense. They are different objects with different interfaces.
If we leave this to be changed and the issue for some reason were to be tackled after a release then it would be even harder.

Up to you but I don't see the need for this to be tackled separated for this resource.

pkg/helper/openapi_utils.go
miguelsorianod: :+1: . Whatever the implementation is, if you think it will be likely for the limit to be reached we should add user-facing documentation. If it's not likely to happen then we are ok a doc level

pkg/controller/openapi/openapi_controller.go
miguelsorianod: I checked the doc already on my first reply.

My original comment was asking confirmation about the implementation and based on the answers I think my explanation is accurate.

If we are talking about the doc, it's not wrong, it's not very accurate.

```
The secret must have at least one field with value set to the openapi document content. The field name will not be read.
```

If it has more than one field, the field to be read will be implementation-specific (random order). Maybe it would be safer to document that it has to have `one field`, and not `at least one field`. I'm fine with current content though so up to you. Again, the original comment was not about doc, was about asking for confirmation of understanding about the implementation.

pkg/apis/capabilities/v1beta1/openapi_types.go
eguzki: It does not make sense because the purpose of the keeping some sensitive data in a secret does not apply. The token will be visible in the Product CR. I would prefer having now as it is and fix the issue separately for both OpenAPI CRD and Product CRD, making both reference to a secret. So the sensitive data lives in a single place "securized" in a secret.

miguelsorianod: :+1: as you prefer

miguelsorianod (approved):
This looks good to me.

There's some comments but they are not blocking to me so up to you if you want to change something or not

pkg/helper/openapi_utils.go
eguzki: added user-facing doc about max length size of the OpenAPI title.

pkg/controller/openapi/openapi_controller.go
eguzki: Added doc and required validation so the secret only contains a single field

pkg/controller/openapi/openapi_controller.go
eguzki: Yes, this is tricky. Like configmaps and secrets. If the user updates them manually, the operator does not get event and does not rollout deployment to refresh configuration.

miguelsorianod (approved):

pkg/controller/openapi/openapi_controller.go
miguelsorianod: :+1:

pkg/controller/openapi/openapi_controller.go
miguelsorianod: :+1:

pkg/helper/openapi_utils.go
miguelsorianod: :+1:

PR #495, opened 2020-10-19
System Web App Grafana Dashboard

guicassolato:
- Adds Prometheus target `<system-app-container>/yabeda-metrics`
    Screenshot 2020-10-19 at 13 18 57
- Adds Grafana panels for the System Web App metrics introduced with [THREESCALE-5697](https://issues.redhat.com/browse/THREESCALE-5697)
    Screenshot 2020-10-19 at 13 19 09

Closes [THREESCALE-6186](https://issues.redhat.com/browse/THREESCALE-6186), [THREESCALE-6204](https://issues.redhat.com/browse/THREESCALE-6204)

PR #498, opened 2020-10-22
upgrade apicast and system sidekiq prometheus rules

eguzki:
I have compared monitoring resources differences between 2.9 and 2.10 that require upgrading procedure not implemented currently. I found:
- [x] Apicast prometheus rules updated
- [x] System Sidekiq prometheus rules updated

### DIFF
--- a/pkg/3scale/amp/operator/upgrade.go
+++ b/pkg/3scale/amp/operator/upgrade.go
@@ -11,6 +11,7 @@ import (
 	"github.com/3scale/3scale-operator/pkg/helper"
 	"github.com/3scale/3scale-operator/pkg/reconcilers"
 
+	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/go-logr/logr"
 	"github.com/google/go-cmp/cmp"
 	appsv1 "github.com/openshift/api/apps/v1"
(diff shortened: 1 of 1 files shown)


pkg/3scale/amp/operator/upgrade.go
```diff
@@ -424,6 +425,20 @@ func (u *UpgradeApiManager) upgradeMonitoringSettings() (reconcile.Result, error
 	}
 	updated = updated || updatedTmp
 
+	// Apicast rules
+	updatedTmp, err = u.upgradePrometheusRules(component.ApicastPrometheusRules(u.apiManager.Namespace))
```
miguelsorianod, line 429: With the current implementation it seems there's no real good solution for a problem like this:

If we upgrade or reconcile Prometheus Rules, possible changes that users might have done will be undone. As @slopezz commented in https://github.com/3scale/3scale-operator/pull/495#issuecomment-713660615 there's a very real possibility users might be doing that for Prometheus Rules, although as far as I know we have never really told/documented that that can be changed so far.
If we don't upgrade nor reconcile Prometheus Rules then the users will not have newer changes that we do on the Prometheus rules.

eguzki: We could implement upgrade process to add only new rules. Do not reconcile entire content, only new content. Of course, if there are conflicts like: customer updates rule A, new release also updates rule A, then there is nothing the operator can do.

miguelsorianod: If we only add new content then updated/fixed content of existing alerts will be not incorporated.

I think at this moment with the current implementation the options we have are reverting possible manually made changes by the users at upgrade time or not providing those updated/fixed changes. It is basically a trade-off.

eguzki:
We have decided to implement the upgrade of the prometheus rules, overwriting any custom update the customer could have done.

miguelsorianod (approved):

PR #499, opened 2020-10-22
Fix network tx query on k8s resources namespace dashboard

slopezz:
There was an error on the *Rate of Transmitted Packets* panel query using `receive` metric instead of `transmit` on kubernetes resources namespace dashboard.

This dashboard comes from the upstream **kubernetes-mixin** repository (that fix have already been added there https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/507)


miguelsorianod (approved):

miguelsorianod:
cc @eguzki we should backport this to 2.10 branch.
Also, this will require an upgrade or implement reconciliation for the dashboard as with other cases we are currently discussing.

eguzki (approved):

PR #502, opened 2020-10-22
Feat/Add system prometheus job down alerts

slopezz:
This PR:
- Replaces initial example (not real) system sidekiq alert with a real prometheis job down alert (now that system monitoring is enabled)
- Add system-app prometheus job down alert
- Rename `system` prometheus rule function to `system-app`, to be consistent with its own `PodMonitor` called `system-app`, and the rest of the components' name


eguzki (approved):
LGTM

The backport for 2.10 will require upgrade process. We will do that.

miguelsorianod (approved):

eguzki:
rebasing and merge when tests pass

PR #509, opened 2020-10-30
Only log grafanadashboards/prometheusrules required error when monitoring is enabled

miguelsorianod:
This PR changes logging behavior for grafanadashboards/prometheusrules Kind checking to only log that information
when monitoring is enabled.

This addresses https://issues.redhat.com/browse/THREESCALE-6279 where the original log messages were confusing because they were logged even when monitoring was disabled if the kind didn't exist.


pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
```diff
@@ -85,14 +87,18 @@ func (r *BaseAPIManagerLogicReconciler) ReconcileRoleBinding(desired *rbacv1.Rol
...
+	if !r.apiManager.IsMonitoringEnabled() {
+		return nil
+	}
+
 	kindExists, err := r.HasGrafanaDashboards()
 	if err != nil {
 		return err
 	}
-
 	if !kindExists {
```
miguelsorianod, line 98: Before we always logged when the kind didn't exist, independently on whether monitoring was enabled or not, and we logged the message as Info type.
Now we only log that message when monitoring is enabled and the kind doesn't exist, and in that case we return as an error. However, I'm not entirely sure about this last behavior. If monitoring is enabled and the kind doesn't exist, should we stop processing the next resources?

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
```diff
@@ -106,11 +113,17 @@ func (r *BaseAPIManagerLogicReconciler) ReconcilePrometheusRules(desired *monito
...
+	if !r.apiManager.IsMonitoringEnabled() && !kindExists {
 		return nil
 	}
 
+	if r.apiManager.IsMonitoringEnabled() && !kindExists {
+		err = fmt.Errorf("Error creating prometheusrule object '%s'", desired.Name)
+		r.Logger().Error(err, "Install prometheus-operator in your cluster to create prometheusrules objects")
+		return err
+	}
+
```
miguelsorianod, line 126: Before we always logged the message when the kind didn't exist, independently on whether monitoring was enabled or not, and we logged the message as Info type.
Now we only log that message when monitoring is enabled and the kind doesn't exist, and in that case we log as error and return an error instead of nil. However, I'm not entirely sure about this last behavior. If monitoring is enabled and the kind doesn't exist, should we stop processing the next resources?

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
```diff
@@ -90,11 +92,16 @@ func (r *BaseAPIManagerLogicReconciler) ReconcileGrafanaDashboard(desired *grafa
...
-	if !kindExists {
-		r.Logger().Info("Install grafana-operator in your cluster to create grafanadashboards objects", "Error creating grafanadashboard object", desired.Name)
+	if !r.apiManager.IsMonitoringEnabled() && !kindExists {
 		return nil
 	}
 
+	if r.apiManager.IsMonitoringEnabled() && !kindExists {
+		err = fmt.Errorf("Error creating grafanadashboard object '%s'", desired.Name)
+		r.Logger().Error(err, "Install grafana-operator in your cluster to create grafanadashboards objects")
+		return err
```
eguzki, line 102: Why return error? with this implementation 3scale will not be deployed when monitoring enabled and CRDs not available.

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
```diff
@@ -90,11 +92,16 @@ func (r *BaseAPIManagerLogicReconciler) ReconcileGrafanaDashboard(desired *grafa
 		return err
 	}
 
-	if !kindExists {
-		r.Logger().Info("Install grafana-operator in your cluster to create grafanadashboards objects", "Error creating grafanadashboard object", desired.Name)
+	if !r.apiManager.IsMonitoringEnabled() && !kindExists {
```
eguzki, line 95: What I would do and, IMO, much easier to understand:

```
if !kindExists {
    if r.apiManager.IsMonitoringEnabled() {
              r.Logger().Info("Install grafana-operator in your cluster to create grafanadashboards objects", "Error reconciling grafanadashboard object", desired.Name)
   
    }
    return nil
}
```

eguzki: I would also consider about creating namespace event for more visibility

eguzki (changes requested):

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
miguelsorianod: See my https://github.com/3scale/3scale-operator/pull/509#discussion_r515076719 comment.

I was commenting exactly this.

If we don't return an error when monitoring is enabled it will proceed with creating other elements, and it will mask that something's gone wrong
If we return an error it will not proceed creating other elements. 3scale will not be deployed but that something's gone wrong will be more evident.

It's up to us what we want to do with this behavior.

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
miguelsorianod: I thought about that other possibility yes.

Implementation is easier to understand depending on what we do with what's commented in https://github.com/3scale/3scale-operator/pull/509#discussion_r515076719.

:+1: about the event.

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
eguzki: I would not change the current behavior. Any reason to not deploy 3scale if monitoring stack is not available?

miguelsorianod: It seems just a conceptual behavior.

"If a user sets monitoring:enabled, and the kind does not exist, is it correct what we don't return an error"?

miguelsorianod: We continue with current behavior of logging and proceeding with other elements.

I'll add an event too to try to make it more clear

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
miguelsorianod: Refactored with this approach :+1:

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
miguelsorianod: We decide to keep current behavior. `nil` is returned to continue deploying other objects.
We now log Error instead of Info.
Additionally we now send event (of type Warning, there's no Error event type)

miguelsorianod:
Ready for review @eguzki

eguzki (approved):

PR #512, opened 2020-11-02
3scale 2.9 stable prepare

eguzki:


miguelsorianod (approved):
I understand the list of commits taken have been taken from the range starting from `3scale-2.9-base` to `3scale-2.9.1-GA` tag, modifying some of the commits to adapt it to community values (quay image urls...).

If my understanding is correct, looks good to me :+1:

PR #514, opened 2020-11-04
Update operator-sdk to v1.1.0

miguelsorianod:


miguelsorianod:
Added 3scale version Prometheus metric registration

eguzki (approved):
Let the hunger games start!

PR #517, opened 2020-11-11
Add app label to 3scale CRDs

miguelsorianod:
Adds the 'app' label to 3scale CRDs with value `3scale-api-management`. The chosen value is the same as the default 'app' label value when an APIManager CR is created (although there it can be changed by the user).

operator-sdk and kubebuilder do not provide a way to set the label through any marker so a patch to include the label has been needed.

I've observed in kustomize there's a global variable named `commonLabels` but that adds the labels to all objects, not to just CRDs and also adds it to selectors so that option has not been chosen.


miguelsorianod:
Ready for review @eguzki

eguzki (approved):

PR #522, opened 2020-11-16
apimanager: Add configurable ServiceAccount ImagePullSecrets for managed DCs

miguelsorianod:
This PR adds the ability for the users to configure Image Pull Secret name references to APIManager's managed DCs.
APIManager's managed DCs use two ServiceAccounts:
* `amp`
* `zync-que-sa`

This allows users specify additional authenticated registries for the images that might be pulled by the managed DCs Pods.

A new field `.spec.ImagePullSecrets` has been introcuced in the APIManager CR. Being a []v1.LocalObjectReference.

If this field is not specified then the `threescale-registry-auth` Secret name is used by default in the previous ServiceAccounts.
If this field is specified then the specified secret names are set into the `amp` and `zync-que-sa`. In this case `threescale-registry-auth` has to be provided by the user as part of the list in `.spec.ImagePullSecrets` (or another secret name that contains registry.redhat.io credentials). Otherwise pods won't be able to pull from registry.redhat.io)

Limitations:
* If an update is done to the `.spec.ImagePullSecrets` field, ServiceAccounts are reconciled appropriately but the changes are NOT effective. The reason for that is that even though the ServiceAccounts are reconciled K8s is not automatically aware of those changes at Pod level. To make those changes effective the user has to redeploy the pod, by increasing the number of replicas to force redeployment of new pods.
* Users cannot specify `-token` or `-dockercfg` substring as a Secret name when being used in `spec.ImagePullSecrets`. See code comment in the PR for details of this.

Pending:

- [x] Testing


pkg/reconcilers/serviceaccount.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	desired, ok := desiredObj.(*v1.ServiceAccount)
+	if !ok {
+		return false, fmt.Errorf("%T is not a *v1.ServiceAccount", desiredObj)
+	}
+
+	updated := false
+
+	// Copy received ServiceAccounts due to we are going to
+	// sort (modify) its ImagePullSecrets
```
miguelsorianod, line 26: Sorting allows us to do the comparison. However, it means that `.spec.ImagePullSecrets` array positions do not matter when reconciliating.

```
- a
- b
```

vs

```
- b
- a

Will not be detected as a change and will not trigger any reconciliation

Comment (eguzki) [reply pkg/reconcilers/serviceaccount.go:26]: no issue, order is not important. The CRD type could be a map also to add some "secret" name at the CR level.

**Comment by eguzki** [comment]
Overall this is a good improvement. 

* I would not have default `threescale-registry-auth`. This is community operator and does not require it at all, does it? In downstream operator, I would add a validation that reads pull secrets referenced and looks for `registry.redhat.io`. If not found, tries to find `threescale-registry-auth` secret and adds it to the CR `imagepullsecrets` field.

* I would change the serviceaccount reconciliation logic to "merge" approach. More robust implementation and does not require to check for specific k8s secrets. The downside is that any extra secret referenced in the existing object would  still be there, but I think it is not a big issue.

File: pkg/reconcilers/serviceaccount.go (hunk 1 above)
Comment (miguelsorianod) [reply pkg/reconcilers/serviceaccount.go:26]: I'll leave it as array to match how is defined in K8s

**Comment by miguelsorianod** [comment]
>  I would not have default threescale-registry-auth. This is community operator and does not require it at all, does it? In downstream operator, I would add a validation that reads pull secrets referenced and looks for registry.redhat.io. If not found, tries to find threescale-registry-auth secret and adds it to the CR imagepullsecrets field.

Your comment seems an interesting suggestion.

In community operator `threescale-registry-auth` it seems it does not need to exist, even if defined in the ServiceAccount ImagePullPolicies attribute. It seems it can run even if not existing.

Doing that would also require reading potentially all secrets specified each reconciliation (for existing and desired) which I was trying to avoid.
Also we would need a mechanism to have differences between upstream and downstream, which we don't have at the moment and I'm not sure whether having differences of this size is desirable


> I would change the serviceaccount reconciliation logic to "merge" approach. More robust implementation and does not require to check for specific k8s secrets. The downside is that any extra secret referenced in the existing object would still be there, but I think it is not a big issue.

I started doing it with a merge approach but I changed it to "set" approach because of the reason you mentioned. It seemed nicer to remove them from the SA when they were removed from the CR

File: doc/apimanager-reference.md [hunk 2]
```diff
@@ -72,6 +72,7 @@ Generated using [github-markdown-toc](https://github.com/ekalinin/github-markdow
...
 | TenantName | `tenantName` | string | No | `3scale` | Tenant name under the root that Admin UI will be available with -admin suffix.
 | ImageStreamTagImportInsecure | `imageStreamTagImportInsecure` | bool | No | `false` | Set to true if the server may bypass certificate verification or connect directly over HTTP during image import |
+| ImagePullSecrets | `imagePullSecrets` | [corev1.LocalObjectReference](https://v1-17.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#localobjectreference-v1-core) | No | `[ { name: "threescale-registry-auth" } ]` | List of image pull secrets to be used on the managed DeploymentConfigs ServiceAccounts. See [imagePullSecrets field in K8s ServiceAccount documentation](https://v1-17.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#serviceaccount-v1-core) for details on Image pull secrets. If not specified, `threescale-registry-auth` is used. Secret names that contain `dockercfg-` or `token-` anywhere in part of its name cannot be specified. If an update to this attribute is performed the corresponding DeploymentConfig pods have to be redeployed by the user to make the changes effective |
```
eguzki, line 75: the type is an array of corev1.LocalObjectReference

pkg/reconcilers/serviceaccount_test.go
```diff
@@ -0,0 +1,128 @@
...
+		v1.LocalObjectReference{Name: "imagepullsecret3"},
+	}
+	desired := serviceAccountTestFactory(desiredImagePullSecrets)
+
+	changed, err := ServiceAccountImagePullPolicyMutator(existing, desired)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if changed == false {
```
eguzki, line 34: :eyes:

eguzki (changes requested):

doc/apimanager-reference.md
miguelsorianod: Fixed

pkg/reconcilers/serviceaccount_test.go
miguelsorianod: Changed to `!changed`

eguzki (approved):

PR #524, opened 2020-11-16
Temporarily remove auth proxy metrics reader Service from bundle

miguelsorianod:
Temporarily remove the auth proxy metrics reader Service
Due to a bug in OLM where upgrading a CSV fails when providing
a Service object as part of the bundle manifests.

At the moment of writing this when an upgrade is performed in OLM and a Service is part of the bundle (in both the previous and current version) the upgrade OLM InstallPlan fails with:
```
  {
    "lastTransitionTime": "2020-11-13T18:25:14Z",
    "lastUpdateTime": "2020-11-13T18:25:14Z",
    "message": "error updating service: apicast-operator-controller-manager-metrics-service: Service \"apicast-operator-controller-manager-metrics-service\" is invalid: [metadata.resourceVersion: Invalid value: \"\": must be specified for an update, spec.clusterIP: Invalid value: \"\": field is immutable]",
    "reason": "InstallComponentFailed",
    "status": "False",
    "type": "Installed"
  }
```


eguzki (approved):

miguelsorianod:
Related OLM bug issue that I've reported: https://bugzilla.redhat.com/show_bug.cgi?id=1898500

PR #527, opened 2020-11-17
Improve APIManager reconcile times

miguelsorianod:
### Summary
I detected reconciliation times for APIManager were abnormally long.
This PR improves the time it takes to perform a reconciliation event from 1.5 minutes to 2-3 seconds approximately (a 96.7%  reduction approximately)

### Detailed investigation
After investigating the issue I noticed that reconciliation was taking abnormally long time in the APIManager controller. After measuring it I observed it was taking abnormally long times:
```
2020-11-17T14:42:13.952+0100	INFO	controllers.APIManager	Full reconciliation took 1m33.955830582s
```
Going into a little bit more of detail I observed that the cause of this long execution times was due to the requests we do to check whether monitoring CRDs exist or not via the K8s DiscoveryClient. This was introduced when monitoring functionality was added into the operator. This slowness happened independently of the value of the `.spec.monitoring.enabled` flag.

I measured that each call to check whether a CRD existed or not with the DiscoveryClient took 3-4 seconds approximately:
```
2020-11-17T14:51:44.736+0100    INFO    controllers.APIManager  Backend Worker PrometheusRules reconcile took 4.119325517s      {"APIManager Controller": "example-apimanager"}
```

In each reconcile event we do multiple calls to check whether CRDs exist. One time per each CR related to monitoring that we have thus explaining this 1.5 minutes for each reconciliation event.

To reduce this long times two optimizations have been applied:
1. The DiscoveryClient call to `ServerGroupsAndResources` has been replaced by `ServerResourcesForGroupVersion`. This reduced the processing that the K8s APIServer had to do and the response size reducing considerably the time taken for the calls
2. In BaseAPIManagerLogicReconciler we have added a cache that contains CRD availability states allowing us avoiding to do real API calls with the DiscoveryClient once cached. Now all APIManager contents receive the same shared BaseAPIManagerLogicReconciler instance so they are able to share the cache. BaseAPIManagerLogicReconciler is created once per reconcile event so the cache lifecycle is one reconciliation event.

### Results
This optimizations have reduced the reconciliation event times to 2-3 seconds:
```
2020-11-17T18:04:52.855+0100    INFO    controllers.APIManager  Full reconciliation took 2.683286778s
```

Now real calls to DiscoveryClient CRD checking take 200-500ms:
```
2020-11-17T18:04:52.267+0100    INFO    controllers.APIManager  Backend Worker PodMonitor reconcile took 292.082717ms   {"APIManager Controller": "example-apimanager"}
```

And cached calls are in the order of microseconds:
```
2020-11-17T18:04:52.267+0100    INFO    controllers.APIManager  Backend Listener PodMonitor reconcile took 24.658µs     {"APIManager Controller": "example-apimanager"}
```

### DIFF
--- a/controllers/apps/apimanager_controller.go
+++ b/controllers/apps/apimanager_controller.go
@@ -191,63 +191,64 @@ func (r *APIManagerReconciler) setAPIManagerDefaults(cr *appsv1alpha1.APIManager
 }
 
 func (r *APIManagerReconciler) reconcileAPIManagerLogic(cr *appsv1alpha1.APIManager) (reconcile.Result, error) {
-	imageReconciler := operator.NewAMPImagesReconciler(operator.NewBaseAPIManagerLogicReconciler(r.BaseReconciler, cr))
+	baseAPIManagerLogicReconciler := operator.NewBaseAPIManagerLogicReconciler(r.BaseReconciler, cr)
+	imageReconciler := operator.NewAMPImagesReconciler(baseAPIManagerLogicReconciler)
 	result, err := imageReconciler.Reconcile()
 	if err != nil || result.Requeue {
 		return result, err
 	}
 
 	if !cr.IsExternalDatabaseEnabled() {
-		redisReconciler := operator.NewRedisReconciler(operator.NewBaseAPIManagerLogicReconciler(r.BaseReconciler, cr))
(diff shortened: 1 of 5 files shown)


eguzki (approved):

miguelsorianod:
Added some tests

eguzki (approved):

PR #531, opened 2020-11-19
Open id connect authentication

### LINKED ISSUE #337: Configure SSO/OAuth integration for API Manager
It would be useful to be able to configure authentication (Authn/Authz) providers from the Operator to fully enable Infrastructure-as-Code. For example, being able to deploy Keycloak and then provision the APIManager to use that KeyCloak instance based on the contents of the CR. Perhaps a new/separate CRD for Auth providers?

### LINKED ISSUE #351: Add support to configure oidc as an authentication mecanism
Add support to configure oidc as an authentication mecanism using API CRD

eguzki:
Fixes #337
Fixes #351


apis/capabilities/v1beta1/product_types.go
```diff
@@ -322,14 +322,86 @@ func (u *UserKeyAuthenticationSpec) GatewayResponseSpec() *GatewayResponseSpec {
 	return u.GatewayResponse
 }
 
+// OIDCAuthenticationFlowSpec defines the desired OAuth2.0 authorization grant type
+type OIDCAuthenticationFlowSpec struct {
```
miguelsorianod, line 326: Are they all optional? if so we should set them as pointers and mark them as optional appropriately as per k8s API conventions guide (https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md)

miguelsorianod (changes requested):
Looks good to me.

Just a question about a clarification regarding the optionality of some attributes

apis/capabilities/v1beta1/product_types.go
eguzki: They are all required. It is the attribute of the OIDCSpec `AuthenticationFlow` the one which is optional.
```
// AuthenticationFlow specifies OAuth2.0 authorization grant type                              
// +optional                                                                                   
AuthenticationFlow *OIDCAuthenticationFlowSpec `json:"authenticationFlow,omitempty"`           
```

miguelsorianod: :+1:

miguelsorianod (approved):

PR #535, opened 2020-11-24
Update CustomResourceDefinition API version to v1

miguelsorianod:
This PR updates the API Version of the `CustomResourceDefinition` API type from `v1beta1` to `v1` for our CRDs.
The motivation of this is that K8s is going to stop serving `v1beta1` in K8s 1.22:
```
Warning: apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
```

related issue: https://github.com/kubernetes/kubernetes/issues/82022

It is important to not confuse the APIVersion of the `CustomResourceDefinition` API type with the APIVersion of our CRD Types. They are different concepts. In this PR we are dealing with the former.

The following changes have been performed:
* Change controller-gen flags to generate `v1` version of CustomResourceDefinition.
* Update CRD Kustomize patch files to reference to CustomResourceDefinition `v1`
* Update the CRD kustomize patch files that deal with OpenAPI validation modifications on the CRD resources to use JSON Patch instead of JSON Strategic Merge patch. The reason for this is that now that the OpenAPI validation appears in an array of versions there's no way to perform a merge with JSON Strategic Merge patch on the array element contents as it lacks a merge key definition
* Update manifests and bundle CRDs with the new `v1` CustomResourceDefinition format. The OpenAPI validation is now specified in the versions array instead of directly in the spec.
* Update CRD manifest tests to use versioned retrieval of the CRD type schema.


miguelsorianod:
Updated PR adapting `ActiveDoc` and `Policy` new CRDs to CRD v1 (including patches, ...)

miguelsorianod:
Added a new commit that removes the `run-operator-e2e-test` command and its execution at pr and nightly level from CircleCI, as e2e tests are now run at pr level on OpenShift CI. Other tasks related to operator are still executed in CircleCI

miguelsorianod:
@eguzki I think this is ready to be merged

.circleci/config.yml
```diff
@@ -590,9 +560,6 @@ workflows:
       - run-unit-tests
       - bundle-validate
       - install-operator
```
eguzki, line 562: why leave install-operator job?

.circleci/config.yml
```diff
@@ -611,10 +578,6 @@ workflows:
           requires:
             - build-push-3scale-images
       - install-operator
```
eguzki, line 580: why leave install-operator job?

eguzki (changes requested):
.circleci/config needs some cleanup, but overall, LGTM

.circleci/config.yml
miguelsorianod: I preferred to leave it as it can be executed properly, even if redundant due to done in e2e tests. It is also used in the "operator-release" and "nightly" jobs

eguzki: I am not saying deleting the job itself. I am suggesting removing from the nightly and operator workflows as it is useless.

miguelsorianod: For nightly I think it might be useful. Same for operator-release job.
It's better than nothing.

For the "operator" job we could remove it as it's already covered by the OpenShift CI tests but anyway it did not harm and runs quick

eguzki: If you mean building the operator image as part of the nightly workflow, fine. But install-operator job does more than that and it it useless. Check it, exporting a whole docker image to tarball and persist it to workspace for nothing.

miguelsorianod: In the "operator-release" workflow those additional steps you mentioned are actually needed.

In the nightly job those additional steps are not needed. I'll change the call to install-operator to build-operator-image and I'll remove it from the operator job as it's already covered by OpenShift CI tests

miguelsorianod: I've applied the changes I've described

eguzki: yes, I never said to delete the install-operator job or remove it from the `operator-release` workflow

miguelsorianod: :+1: 

Take a look at the current changes

.circleci/config.yml
```diff
@@ -610,11 +585,7 @@ workflows:
       - deploy_templates:
           requires:
             - build-push-3scale-images
-      - install-operator
-      - run-operator-e2e-test:
-          requires:
-            - install-operator
-            - build-push-3scale-images
+      - build-operator-image-only
```
eguzki, line 588: I would also add to the `operator` workflow to have the image built for each PR

miguelsorianod: Changed

test/crds/crd_validation_test.go
```diff
@@ -118,6 +183,14 @@ func getSchema(t *testing.T, crd string) validation.Schema {
 	return schema
 }
 
+func getSchemaVersioned(t *testing.T, crd string, version string) validation.Schema {
```
eguzki, line 186: Shouldn't `func getSchema(t *testing.T, crd string) validation.Schema` method be deleted?

miguelsorianod: There are both variants

eguzki: but it is being used?

miguelsorianod: It's not currently being used. We can decide to keep it available for use or deleting it. I decided the former but I'm fine with both

eguzki: private method not being used => delete.

we always have git history to restore it.

miguelsorianod: Deleted

eguzki (approved):

PR #536, opened 2020-11-24
Fix system env var backend route

eguzki:
Recently, it was merged in Porta [this PR](https://github.com/3scale/porta/pull/2268) 

Now, they are respecting the `BACKEND_ROUTE` env var set in deployment. 

Before, Porta was reading some unset env var `THREESCALE_CORE_INTERNAL_API` and it did not exist, so they were using as backend route: `http://backend-listener:3000/internal/`. It worked as expected. 

But now, as `BACKEND_ROUTE` env var is respected, in 3scale operator CI, 3scale system tries to connect to `https://backend-3scale.lvh.me`, and ovbiously, it fails. 
```
Errno::ECONNREFUSED: Connection refused - connect(2) for "backend-3scale.lvh.me" port 443
/opt/system/vendor/bundle/ruby/2.5.0/gems/net-http-persistent-3.0.0/lib/net/http/persistent.rb:692:in `start'
/opt/system/vendor/bundle/ruby/2.5.0/gems/net-http-persistent-3.0.0/lib/net/http/persistent.rb:622:in `connection_for'
/opt/system/vendor/bundle/ruby/2.5.0/gems/net-http-persistent-3.0.0/lib/net/http/persistent.rb:927:in `request'
/opt/system/vendor/bundle/ruby/2.5.0/gems/faraday-0.15.3/lib/faraday/adapter/net_http_persistent.rb:36:in `perform_request'
/opt/system/vendor/bundle/ruby/2.5.0/gems/faraday-0.15.3/lib/faraday/adapter/net_http.rb:43:in `block in call'
/opt/system/vendor/bundle/ruby/2.5.0/gems/faraday-0.15.3/lib/faraday/adapter/net_http.rb:92:in `with_net_http_connection'
/opt/system/vendor/bundle/ruby/2.5.0/gems/faraday-0.15.3/lib/faraday/adapter/net_http.rb:38:in `call'
/opt/system/vendor/bundle/ruby/2.5.0/gems/faraday-0.15.3/lib/faraday/rack_builder.rb:143:in `build_response'
/opt/system/vendor/bundle/ruby/2.5.0/gems/faraday-0.15.3/lib/faraday/connection.rb:387:in `run_request'
/opt/system/vendor/bundle/ruby/2.5.0/gems/faraday-0.15.3/lib/faraday/connection.rb:138:in `get'
/opt/system/vendor/bundle/ruby/2.5.0/gems/pisoni-1.29.0/lib/3scale/core/api_client/operations.rb:103:in `api_http'
/opt/system/vendor/bundle/ruby/2.5.0/gems/pisoni-1.29.0/lib/3scale/core/api_client/operations.rb:186:in `api'
/opt/system/lib/tasks/boot.rake:6:in `block (2 levels) in <top (required)>'
/opt/system/vendor/bundle/ruby/2.5.0/gems/bugsnag-6.11.1/lib/bugsnag/integrations/rake.rb:18:in `execute_with_bugsnag'
/opt/system/vendor/bundle/ruby/2.5.0/gems/rake-13.0.1/exe/rake:27:in `<top (required)>'
/opt/rh/rh-ruby25/root/usr/local/share/gems/gems/bundler-1.17.3/lib/bundler/cli/exec.rb:74:in `load'
/opt/rh/rh-ruby25/root/usr/local/share/gems/gems/bundler-1.17.3/lib/bundler/cli/exec.rb:74:in `kernel_load'
/opt/rh/rh-ruby25/root/usr/local/share/gems/gems/bundler-1.17.3/lib/bundler/cli/exec.rb:28:in `run'
/opt/rh/rh-ruby25/root/usr/local/share/gems/gems/bundler-1.17.3/lib/bundler/cli.rb:463:in `exec'
/opt/rh/rh-ruby25/root/usr/local/share/gems/gems/bundler-1.17.3/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
/opt/rh/rh-ruby25/root/usr/local/share/gems/gems/bundler-1.17.3/lib/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
/opt/rh/rh-ruby25/root/usr/local/share/gems/gems/bundler-1.17.3/lib/bundler/vendor/thor/lib/thor.rb:387:in `dispatch'
/opt/rh/rh-ruby25/root/usr/local/share/gems/gems/bundler-1.17.3/lib/bundler/cli.rb:27:in `dispatch'
/opt/rh/rh-ruby25/root/usr/local/share/gems/gems/bundler-1.17.3/lib/bundler/vendor/thor/lib/thor/base.rb:466:in `start'
/opt/rh/rh-ruby25/root/usr/local/share/gems/gems/bundler-1.17.3/lib/bundler/cli.rb:18:in `start'
/opt/rh/rh-ruby25/root/usr/local/share/gems/gems/bundler-1.17.3/exe/bundle:30:in `block in <top (required)>'
/opt/rh/rh-ruby25/root/usr/local/share/gems/gems/bundler-1.17.3/lib/bundler/friendly_errors.rb:124:in `with_friendly_errors'
/opt/rh/rh-ruby25/root/usr/local/share/gems/gems/bundler-1.17.3/exe/bundle:22:in `<top (required)>'
/opt/rh/rh-ruby25/root/usr/local/bin/bundle:23:in `load'
/opt/rh/rh-ruby25/root/usr/local/bin/bundle:23:in `<main>'
Tasks: TOP => boot => boot:all => boot:backend
(See full trace by running task with --trace)
```

The initial consequence is that CI tests for nightly builds do not work, therefore, nightly images are not being pushed.  

The proposed solution is to pass internal service endpoint which uses OCP service domain name.

An alternative solution would be to leave templates as they are. When OCP cluster domain names are not fake (like mydomain.lvh.me), the route is supposed to work. Maybe it is missing the `/internal` path. With this approach, our CI tests should be changed to override the passed env var to use OCP service endpoint, because cluster domain name is not usable. I did not pick this approach because there is no need to pass FQDN to system to reach backend, as they are pods inside the same namespace and service level connection is preferred. 

Checked that apicast services receive external backend route instead of internal endpoint. Required for self managed apicast services running in another namespace and/or cluster.

TODO:
- [x] upgrade procedure

### DIFF
--- a/pkg/3scale/amp/operator/upgrade.go
+++ b/pkg/3scale/amp/operator/upgrade.go
@@ -6,11 +6,14 @@ import (
 	"reflect"
 
 	appsv1alpha1 "github.com/3scale/3scale-operator/apis/apps/v1alpha1"
+	"github.com/3scale/3scale-operator/pkg/common"
+	"github.com/3scale/3scale-operator/pkg/helper"
 	"github.com/3scale/3scale-operator/pkg/reconcilers"
 
 	"github.com/go-logr/logr"
 	"github.com/google/go-cmp/cmp"
 	appsv1 "github.com/openshift/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
@@ -30,7 +33,15 @@ func NewUpgradeApiManager(b *reconcilers.BaseReconciler, apiManager *appsv1alpha
 }
 
 func (u *UpgradeApiManager) Upgrade() (reconcile.Result, error) {
-	res, err := u.upgradeImages()
+	res, err := u.upgradeBackendRouteEnv()
+	if err != nil {
+		return res, fmt.Errorf("Upgrading backend route env vars: %w", err)
+	}
+	if res.Requeue {
+		return res, nil
+	}
+
+	res, err = u.upgradeImages()
 	if err != nil {
 		return res, fmt.Errorf("Upgrading images: %w", err)
 	}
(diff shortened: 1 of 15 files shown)


eguzki:
@miguelsorianod looking for feedback before proceeding further

eguzki:
@miguelsorianod looking for feedback before proceeding further

miguelsorianod:
Looks good to me.

As an observation, names are super confusing unfortunately because what's called "BACKEND_ROUTE" in system is really backend's Service (as it is called in our Backend code), and we also have an attribute in backend named route_endpoint which is the OCP's Route object for backend, which is not the value used in System's BACKEND_ROUTE

eguzki:
> As an observation, names are super confusing unfortunately because what's called "BACKEND_ROUTE" in system is really backend's Service (as it is called in our Backend code), and we also have an attribute in backend named route_endpoint which is the OCP's Route object for backend, which is not the value used in System's BACKEND_ROUTE

This is System issue, isn't it? we need to use the env vars set by them. 

Is there anything confusing in the naming of this repo?

miguelsorianod:
It's sort of a confusing issue in System yes. Regarding our code, maybe calling the System option `BackendServiceEndpoint  instead of`BackendRouteEndpoint`woud make it more clear and consistent with the source of truth which is the Backend component where the naming contains `*service*`. Although at the end the envvar used by System will be named Route so at some point there will be an inconsistency.

eguzki:
Ok, I see what you mean. The new system option field should be called `BackendServiceEndpoint` instead of `BackendRouteEndpoint`

eguzki:

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -38,6 +41,14 @@ func (u *UpgradeApiManager) Upgrade() (reconcile.Result, error) {
 		return res, nil
 	}
 
+	res, err = u.upgradeBackendRouteEnv()
```
miguelsorianod, line 44: I think we should upgrade the routes before changing the images.

Otherwise the new code that makes use of the environment variable will be effective before we change the route and currently we use an incorrect one.

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -394,6 +405,124 @@ func (u *UpgradeApiManager) findDeploymentTriggerOnImageChange(triggerPolicies [
...
+
+	if len(existing.Spec.Template.Spec.Containers) != 3 {
+		return reconcile.Result{}, fmt.Errorf("DeploymentConfig %s spec.template.spec.containers length is %d, should be 3",
+			existing.Name, len(existing.Spec.Template.Spec.Containers))
+	}
+
+	desiredName := common.ObjectInfo(desired)
+
+	update := false
+	for idx := 0; idx < 3; idx++ {
```
miguelsorianod, line 442: If you want there's no need to hardcode numbers, you can range over the containers attribute

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -394,6 +405,124 @@ func (u *UpgradeApiManager) findDeploymentTriggerOnImageChange(triggerPolicies [
...
+			panic(fmt.Sprintf("%s desired spec.template.spec.containers env var '%s' does not exist", desiredName, "BACKEND_ROUTE"))
+		}
+		tmpUpdate := ensureBackendRouteEnvVar(desiredBackendRouteEnvVar, &existingContainer.Env)
+		update = update || tmpUpdate
+	}
+
+	// Pre hook pod env vars
+	desiredBackendRouteEnvVar, ok := helper.FindEnvVar(desired.Spec.Strategy.RollingParams.Pre.ExecNewPod.Env, "BACKEND_ROUTE")
+	if !ok {
+		panic(fmt.Sprintf("%s desired spec.strategy.rollingparams.pre.execnewpod env var '%s' does not exist", desiredName, "BACKEND_ROUTE"))
```
miguelsorianod, line 456: I knew this shouldn't happen but maybe returning an error so we don't completely crash the operator?

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -394,6 +405,124 @@ func (u *UpgradeApiManager) findDeploymentTriggerOnImageChange(triggerPolicies [
...
+
+	desiredName := common.ObjectInfo(desired)
+
+	update := false
+	for idx := 0; idx < 3; idx++ {
+		existingContainer := &existing.Spec.Template.Spec.Containers[idx]
+		desiredContainer := &desired.Spec.Template.Spec.Containers[idx]
+		desiredBackendRouteEnvVar, ok := helper.FindEnvVar(desiredContainer.Env, "BACKEND_ROUTE")
+		if !ok {
+			panic(fmt.Sprintf("%s desired spec.template.spec.containers env var '%s' does not exist", desiredName, "BACKEND_ROUTE"))
```
miguelsorianod, line 447: I knew this shouldn't happen but maybe returning an error so we don't completely crash the operator?

pkg/3scale/amp/template/adapters/system.go
```diff
@@ -1,6 +1,8 @@
 package adapters
 
 import (
+	"fmt"
```
miguelsorianod, line 4: For 2.11 we need to remember that we'll need to write a templates upgrade procedure for the changes related to this PR

miguelsorianod (changes requested):
Left some comments

pkg/3scale/amp/operator/upgrade.go
eguzki: you mean range over the containers? 

not a big change. And the code just before the loop, makes sure there are 3 containers.

eguzki:
All comments answered :+1:

pkg/3scale/amp/operator/upgrade.go
miguelsorianod: Yeah, precisely because of that. Due to you have checked there are 3 you don't need to worry about the number. Just process them all. As you prefer.

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -394,6 +405,124 @@ func (u *UpgradeApiManager) findDeploymentTriggerOnImageChange(triggerPolicies [
...
+			existing.Name, len(existing.Spec.Template.Spec.Containers))
+	}
+
+	desiredName := common.ObjectInfo(desired)
+
+	existingContainer := &existing.Spec.Template.Spec.Containers[0]
+	desiredContainer := &desired.Spec.Template.Spec.Containers[0]
+	desiredBackendRouteEnvVar, ok := helper.FindEnvVar(desiredContainer.Env, "BACKEND_ROUTE")
+	if !ok {
+		panic(fmt.Sprintf("%s desired spec.template.spec.containers env var '%s' does not exist", desiredName, "BACKEND_ROUTE"))
```
miguelsorianod, line 490: We should apply the same criteria as with the other, just return an error

miguelsorianod (changes requested):
Another comment

pkg/3scale/amp/operator/upgrade.go
eguzki: done

miguelsorianod (approved):

miguelsorianod:
let's do a squash of some commits

PR #539, opened 2020-11-27
ActiveDoc CRD

eguzki:
CRD (and controller) holding OAS3 reference (configmap or URL) to manage 3scale's ActiveDoc

Requires https://github.com/3scale/3scale-porta-go-client/pull/38


eguzki:

Only missing this 3scale/3scale-porta-go-client#38 and updating go.mod to reference the new dependency version. Until then, all tests will fail

apis/capabilities/v1beta1/activedoc_types.go
```diff
@@ -0,0 +1,232 @@
...
+	// Respect 3scale API defaults
+	if a.Spec.SystemName == nil || *a.Spec.SystemName == "" {
+		tmp := activeDocSystemNameRegexp.ReplaceAllString(a.Spec.Name, "")
+		a.Spec.SystemName = &tmp
+		updated = true
+	}
+
+	// 3scale API ignores case of the system name field
+	systemNameLowercase := strings.ToLower(*a.Spec.SystemName)
+	if *a.Spec.SystemName != systemNameLowercase {
```
miguelsorianod, line 202: I wonder if maybe forcing this at openapi validation level would be more appropriate?

config/crd/kustomization.yaml
```diff
@@ -59,6 +69,10 @@ patchesStrategicMerge:
 - patches/additional_app_label_in_crds.yaml
 # +kubebuilder:scaffold:crdkustomizeadditionalapplabelincrdspatch
 
+# [ActiveDoc CRD additional printer columns]. This patch adds spec.additionalPrinterColumns
+- patches/activedoc_additional_columns.yaml
```
miguelsorianod, line 73: I think there's no need to provide a patch for printercolumns.

I've seen there's a kubebuilder CRD marker:
```
// + kubebuilder:printcolumn
```

that we might be able to use. See https://book.kubebuilder.io/reference/markers/crd.html for more details

doc/activedoc-reference.md
```diff
@@ -0,0 +1,155 @@
...
+| Published | `published` | bool | Switch to publish the activedoc. By default it will be `hidden` | No |
+| SkipSwaggerValidations | `skipSwaggerValidations` | bool | Switch to skip OpenAPI validation. By default, the validation is enabled | No |
+
+#### ActiveDocOpenAPIRefSpec
+
+Reference to the OpenAPI Specification
+
+| **Field** | **json field**| **Type** | **Info** | **Required** |
+| --- | --- | --- | --- | --- |
+| SecretRef | `secretRef` | [v1.LocalObjectReference](https://v1-15.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#localobjectreference-v1-core) to [OpenAPI secret reference](#openapi-secret-reference) | The secret that contains the OpenAPI Document | No |
```
miguelsorianod, line 43: This should be v1.ObjectReference instead of v1.LocalObjectReference?

controllers/capabilities/activedoc_threescale_reconciler.go
```diff
@@ -0,0 +1,302 @@
...
+	s.logger.V(1).Info("START")
+
+	remoteActiveDocs, err := s.threescaleAPIClient.ListActiveDocs()
+	if err != nil {
+		return nil, err
+	}
+
+	var remoteActiveDoc *threescaleapi.ActiveDoc
+
+	for idx := range remoteActiveDocs.ActiveDocs {
```
miguelsorianod, line 51: Isn't there a way to get the desired ActiveDoc directly instead of listing them?

controllers/capabilities/activedoc_threescale_reconciler.go
```diff
@@ -0,0 +1,302 @@
...
+				Published:              s.resource.Spec.Published,
+				SkipSwaggerValidations: s.resource.Spec.SkipSwaggerValidations,
+				ServiceID:              desiredProductID,
+			},
+		}
+
+		remoteActiveDoc, err = s.threescaleAPIClient.CreateActiveDoc(newActiveDoc)
+		if err != nil {
+			return nil, err
+		}
```
miguelsorianod, line 91: If there's no error then we continue with the rest of the method related to update?

controllers/capabilities/activedoc_threescale_reconciler.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	update := false
+	updatedActiveDoc := &threescaleapi.ActiveDoc{
+		Element: threescaleapi.ActiveDocItem{
+			ID: remoteActiveDoc.Element.ID,
+		},
+	}
+
+	if remoteActiveDoc.Element.Name == nil || *remoteActiveDoc.Element.Name != s.resource.Spec.Name {
```
miguelsorianod, line 101: Does system allow changing the 'Name' attribute once the object has been created?

controllers/capabilities/activedoc_threescale_reconciler.go
```diff
@@ -0,0 +1,302 @@
...
+		s.logger.V(1).Info("update Name", "Difference", cmp.Diff(remoteActiveDoc.Element.Name, s.resource.Spec.Name))
+		updatedActiveDoc.Element.Name = &s.resource.Spec.Name
+		update = true
+	}
+
+	// s.resource.Spec.SystemName is not nil (defaults are set)
+	if remoteActiveDoc.Element.SystemName == nil || *remoteActiveDoc.Element.SystemName != *s.resource.Spec.SystemName {
```
miguelsorianod, line 108: Does system allow changing the 'SystemName' attribute once the object has been created?

controllers/capabilities/activedoc_threescale_reconciler.go
eguzki: Only by ID. But we use the systemName as the key identifier. If you change the systemname of the CR, the CR will reconcile a 3scale remote object with that systemname, instead of changing the systemname of a existing 3scale remote object (which cannot be done)

controllers/capabilities/activedoc_threescale_reconciler.go
eguzki: Yes

controllers/capabilities/activedoc_threescale_reconciler.go
eguzki: No. The systemName always exist. And 3scale API should always return the systemName. That check is because the field is optional in the client library and if for whatever reason systemname is not in the response, the field would be nil.

controllers/capabilities/activedoc_threescale_reconciler.go
eguzki: This is a common pattern implemented in the latest application capabilities new resources. 

In this specific case, the creation call requires all the parameters and does not make much sense to run reconciliation logic. Anyway does not do any harm and, obviously, it will not update further.

eguzki:
all comments answered

controllers/capabilities/activedoc_threescale_reconciler.go
miguelsorianod: A little bit confusing for me but as you prefer, it should work anyway.

It's common to return early in Go.

miguelsorianod (approved):

PR #544, opened 2020-12-03
Backend grafana charts: set min for y axes in charts

davidor:
Fixes https://issues.redhat.com/browse/THREESCALE-6447

This PR sets 0 as the minimum value for the backend charts.
The problem was that many charts were showing -1 in the y axis when there were no requests for things like requests per second.


eguzki:
Looking good. Re-running failed tests. Happens from time to time, sadly too often lately :(

eguzki (approved):

PR #546, opened 2020-12-07
Policy CRD

eguzki:
CRD (and controller) for managing APIcast Policy schema in the 3scale provider account registry. 

Requires 3scale/3scale-porta-go-client#39

TODO: 
- [x] CR samples
- [x] enable CRD tests
- [x] Usage doc / reference doc


apis/capabilities/v1beta1/policy_types.go
```diff
@@ -0,0 +1,173 @@
...
+
+	// ProviderAccountHost contains the 3scale account's provider URL
+	// +optional
+	ProviderAccountHost string `json:"providerAccountHost,omitempty"`
+
+	// ObservedGeneration reflects the generation of the most recently observed Backend Spec.
+	// +optional
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// Current state of the activedoc resource.
```
miguelsorianod, line 108: Typo. should be policy resource

apis/capabilities/v1beta1/policy_types.go
```diff
@@ -0,0 +1,173 @@
...
+	Version string `json:"version"`
+
+	// Summary is the summary of the policy schema
+	Summary string `json:"summary"`
+
+	// Description is an array of description messages for the policy schema
+	Description *[]string `json:"description,omitempty"`
+
+	// Schema the $schema keyword is used to declare that this is a JSON Schema.
+	Schema string `json:"$schema"`
```
miguelsorianod, line 66: Why the `$` symbol? Can't we call it just `schema`? `$` does not seem common in CR attributes

eguzki: it is a common and well known keyword in JSON schema syntax.  it is a valid character for a CRD field

miguelsorianod: it still seems odd to me seeing that in a CR field, but I have no objection to it so as you prefer :+1:

apis/capabilities/v1beta1/policy_types.go
```diff
@@ -0,0 +1,173 @@
...
+// PolicySchemaSpec defines the desired Policy schema
+type PolicySchemaSpec struct {
+	// Name is the name of the policy schema
+	Name string `json:"name"`
+
+	// Version is the version of the policy schema
+	Version string `json:"version"`
+
+	// Summary is the summary of the policy schema
+	Summary string `json:"summary"`
```
miguelsorianod, line 60: Is this field mandatory?

apis/capabilities/v1beta1/policy_types.go
miguelsorianod, line 66:
```
 	// Schema the $schema keyword is used to declare that this is a JSON Schema.
```

Can we refer to some doc on what values are possible for this? What value/s should I specify as this user? can we link to some doc about that?

miguelsorianod (approved):

apis/capabilities/v1beta1/policy_types.go
eguzki: Yes, otherwise, the API returns:

```
{"errors":{"schema":["The property '#/' did not contain a required property of 'summary' in schema http://apicast.io/policy-v1/schema#"]}}
```

miguelsorianod: :+1:

apis/capabilities/v1beta1/policy_types.go
eguzki: Adding a link to spec doc

PR #548, opened 2020-12-11
Define workers for production apicast

eguzki:
https://issues.redhat.com/browse/THREESCALE-6449
                                            
                                   
* Production apicast workers defined in the CR. Another option would be to add it in the CM. 
* Staging apicast does not have option to tune workers. As this parameter is meant for performance tuning, it should not be necessary.

example
```
kind: APIManager                                                        
metadata:                                                               
  name: apimanager1                                                     
spec:                                                                   
  wildcardDomain: example.com                     
  apicast:                                                              
    productionSpec:                                                     
      workers: 4
```


miguelsorianod:
The option has been added but if someone modifies it then no changes happen at all. It behaves like a deployment-only attribute.
Is that intentional? should we reconcile it? Same kind of scenario will happen with the other request we've had about the log level

eguzki:
Good point. I was just testing reconciliation and noticed that. I assumed env vars where reconciled (the apicast operator does, hence my mistake).

Should we reconcile specific env vars or the entire array? Maybe we can try reconciling the entire array but with a merge approach, which would allow adding custom env var and the operator would not revert it.

pkg/3scale/amp/operator/apicast_reconciler.go
```diff
@@ -125,6 +135,25 @@ func (r *ApicastReconciler) Reconcile() (reconcile.Result, error) {
 	return reconcile.Result{}, nil
 }
 
+func (r *ApicastReconciler) apicastProductionWorkersEnvVarMutator(desired, existing *appsv1.DeploymentConfig) bool {
```
eguzki, line 138: Not sure this is the best way to do this. wdyt @miguelsorianod ?

doc/apimanager-reference.md
```diff
@@ -102,6 +102,7 @@ Generated using [github-markdown-toc](https://github.com/ekalinin/github-markdow
 | Affinity | `affinity` | [v1.Affinity](https://v1-17.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#affinity-v1-core) | No | `nil` | Affinity is a group of affinity scheduling rules |
 | Tolerations | `tolerations` | \[\][v1.Tolerations](https://v1-17.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#toleration-v1-core) | No | `nil` | Tolerations allow pods to schedule onto nodes with matching taints |
 | Resources | `resources` | [v1.ResourceRequirements](https://v1-17.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#resourcerequirements-v1-core) | No | `nil` | Resources describes the compute resource requirements. Takes precedence over `spec.resourceRequirementsEnabled` with replace behavior |
+| Workers | `workers` | integer | No | auto | Defines the number of worker processes |
```
miguelsorianod, line 105: If `auto` it's a possible value then Workers is not really an integer.
The way for users to set it as `auto` would then be telling them "leave this empty if you want auto be used". In that case, would be great to clarify in the doc

pkg/reconcilers/deploymentconfig.go
```diff
@@ -13,71 +13,42 @@ import (
...
 )
 
-func DeploymentConfigResourcesMutator(existingObj, desiredObj common.KubernetesObject) (bool, error) {
-	existing, ok := existingObj.(*appsv1.DeploymentConfig)
-	if !ok {
-		return false, fmt.Errorf("%T is not a *appsv1.DeploymentConfig", existingObj)
+// DCMutateFn is a function which mutates the existing DeploymentConfig into it's desired state.
+type DCMutateFn func(desired, existing *appsv1.DeploymentConfig) bool
+
+func DeploymentConfigMutator(opts ...DCMutateFn) MutateFn {
```
miguelsorianod, line 19: I like this approach :+1:, makes easier adding different mutators in a single mutator

pkg/reconcilers/deploymentconfig.go
```diff
@@ -140,3 +108,30 @@ func DeploymentConfigContainerResourcesReconciler(desired, existing *appsv1.Depl
 
 	return update
 }
+
+func DeploymentConfigEnvVarMergeMutator(desired, existing *appsv1.DeploymentConfig) bool {
```
miguelsorianod, line 112: Not sure about this having the https://github.com/3scale/3scale-operator/pull/548 where it only reconciles the log envvar (https://github.com/3scale/3scale-operator/pull/557).
If we are not sure probably would be a good idea to be safe and just reconcile the worker for now. We can always change to reconcile everything later

pkg/3scale/amp/operator/apicast_reconciler.go
miguelsorianod: See https://github.com/3scale/3scale-operator/pull/548/files#r553952791 comment

miguelsorianod (changes requested):
:+1: . I left some comments

doc/apimanager-reference.md
eguzki: The value in the `Default value` field has the implicit meaning that the value will be set when the field is empty/unset. No need to state it explicitly. 

Updated "auto" for more meaningful content with link to apicast upstream doc.

pkg/reconcilers/deploymentconfig.go
eguzki: removing this method for another one focused in a single env var.

pkg/3scale/amp/operator/apicast_reconciler.go
eguzki: Updating for a new implementation based on reconciliation of a single env var

eguzki:
Answered all comments and applied changes, ready for a new review @miguelsorianod

doc/apimanager-reference.md

pkg/reconcilers/deploymentconfig.go
miguelsorianod: :+1:

pkg/3scale/amp/operator/apicast_reconciler.go
miguelsorianod: :+1:

miguelsorianod (approved):

PR #549, opened 2020-12-15
Improve APIManager status

miguelsorianod:
This PR adds a new condition in APIManager's Status section named `Available`.

We define the `Available` condition to `true` when ALL of the following scenarios are true:
* All expected DeploymentConfigs to be deployed exist and have the `Available` condition set to true
* All 3scale default OpenShift routes exist and have the `Admitted` condition set to true. The default routes are:
  * Master route
  * Backend Listener route
  * Default tenant admin route, developer route, APIcast staging and production routes beloinging to the default tenant

Design decisions / notable changes:

* A new `APIManagerStatusReconciler` reconciler has been created that is now used in the APIManager controller
* The previously defined conditions field in APIManager has been replaced by the `common.Conditions` type that we have available. This is a change in schema but if I'm not wrong this change in type has implied the addition of fields to the `conditions` field so this specific change should not be breaking compatibility.
* The upgrade process does not update the status section. This keeps the same behavior as before this PR.

Pending:

- [x] Documentation
- [x] Testing
- [x] Clean commit history


apis/apps/v1alpha1/apimanager_types.go
```diff
@@ -76,13 +80,39 @@ type APIManagerStatus struct {
...
+	// Marshalling sorts by condition type
+	currentMarshaledJSON, _ := s.Conditions.MarshalJSON()
+	otherMarshaledJSON, _ := other.Conditions.MarshalJSON()
+	if string(currentMarshaledJSON) != string(otherMarshaledJSON) {
+		diff := cmp.Diff(string(currentMarshaledJSON), string(otherMarshaledJSON))
+		logger.V(1).Info("Conditions not equal", "difference", diff)
+		return false
+	}
+
+	// TODO should we sort the DCs? when we set the status section we are
```
miguelsorianod, line 105: See this TODO

eguzki:
missing error reporting. Are you implementing it in this PR?

miguelsorianod:
> missing error reporting. Are you implementing it in this PR?

Do you mean `spec` validation implementation? If that's the case I was not intending to implement it in this PR.

controllers/apps/apimanager_controller.go
```diff
@@ -117,20 +113,23 @@ func (r *APIManagerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
 		return result, err
 	}
-	if result.Requeue {
-		logger.Info("Reconciling not finished. Requeueing.")
-		return result, nil
-	}
 
-	statusResult, err := r.reconcileAPIManagerStatus(instance)
-	if err != nil {
+	statusResult, statusErr := r.reconcileAPIManagerStatus(instance, err)
```
miguelsorianod, line 117: I'm unsure on the requeue/received errors management at this point between logic and status reconciliation. Let's iterate from this.

pkg/3scale/amp/component/deployments_lister.go
```diff
@@ -0,0 +1,52 @@
+package component
```
miguelsorianod, line 1: I created a new component that returns you the desired deployment names depending on the configuration set.

controllers/apps/apimanager_status_reconciler.go
```diff
@@ -0,0 +1,234 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type APIManagerStatusReconciler struct {
+	*reconcilers.BaseReconciler
+	apimanagerResource *appsv1alpha1.APIManager
+	syncError          error
```
miguelsorianod, line 27: Added to be prepared but not currently used

miguelsorianod: Finally I've removed this due to not being used. We can always add it in the future in case is needed.

controllers/apps/apimanager_controller.go
miguelsorianod: For now if there's a requeue or error in the logic reconciler we'll reconcile at that point. Same when status scenario is reached.

apis/apps/v1alpha1/apimanager_types.go
miguelsorianod: I've replaced the TODO by a comment indicating that at that point the DCs should be  sorted already

miguelsorianod:
> missing error reporting. Are you implementing it in this PR? 

I'll open an issue to investigate how we could improve in that aspect. I understand the idea is to further improve the status by checking usual errors that might happen (imagestreams cannot pull the image, etc...)

eguzki:
From https://issues.redhat.com/browse/THREESCALE-4753, this "state machine" would be a nice to have: 
```
Create event when all DC condition is available when current status is not available
Create event when any DC condition is unavailable when current status is available
```

miguelsorianod:
When testing the functionality the following scenario has arisen:

We check the readiness of the OCP Routes. Due to Zync-related routes are not created by the APIManager they don't have an APIManager OwnerReference. This means that we are not able to detect changes on routes, and this implies that the reconciler is not able to detect them and the state calculation cannot be done appropriately. The following scenario happens:
1. We deploy an APIManager
2. Eventually the APIManager Deployments and Routes end up being deployed appropriately and in a ready state but the last state that the APIManager receives is one corresponding to one of its Deployments. At that point in time of the event all DCs are already ready but not all routes. Due to we don't receive events related to Zync-related routes for the reasons explained above this change in status is not observed by the APIManager reconcile and it leaves the APIManager's availability condition permanently set to `false` which would be incorrect.

The following solutions have been thought and are not feasible:
* Watch all routes. The problem with this is that the event received is the name and namespace of the route and not the APIManager one
* Watch all routes that contain an APIManager OwnerReference with the `EnqueueRequestForOwner` method. The issue here is that Zync-related routes do not directly have an APIManager OwnerReference but an OwnerReference to Zync-Que DC

Possible solutions that might be feasible:
* System somehow adds the APIManager to which it belongs as an annotation or some information that from the operator we can take a look at. The issue with this is that it sort of ties APIManager knowledge in System
* We create our own handler that recursively resolves OwnerReferences. This happens before the Event is received and we would watch Route events with this handler and this handler would try to remap from Route objects to its APIManager object. Related to this approach there are several unknowns and the implementation could have several edge cases. We should be careful if we decide to implement this
* If we detect that routes are not ready then we reenqueue the request. The issue with this is that if all routes are ready and then some of them go into a "not ready" state we would not detect this either so we would end up having the same issue. There's also the issue on deciding WHEN should we reenqueue and how much delay. We have to keep in mind that we don't always want to wait when it's not ready

Maybe the best idea for now is to just take into account Deployments only and not consider routes. In this way we have something already and we can continue thinking about whether we want to include Routes and how. I'm not 100% sure though as I'm a little bit concerned that introducing this would be more problematic that if we don't due to users might report that "routes are not available but the 'Available' flag is set to true". At the end it will depend on how we define the Available flag and what expectations we set to the users / users have.

miguelsorianod:
I'm currently exploring the following possible solution that I commented in my previous reply:
```
We create our own handler that recursively resolves OwnerReferences. This happens before the Event is received and we would watch Route events with this handler and this handler would try to remap from Route objects to its APIManager object. Related to this approach there are several unknowns and the implementation could have several edge cases. We should be careful if we decide to implement this
```

miguelsorianod:
New changes:
* Incorporated changes from https://github.com/3scale/3scale-operator/pull/561 PR to detect zync-managed route changes events
* Added and updated some tests
* Documentation

miguelsorianod:
@eguzki I think maybe this could be good enough already?

Regarding your previous comment https://github.com/3scale/3scale-operator/pull/549#issuecomment-746183540, depending on what's the objective we want to achieve with that it might not be needed: The Condition provides the field `LastTransitionTime` which provides the last time the condition transit from one status to another.

miguelsorianod:
Examples:

APIManager status when it is not available:

eguzki:
About the generation of k8s events on:
```
Create event when all DC condition is available when current status is not available
Create event when any DC condition is unavailable when current status is available
```
The state can be the `status` field itself. No need to inspect the transition time.  

It can be done in another PR.

controllers/apps/apimanager_status_reconciler.go
```diff
@@ -0,0 +1,234 @@
...
+		routes = append(routes, route)
+	}
+	sort.Slice(routes, func(i, j int) bool { return routes[i].Name < routes[j].Name })
+
+	allDefaultRoutesReady := true
+	for _, routeHost := range routeHosts {
+		routeFound := false
+		var matchedRoute *routev1.Route
+		for _, route := range routes {
+			// TODO should we check that there are no two routes
```
eguzki, line 213: I think that openshift takes care of it. It is not allowed.

controllers/apps/apimanager_status_reconciler.go
```diff
@@ -0,0 +1,234 @@
...
+	if deploymentsAvailable && defaultRoutesReady {
+		newAvailableCondition.Status = v1.ConditionTrue
+	}
+
+	return newAvailableCondition, nil
+}
+
+func (s *APIManagerStatusReconciler) defaultRoutesReady() (bool, error) {
+	wildcardDomain := s.apimanagerResource.Spec.WildcardDomain
+	routeHosts := []string{
```
eguzki, line 183: `expectedRouteHosts` is more meaningful

controllers/apps/apimanager_status_reconciler.go
```diff
@@ -0,0 +1,234 @@
...
+			// TODO should we check that there are no two routes
+			// with the same host field value?
+			if route.Spec.Host == routeHost {
+				routeFound = true
+				matchedRoute = &route
+			}
+		}
+
+		if !routeFound {
+			allDefaultRoutesReady = false
```
eguzki, line 222: you can `break` the loop, right?

controllers/apps/apimanager_status_reconciler.go
```diff
@@ -0,0 +1,234 @@
...
+		}
+
+		if routeFound {
+			routeReady := helper.IsRouteReady(matchedRoute)
+			if !routeReady {
+				allDefaultRoutesReady = false
```
eguzki, line 228: you can `break` the loop, right?

In the future, we can build a list or errors (`errorList`) to report `which` route is not ready. The `allDefaultRoutesReady` would be when the length of the errList is 0.

doc/apimanager-reference.md
```diff
@@ -7,54 +7,56 @@ One APIManager custom resource per project is allowed.
 ## Table of Contents
 
 * [APIManager](#apimanager)
-   * [APIManagerSpec](#apimanagerspec)
```
eguzki, line 10: This change is generated with the tool? If so, I am ok with it. Otherwise, we should stick to the tool output to be more easy to manage.

pkg/3scale/amp/component/deployments_lister.go
```diff
@@ -0,0 +1,52 @@
...
+
+	if !d.ExternalRedisDatabases {
+		deployments = append(deployments,
+			BackendRedisDeploymentName,
+			SystemRedisDeploymentName,
+		)
+	}
+
+	if !d.ExternalZyncDatabase {
+		deployments = append(deployments)
```
eguzki, line 48: missing something, right?

controllers/apps/apimanager_status_reconciler.go
```diff
@@ -0,0 +1,234 @@
...
+	for _, route := range routeList.Items {
+		routes = append(routes, route)
+	}
+	sort.Slice(routes, func(i, j int) bool { return routes[i].Name < routes[j].Name })
+
+	allDefaultRoutesReady := true
+	for _, routeHost := range routeHosts {
+		routeFound := false
+		var matchedRoute *routev1.Route
+		for _, route := range routes {
```
eguzki, line 212: minor proposal:

add to ` pkg/helper/route.go` RouteFindByHost method:

```
// RouteFindByHost returns the smallest index i at which a route with a given host is found
// or -1 if there is no such index.                                 
func RouteFindByHost(a []routev1.Route, routeHost string) int {                      
    for i, n := range a {                                           
        if n.Spec.Host == routeHost {                                            
            return i                                                
        }                                                           
    }                                                               
    return -1                                                       
}                                                                   

```

eguzki (changes requested):

eguzki:
Looking good. Added some feedback

controllers/apps/apimanager_controller.go
```diff
@@ -122,15 +122,14 @@ func (r *APIManagerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
+	if statusErr != nil {
 		logger.Error(err, "Error updating status")
 		return ctrl.Result{}, err
 	}
-
 	if statusResult.Requeue {
-		logger.V(1).Info("Reconciling status not finished. Requeueing.")
-		return statusResult, nil
+		logger.Info("Reconciling not finished. Requeueing.")
+		return result, nil
```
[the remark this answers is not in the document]
miguelsorianod: good catch :+1:

controllers/apps/apimanager_controller.go
```diff
@@ -291,55 +296,11 @@ func (r *APIManagerReconciler) reconcileSystemMySQLLogic(cr *appsv1alpha1.APIMan
 }
 
 func (r *APIManagerReconciler) reconcileAPIManagerStatus(cr *appsv1alpha1.APIManager) (reconcile.Result, error) {
```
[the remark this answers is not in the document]
miguelsorianod: even though it only does that at the moment I tend to prefer having it into its own contained method so it does not deal with the reconciler on the Reconcile method itself

controllers/apps/apimanager_status_reconciler.go
miguelsorianod: :+1:

controllers/apps/apimanager_status_reconciler.go
miguelsorianod: :+1:

controllers/apps/apimanager_status_reconciler.go
miguelsorianod: You are correct.
The reason I didn't break the loop is exactly what you said. I left it prepared in case we want to add the list of routes.

doc/apimanager-reference.md
miguelsorianod: It was generated bye the tool yes

pkg/3scale/amp/component/deployments_lister.go
miguelsorianod: :scream: you are right. Fixed, thanks! :+1:

controllers/apps/apimanager_status_reconciler.go
miguelsorianod: :+1:  changed

controllers/apps/apimanager_status_reconciler.go
miguelsorianod: changed

controllers/apps/apimanager_controller.go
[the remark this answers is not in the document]
miguelsorianod: changed

miguelsorianod:
Changes applied. Ready for a new review

eguzki:
looking good

eguzki (approved):

PR #550, opened 2020-12-16
Fix Zync monitoring annotations location

miguelsorianod:
Zync DeploymentConfig had the following annotations defined as part of the DC annotations:
```
    annotations:
      prometheus.io/port: "9393"
      prometheus.io/scrape: "true"
```

They should belong to Zync's PodTemplate annotations and not Zync's DC annotations themselves.

This does not affect operator monitoring functionality because we use `PodMonitor` there but we should fix this anyway (it's provided in templates too and there there's no CRs)

This PR takes care of changing that.

Related issue: https://issues.redhat.com/browse/THREESCALE-6509

### DIFF
--- a/pkg/3scale/amp/operator/upgrade.go
+++ b/pkg/3scale/amp/operator/upgrade.go
@@ -41,6 +41,14 @@ func (u *UpgradeApiManager) Upgrade() (reconcile.Result, error) {
 		return res, nil
 	}
 
+	res, err = u.upgradeZyncPodTemplateAnnotations()
+	if err != nil {
+		return res, fmt.Errorf("Upgrading Zync DC PodTemplate: %w", err)
+	}
+	if res.Requeue {
+		return res, nil
+	}
+
 	res, err = u.upgradeImages()
 	if err != nil {
 		return res, fmt.Errorf("Upgrading images: %w", err)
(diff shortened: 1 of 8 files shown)


pkg/3scale/amp/operator/upgrade.go
```diff
@@ -523,6 +531,43 @@ func ensureBackendRouteEnvVar(desired v1.EnvVar, existingEnvVars *[]v1.EnvVar) b
...
+		return reconcile.Result{}, err
+	}
+
+	if existing.Annotations == nil {
+		existing.Annotations = map[string]string{}
+	}
+	update := false
+
+	for desiredAnnotationKey, desiredAnnotationVal := range desired.Annotations {
+		existingAnnotationVal, ok := existing.Annotations[desiredAnnotationKey]
```
eguzki, line 553: 1.- Existing "wrong" annotations will no be deleted from existing `metadata.annotations`
2.- Desired annotations will not be included in `spec.template.metadata.annotations`

eguzki (changes requested):

pkg/3scale/amp/operator/upgrade.go
miguelsorianod: this is still WIP (I had to test it yet) but you are right about 2 :+1: . It should be fixed now (pending testing). Thanks!

miguelsorianod: About 1... I hoped our reconciliation of the metadata annotations might fix it (/home/msoriano/go/src/github.com/3scale/3scale-operator/pkg/helper/object_meta_merger.go) but I think it'd only merge desired and existing so it would not remove the old ones. I'll take a look at it when testing it :+1:

miguelsorianod: About 1, Our object metamerger would not remove the old ones indeed.
I've updated the upgrade procedure to delete the old ones from `metadata.Annotations`.

Pending testing it/trying it

miguelsorianod: It seems to work ok

miguelsorianod:
Ready to be merged @eguzki

eguzki (approved):

PR #551, opened 2020-12-16
DeveloperAccount and DeveloperUser CRDs

eguzki:
CRD (and controller) for managing 3scale developer accounts.
CRD (and controller) for managing 3scale developer user accounts.

Requires https://github.com/3scale/3scale-porta-go-client/pull/40

- [x] CR samples
- [x] Usage doc / reference doc
- [x] CRD diagram update
- [x] User Suspend/Unsuspend
- [x] User Password from secret
- [x]  validation of user email


eguzki:
@miguelsorianod ready for initial review. Only doc is missing.

pkg/helper/boolean_utils.go
```diff
@@ -0,0 +1,17 @@
+package helper
+
+func Any(arr []bool) bool {
+	result := false
+	for _, v := range arr {
+		result = result || v
```
miguelsorianod, line 6: Is this intended to check whether at least one of the bools in the array is `true`? if so, we can just check it and return without needing to iterate all the array.
Same would happen with `All` but just checking the opposite.

apis/capabilities/v1beta1/developeraccount_types.go
```diff
@@ -0,0 +1,176 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+const (
+	DeveloperAccountKind = "DeveloperAccount"
```
miguelsorianod, line 35: As a side comment, conceptually this should really belong to one package level above as it's independent of the API version but we can leave it here

apis/capabilities/v1beta1/developeruser_types.go
```diff
@@ -0,0 +1,207 @@
...
+	// Username
+	Username string `json:"username"`
+
+	// Email
+	Email string `json:"email"`
+
+	// Password
+	PasswordCredentialsRef corev1.SecretReference `json:"passwordCredentialsRef"`
+
+	// DeveloperAccountRef is the reference to the parent developer account
```
miguelsorianod, line 76: This means we only support specifying developer accounts that have been created through the capabilities functionality right?

controllers/capabilities/developeraccount_status_reconciler.go
```diff
@@ -0,0 +1,171 @@
...
+		ID:                  s.resource.Status.ID,
+		AccountState:        s.resource.Status.AccountState,
+		CreditCardStored:    s.resource.Status.CreditCardStored,
+		ProviderAccountHost: s.resource.Status.ProviderAccountHost,
+		Conditions:          s.resource.Status.Conditions.Copy(),
+		ObservedGeneration:  s.resource.Status.ObservedGeneration,
+	}
+
+	// If there is an error and s.remoteDeveloperAccount is nil
+	// ID is not changed, as it is initialized with previous values d
```
miguelsorianod, line 90: Is the comment incomplete?

I' trying to understand this use case

controllers/capabilities/developeraccount_threescale_reconciler.go
```diff
@@ -0,0 +1,235 @@
...
+	desiredMonthlyBillingEnabled := true
+	if s.resource.Spec.MonthlyBillingEnabled != nil {
+		desiredMonthlyBillingEnabled = *s.resource.Spec.MonthlyBillingEnabled
+	}
+	if devAccount.Element.MonthlyBillingEnabled != nil && *devAccount.Element.MonthlyBillingEnabled != desiredMonthlyBillingEnabled {
+		update = true
+		deltaAccount.Element.MonthlyBillingEnabled = &desiredMonthlyBillingEnabled
+	}
+
+	// MonthlyChargingEnabled defaults to False
```
miguelsorianod, line 186: Is the comment outdated or is the value incorrect?

miguelsorianod (commented):
It looks good :+1: 

I left some comments

pkg/helper/boolean_utils.go
eguzki: right, this is an alternative implementation. But both approaches work, right?

apis/capabilities/v1beta1/developeruser_types.go
eguzki: Right. 

Would be more flexible to make reference to Account ID? I think we should be generic and make references to other CR names or unique fields in the CR (like system_name in other CRD's). But I do not have strong opinion.

controllers/capabilities/developeraccount_status_reconciler.go
eguzki: This comment does not make sense here, the message was added few lines above. 

The use case is that `status.ID` has been set correctly. In later reconciliation loop, there is some unexpected error and `s.remoteDeveloperAccount` (the 3scale developer account object read from the API) can be nil. Therefore, the `status.ID` is not changed to nil, instead, the existing value remains untouched.

controllers/capabilities/developeraccount_threescale_reconciler.go
eguzki: fixed

eguzki:
:+1: I am going to write the doc

pkg/helper/boolean_utils.go
miguelsorianod: both should work yes. I was pointing there was no need to iterate all array

apis/capabilities/v1beta1/developeruser_types.go
miguelsorianod: I don't have a strong opinion about it either.

config/crd/patches/webhook_in_developerusers.yaml
```diff
@@ -0,0 +1,17 @@
+# The following patch enables conversion webhook for CRD
+# CRD conversion requires k8s 1.13 or later.
+apiVersion: apiextensions.k8s.io/v1beta1
```
miguelsorianod, line 3: The APIVersion i for `webhook_*` files and `cainjection_*` files should be `v1`

miguelsorianod (changes requested):
Doc looks good to me :+1: .

There's one pending change regarding APIversion of some files

config/crd/patches/webhook_in_developerusers.yaml
eguzki: fixed

miguelsorianod (approved):

PR #552, opened 2020-12-17
fix upgrade when monitoring is disabled

eguzki:
When monitoring was disabled, the operator still tried to fetch monitoring resources, which were expected to exist. 

This PR frees the upgrade procedure from upgrading monitoring resources. 

However, the operator still upgrades deployment config env vars for monitoring ports (for system components)

### DIFF
--- a/pkg/3scale/amp/operator/upgrade.go
+++ b/pkg/3scale/amp/operator/upgrade.go
@@ -530,6 +530,10 @@ func (u *UpgradeApiManager) ensureSystemAppMonitoringSettings() (bool, error) {
 }
 
 func (u *UpgradeApiManager) upgradePrometheusRules(desired *monitoringv1.PrometheusRule) (bool, error) {
+	if !u.apiManager.IsMonitoringEnabled() {
+		return false, nil
+	}
+
 	existing := &monitoringv1.PrometheusRule{}
 	err := u.Client().Get(context.TODO(), types.NamespacedName{Name: desired.Name, Namespace: u.apiManager.Namespace}, existing)
 	if err != nil {


pkg/3scale/amp/operator/upgrade.go
```diff
@@ -530,6 +530,10 @@ func (u *UpgradeApiManager) ensureSystemAppMonitoringSettings() (bool, error) {
 }
 
 func (u *UpgradeApiManager) upgradePrometheusRules(desired *monitoringv1.PrometheusRule) (bool, error) {
+	if !u.apiManager.IsMonitoringEnabled() {
+		return false, nil
+	}
```
gsaslis, line 535: @eguzki apologies if this seems obvious, but seeing as I'm not at all acquainted with the codebase here: 

i see this changes the behaviour of `upgradePrometheusRules` which I presume crashed with some error when monitoring was disabled. 

am I right in understanding the bug didn't allow control flow to reach [here](http://github.com/3scale/3scale-operator/blob/5b54f6adffc48785d35d444649ac1169d0dc26c8/pkg/3scale/amp/operator/upgrade.go#L442-L442), because of some error [here](http://github.com/3scale/3scale-operator/blob/5b54f6adffc48785d35d444649ac1169d0dc26c8/pkg/3scale/amp/operator/upgrade.go#L410-L410) ?

eguzki: > am I right in understanding the bug didn't allow control flow to reach here, because of some error here ?
Yes. That is what happened before the fix. The root issue is that [here](https://github.com/3scale/3scale-operator/blob/5b54f6adffc48785d35d444649ac1169d0dc26c8/pkg/3scale/amp/operator/upgrade.go#L538), the operator tries to read a k8s resource that expects to exist to upgrade. The operation fails when the object does not exist. That expectation is wrong when monitoring is disabled.

gsaslis: ok, thanks, yes. 👍

gsaslis:
@eguzki code change makes sense. 

I see some failure in CI - is that normal?

gsaslis (approved):

eguzki:
no, CI has to pass. Sometimes (more often than wanted), it fails because of this System [issue](https://issues.redhat.com/browse/THREESCALE-6391). This is currently happening. I am re-running the tests and should work

eguzki:
It is failing because e2e tests uses `nightly` images and [this PR in Porta](https://github.com/3scale/porta/pull/2268) breaks master with operator 2.10.  3scale operator has already been [adapted to that change](https://github.com/3scale/3scale-operator/pull/536), but 2.10 operator cannot deploy current porta nightly image.

gsaslis:
@eguzki so we need to use a "2.10" porta image, instead of `nightly` right? could we produce one manually and change the e2e tests on the 2.10 branch for that?

eguzki:
Yes, we need to merge this https://github.com/3scale/3scale-operator/pull/553 first

PR #553, opened 2020-12-17
E2e test from CR1 tag

eguzki:
* e2e test from tag name
* fix circleci job to build images from any tag

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -225,16 +225,20 @@ commands:
             done
 
   deploy-3scale-eval-from-template:
+    parameters:
+      components_tag_name:
+        default: "nightly"
+        type: string
     steps:
       - run:
           name: Deploy 3scale from amp-eval template with nightly images
           command: |
             WILDCARD_DOMAIN="lvh.me"
             oc new-app --file pkg/3scale/amp/auto-generated-templates/amp/amp-eval.yml \
-              --param AMP_BACKEND_IMAGE=quay.io/3scale/apisonator:nightly \
-              --param AMP_ZYNC_IMAGE=quay.io/3scale/zync:nightly \
-              --param AMP_APICAST_IMAGE=quay.io/3scale/apicast:nightly \
-              --param AMP_SYSTEM_IMAGE=quay.io/3scale/porta:nightly \
+              --param AMP_BACKEND_IMAGE=quay.io/3scale/apisonator:<< parameters.components_tag_name >> \
+              --param AMP_ZYNC_IMAGE=quay.io/3scale/zync:<< parameters.components_tag_name >> \
+              --param AMP_APICAST_IMAGE=quay.io/3scale/apicast:<< parameters.components_tag_name >> \
+              --param AMP_SYSTEM_IMAGE=quay.io/3scale/porta:<< parameters.components_tag_name >> \
               --param MEMCACHED_IMAGE=registry.redhat.io/3scale-amp2/memcached-rhel7:3scale2.9 \
               --param WILDCARD_DOMAIN=lvh.me --param TENANT_NAME=3scale
           no_output_timeout: 30m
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -254,10 +263,12 @@ commands:
...
-            oc image mirror $(for component in apicast zync ; do
-              echo 172.30.1.1:5000/$project/amp-$component:${imagestream_tag_name}=quay.io/3scale/$component:nightly
-            done) 172.30.1.1:5000/$project/amp-backend:${imagestream_tag_name}=quay.io/3scale/apisonator:nightly \
-            172.30.1.1:5000/$project/amp-system:${imagestream_tag_name}=quay.io/3scale/porta:nightly --insecure
+            oc image mirror \
+              172.30.1.1:5000/$project/amp-apicast:${imagestream_tag_name}=quay.io/3scale/apicast:${imagestream_tag_name} \
+              172.30.1.1:5000/$project/amp-zync:${imagestream_tag_name}=quay.io/3scale/zync:${imagestream_tag_name} \
+              172.30.1.1:5000/$project/amp-backend:${imagestream_tag_name}=quay.io/3scale/apisonator:${imagestream_tag_name} \
+              172.30.1.1:5000/$project/amp-system:${imagestream_tag_name}=quay.io/3scale/porta:${imagestream_tag_name} \
+              --insecure
```
gsaslis, line 271: appreciate the better formatting and simpler syntax here 👏

.circleci/config.yml
```diff
@@ -117,13 +117,18 @@ commands:
...
       - run:
           name: Deploy 3scale from amp-eval template without imagestreams
           command: |
             WILDCARD_DOMAIN="lvh.me"
-            ruby -ryaml -rjson -e 'puts YAML.load(ARGF).tap{|t| t["objects"].reject!{|o| o["kind"]=="ImageStream"}}.to_json' pkg/3scale/amp/auto-generated-templates/amp/amp-eval.yml | \
-              oc new-app -f- --param WILDCARD_DOMAIN=${WILDCARD_DOMAIN}
+            imagestream_tag_name=<< parameters.components_imagestream_tag_name >>
+            oc new-app -f pkg/3scale/amp/auto-generated-templates/amp/amp-eval.yml -o json \
+              --param WILDCARD_DOMAIN=$WILDCARD_DOMAIN --param AMP_RELEASE="${imagestream_tag_name}" | \
+              jq -j '.items[] | select(.kind != "ImageStream")' | oc create -f -
```
gsaslis, line 131: +1 on no longer needing ruby to run this task. 👍

.circleci/config.yml
```diff
@@ -117,13 +117,18 @@ commands:
...
     steps:
       - run:
           name: Deploy 3scale from amp-eval template without imagestreams
           command: |
             WILDCARD_DOMAIN="lvh.me"
-            ruby -ryaml -rjson -e 'puts YAML.load(ARGF).tap{|t| t["objects"].reject!{|o| o["kind"]=="ImageStream"}}.to_json' pkg/3scale/amp/auto-generated-templates/amp/amp-eval.yml | \
-              oc new-app -f- --param WILDCARD_DOMAIN=${WILDCARD_DOMAIN}
+            imagestream_tag_name=<< parameters.components_imagestream_tag_name >>
+            oc new-app -f pkg/3scale/amp/auto-generated-templates/amp/amp-eval.yml -o json \
+              --param WILDCARD_DOMAIN=$WILDCARD_DOMAIN --param AMP_RELEASE="${imagestream_tag_name}" | \
```
gsaslis, line 130: may be better to keep the `${WILDCARD_DOMAIN}` form?

.circleci/config.yml
```diff
@@ -294,10 +305,17 @@ commands:
               --param AMP_ZYNC_IMAGE=quay.io/3scale/zync:nightly \
               --param AMP_APICAST_IMAGE=quay.io/3scale/apicast:nightly \
               --param AMP_SYSTEM_IMAGE=quay.io/3scale/porta:nightly \
+              --param MEMCACHED_IMAGE=registry.redhat.io/3scale-amp2/memcached-rhel7:3scale2.9 \
```
gsaslis, line 308: should we extract the `2.9` in some param / global var, to make it a little more prominent (so we don't forget to update it? when 2.10 is ga)

...not that it really matters for this particular image though... still ok if you don't want to change.

gsaslis (approved):
as far as i can tell, this looks good 👍

.circleci/config.yml
eguzki: Updating it. Makes all the sense for the master branch. Less important for this stable branch, but still I will implement it

PR #554, opened 2020-12-17
Generalized circleci e2e tests and builds for any tag

eguzki:
Generalize circleci tests for:
* e2e test to be passed for any 3scale component tag
* 3scale image builds for any 3scale component tag

Useful in (future) stable branches to specify tag for e2e tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 
 ##################################### YAML ANCHORS  ############################################
 
-tag-triggerr: &tag-trigger
+tag-trigger: &tag-trigger
   tags:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,8 +1,23 @@
+---
 version: 2.1
 
+parameters:
```
miguelsorianod, line 4: The idea is that in each branch we would change this parameters right?

miguelsorianod (approved):
:+1:

.circleci/config.yml
eguzki: exactly!

PR #556, opened 2020-12-21
Add more labels to monitoring resources

eguzki:
https://issues.redhat.com/browse/THREESCALE-6496

- [x] apicast monitoring resources
- [x] generic monitoring resources
- [x] backend monitoring resources
- [x] system monitoring resources
- [x] zync monitoring resources

The PR adds more labels to existing list of labels. Thus, upgrade procedure is not needed, as first reconciliation loop will add missing labels.

### DIFF
--- a/pkg/3scale/amp/component/system_monitoring.go
+++ b/pkg/3scale/amp/component/system_monitoring.go
@@ -76,39 +76,34 @@ func (system *System) SystemAppPodMonitor() *monitoringv1.PodMonitor {
 	}
 }
 
-func SystemGrafanaDashboard(ns string) *grafanav1alpha1.GrafanaDashboard {
+func (system *System) SystemGrafanaDashboard() *grafanav1alpha1.GrafanaDashboard {
 	data := &struct {
 		Namespace string
 	}{
-		ns,
+		system.Options.Namespace,
 	}
 	return &grafanav1alpha1.GrafanaDashboard{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: "system",
-			Labels: map[string]string{
-				"monitoring-key": common.MonitoringKey,
-			},
+			Name:   "system",
+			Labels: system.monitoringLabels(),
 		},
 		Spec: grafanav1alpha1.GrafanaDashboardSpec{
 			Json: assets.TemplateAsset("monitoring/system-grafana-dashboard-1.json.tpl", data),
-			Name: fmt.Sprintf("%s/system-grafana-dashboard-1.json", ns),
+			Name: fmt.Sprintf("%s/system-grafana-dashboard-1.json", system.Options.Namespace),
 		},
 	}
 }
 
-func SystemAppPrometheusRules(ns string) *monitoringv1.PrometheusRule {
+func (system *System) SystemAppPrometheusRules() *monitoringv1.PrometheusRule {
 	return &monitoringv1.PrometheusRule{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: "system-app",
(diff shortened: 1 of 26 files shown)


pkg/3scale/amp/template/adapters/apicast.go
```diff
@@ -111,6 +111,9 @@ func (a *Apicast) options() (*component.ApicastOptions, error) {
 	ao.StagingPodTemplateLabels = a.stagingPodTemplateLabels()
 	ao.ProductionPodTemplateLabels = a.productionPodTemplateLabels()
 
+	// Currently, only used for monitoring resources. Thus, it does not apply to templates.
+	ao.Namespace = "mynamespace"
```
miguelsorianod, line 115: Maybe the `${NAMESPACE}` value would work? I know the field is not being used right now in Templates but if that approach works it's probably better we set it already.

I've seen this value used on the following PR: https://github.com/openshift/origin/pull/16021/
I've never used it or checked if it exists in the templates but it seems something was done in some version related to that


`

pkg/3scale/amp/component/system_monitoring.go
```diff
@@ -173,3 +165,26 @@ func SystemSidekiqPrometheusRules(ns string) *monitoringv1.PrometheusRule {
...
+}
+
+func (system *System) prometheusRulesMonitoringLabels() map[string]string {
+	labels := make(map[string]string)
+
+	for key, value := range system.Options.CommonLabels {
+		labels[key] = value
+	}
+
+	labels["prometheus"] = "application-monitoring"
```
miguelsorianod, line 187: What's the purpose of this additional labels? Depending on it I'm wondering whether namespacing the label would be more appropriate.

I know the "monitoring-key" cannot be namespaced as we don't have control over it but if the `prometheus` and `role` labels are labels what we are arbitrarily defining here and are just informational maybe we could namespace them.

What do you think?

miguelsorianod (changes requested):
I left some comments. It looks good :+1:

pkg/3scale/amp/component/system_monitoring.go
eguzki: Those are labels already existing in the first implementation of the prometheus rule objects. TBH, I cannot tell the reason to be there or who is using it. But I would not change it, unless there is a good reason to do so.

miguelsorianod: :+1: makes sense. We can take a look at it in the future if we want to improve this or figure it out

pkg/3scale/amp/component/system_options.go
```diff
@@ -92,6 +92,8 @@ type SystemOptions struct {
 	IncludeOracleOptionalSettings bool
 
 	BackendServiceEndpoint string `validate:"required"`
+
+	Namespace string `validate:"required"`
```
miguelsorianod, line 96: A thought about this, I wonder whether in the future seeing this `Namespace` generic option (in all options components) might be confusing to people looking at the components code (like, "why is the Namespace an option and only being used in monitoring objects?"). Right now it is only used for monitoring-related objects. All the other objects of the components do not make use of this namespace because the Namespace is set outside.

miguelsorianod (approved):

pkg/3scale/amp/component/system_options.go
eguzki: Maybe it deserves a comment. Objects created are namespaced. However, objects includes labels, rules and expressions that need namespace filtering because they are "global" once imported to the prometheus or grafana services.

PR #558, opened 2020-12-22
Upgrade containerd dep cve 2020 15257

eguzki:
Upgrading to [containerd v1.3.9](https://github.com/containerd/containerd/releases/tag/v1.3.9) which includes a security release to address CVE-2020-15257

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -63,3 +63,5 @@ replace github.com/openshift/api => github.com/openshift/api v0.0.0-202005271843
 replace github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20191125132246-f6563a70e19a
 
 replace github.com/operator-framework/operator-sdk => github.com/operator-framework/operator-sdk v0.15.2
+
+replace github.com/containerd/containerd => github.com/containerd/containerd v1.3.9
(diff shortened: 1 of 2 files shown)


miguelsorianod:
Are we sure we actually affected by the CVE? A comment in the CVE says:
```
If you are not providing the ability for untrusted users to start
containers in the same network namespace as the shim (typically the
"host" network namespace, for example with `docker run --net=host` or
`hostNetwork: true` in a Kubernetes pod) and run with an effective UID
of 0, you are not vulnerable to this issue.
```

So it seemed to me we are not making use of the functionality of the issue described in the CVE. See https://bugzilla.redhat.com/show_bug.cgi?id=1899487 for details on comments.

Overwriting indirect dependencies of our dependencies does not seem a very good idea, how do we know by doing that we won't be breaking something in our dependencies? If we can avoid it I would wait until our indirect dependencies update their dependencies and then we can update our dependency. Even in that case, if we don't make use of this functionality maybe we would not even need to do that.

eguzki:
1) it is a patch update. It is safe to assume it is going to keep working as expected.
2) We do not know what intermediate dependencies are using from this library with CVE detected. 
3) I agree that doing this kind of "replacements" is not desired. It works as long as we only have up to a handful of replacements. Otherwise, it gets unmanageable. The alternative is to tell CVE detecting tooling that we are not using something and we are safe. First, we do not know for sure. Second, I find easier to update our dependencies and the CVE tooling will be happy, even if it is true that we are already safe. 

I think we have to go on with this change.

miguelsorianod (approved):

PR #560, opened 2021-01-12
development mode running controlled by flag

eguzki:
* `make run` will run in development mode with DEBUG log enabled
* operator deployed by deployment manifests will not run in dev mode.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,7 @@ manager: generate fmt vet
 run: export WATCH_NAMESPACE=$(LOCAL_RUN_NAMESPACE)
 run: export THREESCALE_DEBUG=1
 run: generate fmt vet manifests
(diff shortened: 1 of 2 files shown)


miguelsorianod:
I think instead of using this custom flag and the DevMode we should take advantage of the flags zap provides automatically for us by adding them and set them appropriately in each of our cases as desired: https://v1-2-x.sdk.operatorframework.io/docs/building-operators/golang/references/logging/#a-simple-example.

If you prefer it I can open a PR adding it.

eguzki:
Much better using operator-sdk flags. I will update

eguzki:

miguelsorianod (approved):
:+1: great. Thanks!

PR #566, opened 2021-01-15
Add configurable noreply FROM mail address

miguelsorianod:
Make the 'from' address value of the noreply mail address configurable.
This is an operator-only functionality.

In system, the `NOREPLY_EMAIL` environment variable is used to set the 'from' address of the noreply mail. The behavior in system with this environment variable currently is:
* If the environment variable is defined then its value is used. If it is defined as the empty string `""` that value is literally used instead of a default one
* If the environment variable is not defined then a default value specified by system is used.

All SMTP-related settings are stored in the `system-smtp` Secret. This PR adds the ability to
set the field `from_address` in the `system-smtp` secret. The field value must be set before deploying the APIManager in case the user wants a non-default behavior for that field, as with all other non-default field values in secrets.

As with other fields in secrets, it is a creation-only field and it is not automatically reconciled when changed once set.

* We only set System's environment variable when the from address secret field is set AND it is set to a non-empty value.
* By default, we explicitly set the from address field into the 'smtp' secret with the vale "" to be explicit and consistent with the other fields. From the point of view of the user "" means "use the default value" as with the other fields and the user can explicitly set it too with that meaning.
* An upgrade process is not needed for this due to secret defaults are automatically reconciled during the usual reconcile process.
* This is an operator-only functionality: On templates no from address field can be set on the secret nor the environment variable is set in any way (this is also the reason in System options component we still define the smtp from address field as a pointer, to avoid setting it into the secret).

- [x] Documentation
- [x] Testing

### DIFF
--- a/pkg/3scale/amp/component/system.go
+++ b/pkg/3scale/amp/component/system.go
@@ -85,6 +85,7 @@ const (
 	SystemSecretSystemSMTPPortFieldName              = "port"
 	SystemSecretSystemSMTPAuthenticationFieldName    = "authentication"
 	SystemSecretSystemSMTPOpenSSLVerifyModeFieldName = "openssl.verify.mode"
+	SystemSecretSystemSMTPFromAddressFieldName       = "from_address"
 )
 
 const (
@@ -148,6 +149,11 @@ func (system *System) getSystemSMTPEnvsFromSMTPSecret() []v1.EnvVar {
 		helper.EnvVarFromSecret("SMTP_PORT", SystemSecretSystemSMTPSecretName, SystemSecretSystemSMTPPortFieldName),
 		helper.EnvVarFromSecret("SMTP_AUTHENTICATION", SystemSecretSystemSMTPSecretName, SystemSecretSystemSMTPAuthenticationFieldName),
 		helper.EnvVarFromSecret("SMTP_OPENSSL_VERIFY_MODE", SystemSecretSystemSMTPSecretName, SystemSecretSystemSMTPOpenSSLVerifyModeFieldName),
+	}
+
+	if system.Options.SmtpSecretOptions.FromAddress != nil &&
+		*system.Options.SmtpSecretOptions.FromAddress != "" {
+		result = append(result, helper.EnvVarFromSecret("NOREPLY_EMAIL", SystemSecretSystemSMTPSecretName, SystemSecretSystemSMTPFromAddressFieldName))
 	}
 
 	return result
@@ -1043,7 +1049,7 @@ func (system *System) MemcachedService() *v1.Service {
 }
 
 func (system *System) SMTPSecret() *v1.Secret {
-	return &v1.Secret{
+	res := &v1.Secret{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Secret",
 			APIVersion: "v1",
@@ -1062,6 +1068,12 @@ func (system *System) SMTPSecret() *v1.Secret {
 			SystemSecretSystemSMTPUserNameFieldName:          *system.Options.SmtpSecretOptions.Username,
 		},
 	}
+
+	if system.Options.SmtpSecretOptions.FromAddress != nil {
+		res.StringData[SystemSecretSystemSMTPFromAddressFieldName] = *system.Options.SmtpSecretOptions.FromAddress
+	}
+
+	return res
 }
 
 func (system *System) SystemConfigMap() *v1.ConfigMap {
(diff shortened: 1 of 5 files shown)


miguelsorianod:
@eguzki ready for an initial review.

@guicassolato please take a look also. I'm currently not sure whether the field name is appropriate to convey the meaning of what's being set (maybe there are other possible "from addresses" for other mails potentially configurable and having "from_address" is not specific enough?). Also, could you confirm that it makes sense for the field to be in the same context as the other fields in 'system-smtp' secret? See the secret reference: https://github.com/3scale/3scale-operator/blob/master/doc/apimanager-reference.md#system-smtp

pkg/helper/secretutils.go
```diff
@@ -97,6 +97,24 @@ func NewSecretSource(client client.Client, namespace string) *SecretSource {
 	}
 }
 
+// GetField returns the value of the secret field named fieldName in
+// the Secret named secretName.
+// If the field fieldName does not exist in the secret secretName or the secret
+// secretName does not exists a nil string is returned.
+// If there's an error processing the SecretSource an error is returned.
+func (s *SecretSource) GetField(secretName, fieldName string) (*string, error) {
```
eguzki, line 105: Try to re-use `func (s *SecretSource) fieldReader(secretName, fieldName string, secretRequired, fieldRequired bool, def string) (string, error)` instead of re implementing. 


Your use case is about having different logic when the secret (or the secret-key) does not exist and when the value exists. For that regard, call `RequiredFieldValueFromRequiredSecret` method and check for errors about a) secret not found and b) field value not found. If any of those errors happen, you know you shouldn't add the env var. If any other error type is found, then report back and done. It would require raising some known specific type of error [here](https://github.com/3scale/3scale-operator/blob/f504a03941ec566e301a146132b3f02575caee62/pkg/helper/secretutils.go#L126) in order to check later, like
```
if errors.IsNotFound(err) || helper.IsSecretKeyNotFound(err) {
// do not add the env var
}
```


An alternative implementation is using the empty string as the default value `""`. It's valid approach because the empty string is not a valid FROM field value, hence if you get the default value, you know you should not set the env var. With this approach, if the customer sets the empty value, the value will not be propagated to system. This is the main difference with the previous approach.

eguzki (changes requested):

pkg/helper/secretutils.go
miguelsorianod: I prefer the later proposed approach.

I've updated the PR with a new commit to change the behavior. If it looks ok I'll squash it.
The new behavior is:
* We only set System's environment variable when the from address secret field is set AND it is set to a non-empty value.
* By default, we explicitly set the from address field into the 'smtp' secret with the vale `""` to be explicit and consistent with the other fields. From the point of view of the user `""` means "use the default value" as with the other fields and the user can explicitly set it too with that meaning. An upgrade process is not needed for this due to secret defaults are automatically reconciled during the usual reconcile process.
* This is an operator-only functionality: On templates no from address field can be set on the secret nor the environment variable is set in any way (this is also the reason in System options component we still define the smtp from address field as a pointer, to avoid setting it into the secret).

@eguzki let me know about the new changes

eguzki:
Looking good. Proceed with the doc and tests

miguelsorianod:
:+1: thanks @eguzki.

Before going with the tests and doc I'd like the feedback from @guicassolato so I don't need to apply potential changes in the doc/tests later. Could you review my previous comments? thanks!

pkg/3scale/amp/component/system.go
```diff
@@ -150,6 +151,11 @@ func (system *System) getSystemSMTPEnvsFromSMTPSecret() []v1.EnvVar {
 		helper.EnvVarFromSecret("SMTP_OPENSSL_VERIFY_MODE", SystemSecretSystemSMTPSecretName, SystemSecretSystemSMTPOpenSSLVerifyModeFieldName),
 	}
 
+	if system.Options.SmtpSecretOptions.FromAddress != nil &&
+		*system.Options.SmtpSecretOptions.FromAddress != "" {
```
guicassolato, line 155: The description of the PR says

> If the field is set in the secret then the `NOREPLY_EMAIL` is sent to system, even if its value is the empty string

but it looks like it won't set the env var if the value in the secret is the empty string after all.

guicassolato (approved):
IDK if planning to add a test for the non-default cases, but behaviour looks good to me.

pkg/3scale/amp/component/system.go
miguelsorianod: hi @guicassolato ,

Sorry, that initial comment is outdated :( . See https://github.com/3scale/3scale-operator/pull/566#discussion_r561932141

miguelsorianod:
> IDK if planning to add a test for the non-default cases, but behaviour looks good to me.

Yeah, testing and documentation was pending until we are sure the proposed solution makes sense. I'll work on it as soon as it's clear.

Thanks for reviewing it. My ask to you was about appropriate variable naming and secret placement more than the code itself. See this comment https://github.com/3scale/3scale-operator/pull/566#issuecomment-760915437

pkg/3scale/amp/component/system.go
guicassolato: What's implemented is better than what's described anyway 🙂 Nice job!

miguelsorianod:
@eguzki I think this should be ready

eguzki (approved):

PR #569, opened 2021-01-27
APIManager e2e tests with S3

miguelsorianod:
This changes the E2E tests execution to deploy an APIManager using System's FileStorage as S3 instead of a PVC.
The motivation of this change is to be able to deploy them in a cluster environment that does not provide RWX volumes, which is required by System's FileStorage.

The provided S3 data for the E2E tests is fake data, as it's not needed for the kind of tests we are performing at the moment. In case there's some test functionality that we end up doing that exercises S3 then additional changes will need to be done or change the provided approach.

### DIFF
--- a/controllers/apps/apimanager_controller_test.go
+++ b/controllers/apps/apimanager_controller_test.go
@@ -7,12 +7,14 @@ import (
 	"time"
 
 	appsv1alpha1 "github.com/3scale/3scale-operator/apis/apps/v1alpha1"
+	"github.com/3scale/3scale-operator/pkg/3scale/amp/component"
 	"github.com/google/uuid"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	appsv1 "github.com/openshift/api/apps/v1"
 	routev1 "github.com/openshift/api/route/v1"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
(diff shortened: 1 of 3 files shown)


miguelsorianod:
Additionally the tests have been modified to retry the `envTest` start. The reason for that is that sometimes it failed due to CRDs tried to be updated at the same time by the test suites of the capabilities and apps packages and showed the error:
```
[37m/go/src/github.com/3scale/3scale-operator/controllers/capabilities/suite_test.go:53[0m
  [91mUnexpected error:
      <*errors.StatusError | 0xc0010181e0>: {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "Operation cannot be fulfilled on customresourcedefinitions.apiextensions.k8s.io \"apimanagers.apps.3scale.net\": the object has been modified; please apply your changes to the latest version and try again",
              Reason: "Conflict",
              Details: {
                  Name: "apimanagers.apps.3scale.net",
                  Group: "apiextensions.k8s.io",
                  Kind: "customresourcedefinitions",
                  UID: "",
                  Causes: nil,
                  RetryAfterSeconds: 0,
              },
              Code: 409,
          },
      }
      Operation cannot be fulfilled on customresourcedefinitions.apiextensions.k8s.io "apimanagers.apps.3scale.net": the object has been modified; please apply your changes to the latest version and try again
  occurred[0m
```

eguzki (approved):
Not sure that error is due to concurrent processing of different test suites. Ginkgo has to be run with `-p` flag for parallel running and the test do not add that flag:
```
USE_EXISTING_CLUSTER=true $(GO) test $(TEST_E2E_PKGS) -coverprofile cover.out -ginkgo.v -ginkgo.progress -v -timeout 0
```

Anyway, that retry approach should not be harmful

miguelsorianod:
if you take a look at CircleCI of a test I did when removing the CRDs https://app.circleci.com/pipelines/github/3scale/3scale-operator/3201/workflows/49a6e6d4-cea1-471d-bf6a-288840f91bf3/jobs/23546 you'll observe that it seems they are being executed on parallel: you'll see that one of the tests fails trying to interact with APIManager due to the CRD not being there.

PR #588, opened 2021-03-02
Fix go bindata assets

eguzki:
There is no magic. assets have to be generated with the go-bindata tool. The asset generation is added to the `generate` makefile target.

In this PR I learnt that the generated code does not change each time you execute go-bindata. There will be changes in the generated code if template file modification times or content is changed. The file modification times are changed each time the project is cloned. In order to avoid this behavior, adding `-nometadata` to the generate command to not preserve size, mode, and modtime info

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -565,7 +565,6 @@ jobs:
     steps:
       - checkout
       - install-operator-sdk
-      - run: make
       - run: make assets-update-test
 
 workflows:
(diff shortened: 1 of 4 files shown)


miguelsorianod (approved):

miguelsorianod:
Approved, but tests are failing due to generating the assets modifies the timestamp.

eguzki:
adding -nometadata to the generate command to not preserve size, mode, and modtime info

miguelsorianod (approved):

PR #589, opened 2021-03-02
Unit test coverage github.com/3scale/3scale-operator/pkg/controller/helper

eguzki:
Added unittest coverage for `github.com/3scale/3scale-operator/pkg/controller/helper`

Initially, unittest coverage was **33.8%**

```
$ make coverage_total_report
....
33.8%
```

After this PR, **42.4%**

```
$ make coverage_total_report
....
42.4%
```

### DIFF
--- a/pkg/controller/helper/developeruser_test.go
+++ b/pkg/controller/helper/developeruser_test.go
@@ -1,0 +1,71 @@
+package helper
+
+import (
+	"testing"
+
+	capabilitiesv1beta1 "github.com/3scale/3scale-operator/apis/capabilities/v1beta1"
+
+	logrtesting "github.com/go-logr/logr/testing"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+func TestFindDeveloperUserList(t *testing.T) {
+	ns := "some_namespace"
+	adminRole := "admin"
+	memberRole := "member"
+	providerAccountURLStr := "https://example.com"
+	providerAccountToken := "12345"
+	anotherProviderSecretName := "anotherSecretName"
+
+	data := map[string]string{
+		providerAccountSecretURLFieldName:   providerAccountURLStr,
+		providerAccountSecretTokenFieldName: providerAccountToken,
+	}
+	providerSecret := GetTestSecret(ns, providerAccountDefaultSecretName, data)
+
+	data = map[string]string{
+		providerAccountSecretURLFieldName:   "https://other.example.com",
+		providerAccountSecretTokenFieldName: providerAccountToken,
+	}
+	anotherProviderSecret := GetTestSecret(ns, anotherProviderSecretName, data)
+
+	devUser1 := &capabilitiesv1beta1.DeveloperUser{
+		ObjectMeta: metav1.ObjectMeta{Name: "devUser1", Namespace: ns},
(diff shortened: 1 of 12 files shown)


eguzki:
progress: **35.6%**

pkg/controller/helper/url_test.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+func TestURLFromDomain(t *testing.T) {
+	cases := []struct {
+		testName    string
+		domain      string
+		expectedErr bool
+		expectedURL url.URL
+	}{
+		{"Invalid domain", ":foo", true, url.URL{}},
```
miguelsorianod, line 15: It would be nice to have a case for a valid domain too

pkg/controller/helper/helper_test.go
```diff
@@ -0,0 +1,181 @@
...
+		},
+		StringData: data,
+		Type:       v1.SecretTypeOpaque,
+	}
+	secret.Data = helper.GetSecretDataFromStringData(secret.StringData)
+	return secret
+}
+
+// RoundTripFunc .
+type RoundTripFunc func(req *http.Request) *http.Response
```
miguelsorianod, line 71: Go includes a RountTripper interface which I think is intended to do the same: https://golang.org/pkg/net/http/#RoundTripper

Would it be a better idea to reuse that?

eguzki: That interface is what `RoundTripFunc` type implements. It needs to be like that to be assigned in the `Transport` field of the http client. 

This `RoundTripFunc` makes the interface to the test users a little bit easier to use. Instead of
```
RoundTrip(*Request) (*Response, error)
```
the test user must implement
```
func(req *http.Request) *http.Response
```
and if there is an error, the test should raise the error. For instance:
```
httpClient := NewTestClient(func(req *http.Request) *http.Response {              
    respObject := struct {                                                        
        Errors map[string][]string `json:"errors"`                                
    }{                                                                            
        map[string][]string{"some_attr": []string{"not valid"}},                  
    }                                                                             
                                                                                  
    responseBodyBytes, err := json.Marshal(respObject)                            
    ok(t, err)             <- raise test failed if there is error                                                        
                                                                                  
    return &http.Response{                                                        
        StatusCode: http.StatusUnprocessableEntity,                               
        Body:       ioutil.NopCloser(bytes.NewBuffer(responseBodyBytes)),         
        Header:     make(http.Header),                                            
    }                                                                             
})                                                                                
```
What do you have in mind?

miguelsorianod: I see.

Thanks for the explanation. This approach looks good to me. It allows the user not having to create its own type and provide a method instead.

pkg/controller/helper/developeruser_test.go
```diff
@@ -0,0 +1,71 @@
...
+		ObjectMeta: metav1.ObjectMeta{Name: "devUser3", Namespace: ns},
+		Spec: capabilitiesv1beta1.DeveloperUserSpec{
+			Username: "devUser3", Email: "devUser3@example.com", Role: &adminRole,
+			ProviderAccountRef: &corev1.LocalObjectReference{Name: anotherProviderSecretName},
+		},
+	}
+
+	cl := fake.NewFakeClient(anotherProviderSecret, providerSecret, devUser1, devUser2, devUser3)
+
+	// Find only not admin users from the same provider account
```
miguelsorianod, line 58: Is this comment right? it seems the test looks for admin users and not for "not admin users"

PR #593, opened 2021-03-11
go.mod,sum: fix CVE-2020-9283

miguelsorianod:
This PR addresses CVE-2020-9283.
The first `golang.org/x/crypto` version that addresses the CVE is `v0.0.0-20200220183623-bac4c82f6975` but the version has been replaced to the most updated one that it's been seen in `go.sum` file before this change, which is: `v0.0.0-20200414173820-0848c9571904`


unleashed (approved):
LGTM

PR #596, opened 2021-03-26
Update backend-redis internal database version from 3.2 to 5.x

miguelsorianod:
This PR implements changing the used backend-redis internal database Redis version from 3.2 to 5.x
This has been done for both templates and operator.
In case of the productized version the image URLs will be different than the upstream ones that are the ones that can be seen in this PR. The productized URLs will be updated/set on the version-specific branch.

In the operator case it also includes upgrade automation to migrate from one version to another. The upgrade migration includes:
* Container image change
* Needed command changes
* PodTemplate label changes

In the templates case upgrade documentation will be needed.

### DIFF
--- a/pkg/3scale/amp/auto-generated-templates/amp/amp-eval-s3.yml
+++ b/pkg/3scale/amp/auto-generated-templates/amp/amp-eval-s3.yml
@@ -238,7 +238,7 @@ objects:
           - --daemonize
           - "no"
           command:
-          - /opt/rh/rh-redis32/root/usr/bin/redis-server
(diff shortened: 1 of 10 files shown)


miguelsorianod:
You will see that there's an additional commit that in case of the templates it hardcodes system-redis database to redis 3.2 instead of taking the "REDIS_URL" OCP template parameter. The reason for that is that that variable was shared for both redis DBs in templates case and this PR is intended to only change backend-redis.

In the future that commit should be reverted and both should point to REDIS_URL again.
Another alternative is updating system-redis to 5.x too but for that we would need to wait until that has been tested and implemented.

davidor (approved):

PR #598, opened 2021-04-19
Prepare 2.10 stable

miguelsorianod:
Apply 3scale 2.10-specific upstream changes for the 3scale-2.10-stable (upstream) branch


miguelsorianod:
Another PR after this one is merged will be opened with the cherry-picks that were performed between ER1 and GA on the 3scale-2.10-stable-prod branch

pkg/3scale/amp/auto-generated-templates/amp/amp.yml
```diff
@@ -4187,16 +4187,16 @@ parameters:
   value: "null"
 - name: AMP_BACKEND_IMAGE
   required: true
-  value: quay.io/3scale/apisonator:nightly
+  value: quay.io/3scale/3scale210:apisonator-3scale-2.10.0-GA
```
miguelsorianod, line 4190: We will need to build 2.10 upstream GA images at some point

eguzki:
```
CIRCLE_CI_API_TOKEN=XXX IMAGES_GIT_REF=3scale-2.10.0-GA IMAGES_REMOTE_TAG=3scale-2.10.0-GA make build-3scale-images

miguelsorianod: :+1:

miguelsorianod: Unfortunately that mechanism was still not added in 2.10.

I tried to run it from master branch modifying the OPERATOR_BRANCH to point to 3scale-2.10-stable, to get the code from there but it fails because the CircleCI config was still not there.
Leaving the default OPERATOR_BRANCH value, master in this case, is not an option because it would take source code from the master branch which is not what we want for 2.10

eguzki (approved):

pkg/3scale/amp/auto-generated-templates/amp/amp.yml
miguelsorianod: Image creation will be addressed separately due to identified issues with one of the component's image

PR #602, opened 2021-04-28
clarify wildcard domain

akostadinov:
I'm not sure how to clearly express the requirement for the wildcard domain. You can see my attempt here. I will welcome any other suggestions though.


eguzki (approved):

eguzki:
Thank you for the contribution!

PR #612, opened 2021-06-01
local installation of development tools

eguzki:
Install development tools at `${PROJECT_BASE_PATH}/bin`. Avoid conflict with the same development tools for different releases required in other working projects. 

* Operator-sdk
* kustomize
* controller-gen
* go-bindata

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,6 +102,13 @@ $(OPERATOR_SDK):
 .PHONY: operator-sdk
 operator-sdk: $(OPERATOR_SDK)
 
+GO_BINDATA=$(PROJECT_PATH)/bin/go-bindata
+$(GO_BINDATA):
+	$(call go-get-tool,$(GO_BINDATA),github.com/go-bindata/go-bindata/v3/...@v3.1.3)
+
+.PHONY: go-bindata
+go-bindata: $(GO_BINDATA)
+
 # Install CRDs into a cluster
 install: manifests $(KUSTOMIZE)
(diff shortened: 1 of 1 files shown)


miguelsorianod:
> Note: go-bindata could not be installed locally as go generate command runs command from annotation and the binary must be found in the $PATH. Usually the project local folders are not included in the system $PATH.

To solve this issue we can update the PATH in the Makefile to add the new local bin path with preference over the other directories and export the variable

Makefile
```diff
@@ -79,21 +78,45 @@ run: export THREESCALE_DEBUG=1
...
+KUSTOMIZE=$(PROJECT_PATH)/bin/kustomize
+$(KUSTOMIZE):
+	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v3@v3.5.4)
+
+.PHONY: kustomize
+kustomize: $(KUSTOMIZE)
+
+OPERATOR_SDK = $(PROJECT_PATH)/bin/operator-sdk
+$(OPERATOR_SDK):
+	curl -sSL https://github.com/operator-framework/operator-sdk/releases/download/v1.2.0/operator-sdk-v1.2.0-x86_64-linux-gnu -o $(OPERATOR_SDK)
```
miguelsorianod, line 99: A note here: This will be architecture dependent. If other people are trying to use it from other architectures it will fail.

eguzki: Updated. 

It would be good if @krishvoor could test on `ppc64le` arch

miguelsorianod (approved):

PR #613, opened 2021-06-01
Enable builds/ testing on ppc64le architecture

krishvoor:
This PR will allow to build container image
and allow e2e testing on ppc64le architecture

Signed-off-by: Krishna Harsha Voora <krishvoor@in.ibm.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,8 +18,7 @@ COPY pkg/ pkg/
 COPY version/ version/
 
 # Build
-RUN export GOARCH=`go env GOARCH` \
-   && CGO_ENABLED=0 GOOS=linux GOARCH=${GOARCH} GO111MODULE=on go build -a -o manager main.go
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=`go env GOARCH` GO111MODULE=on go build -a -o manager main.go
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,8 @@ COPY pkg/ pkg/
 COPY version/ version/
 
 # Build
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -18,7 +18,8 @@ COPY pkg/ pkg/
 COPY version/ version/
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
```
eguzki, line 21: Why not

```
RUN CGO_ENABLED=0 GOOS=linux GOARCH=`go env GOARCH` GO111MODULE=on go build -a -o manager main.go
```

Makefile
```diff
@@ -66,7 +66,7 @@ TEST_E2E_PKGS = $(shell $(GO) list ./... | grep 'github.com/3scale/3scale-operat
 ENVTEST_ASSETS_DIR=$(PROJECT_PATH)/testbin
 test-e2e: generate fmt vet manifests
 	mkdir -p ${ENVTEST_ASSETS_DIR}
-	test -f $(ENVTEST_ASSETS_DIR)/setup-envtest.sh || curl -sSLo $(ENVTEST_ASSETS_DIR)/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/v0.6.3/hack/setup-envtest.sh
```
eguzki, line 69: `controller-runtime v0.6.3` is being used when compiling the operator. 

Why should we change the version?

eguzki (changes requested):

Dockerfile
krishvoor: Makes sense, will update.

Makefile
krishvoor: `controller-runtime v0.6.3` does not contain ppc64le compatible `kubebuilder-tools`.
Bumping up the `controller-runtime v0.8.0` will allow `test-e2e` to be executed on ppc64le as well.

eguzki (approved):

PR #618, opened 2021-06-09
Add System's USER_SESSION_TTL configurability

miguelsorianod:
This PR exposes the `USER_SESSION_TTL` System's setting.
Related issue: 
It exposes this setting as an attribute in the `system-app` secret as usual.

This is the behavior:

Secret side:
* If the attribute is not set with secret pre-creation then it will automatically be set to the `""` string

System pod side
* The value is referenced as an environment variable in the system-app pod containers, as seret entry reference, which should exist.
* If the referenced value is the empty string `""` then USER_SESSION_TTL on system side will be the default at system level

An upgrade procedure has been implemented too to make sure the secret entry is created and that the system-app pod containers reference this secret entry.

The change has also been applied to templates. This change will require an upgrade procedure for templates too.

Pending:
[ ] Functionality verification in a real deployment, in both a clean environment and an upgrade


miguelsorianod:
Pending to do some verifications in a real deployment but let's review it first and if it looks ok I'll proceed to do the verifications

akostadinov:
Wow, a lot of changes just for one new option.

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -638,11 +688,11 @@ func (u *UpgradeApiManager) upgradeSidekiqBackendRouteEnv(desired *appsv1.Deploy
 	return reconcile.Result{Requeue: update}, nil
 }
 
-func ensureBackendRouteEnvVar(desired v1.EnvVar, existingEnvVars *[]v1.EnvVar) bool {
+func ensureEnvVar(desired v1.EnvVar, existingEnvVars *[]v1.EnvVar) bool {
```
eguzki, line 691: This method belongs to `pkg/helper/envvarutils.go`

eguzki (changes requested):
Looking good overall. 

I wonder if the upgrade procedure is really needed. Maybe delegating to the regular reconciliation loop would be enough. 

Few comments left

miguelsorianod:
> I wonder if the upgrade procedure is really needed. Maybe delegating to the regular reconciliation loop would be enough.

I thought about that too but the upgrade is needed because the environment variable needs to be added to the deployment, during regular reconciliation that part is not reconciled, and the value has to be added to the secret because the environment variable references it.

pkg/3scale/amp/operator/upgrade.go
miguelsorianod: :+1: I'll add it there

miguelsorianod: Added

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -542,6 +547,23 @@ func (u *UpgradeApiManager) findDeploymentTriggerOnImageChange(triggerPolicies [
 	return result, nil
 }
 
+func (u *UpgradeApiManager) upgradeSystemAppUserSessionTTL() (reconcile.Result, error) {
+	system, err := System(u.apiManager, u.Client())
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	baseAPIManagerLogicReconciler := NewBaseAPIManagerLogicReconciler(u.BaseReconciler, u.apiManager)
```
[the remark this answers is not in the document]
miguelsorianod: :+1: I'll take a look at it and open a new PR in that case

eguzki (approved):

PR #620, opened 2021-06-10
3scale monitoring deploys prometheus rules as opt-in feature

eguzki:
https://issues.redhat.com/browse/THREESCALE-7137

* Updated doc with new APImanager CRD info and feature explanation
* Upgrade procedure not implemented as not needed. Just one new optional parameter added. 
* PrometheusRule serialization CLI command implemented to auto-generate published prometheus rules 
* CI testing to validate autogenerated content is not updated manually


miguelsorianod (commented):
Looks good to me. Can we have some doc about it?
In that doc we should mention that users should locally modify all prometheus rules yaml to replace the `__NAMESPACE__` placeholder everywhere for the namespace where they want to deploy them.

An alternative instead of committing the rules would be being able to provide the namespace in the CLI command. That means they would need to run that command to generate them

eguzki:
> An alternative instead of committing the rules would be being able to provide the namespace in the CLI command. That means they would need to run that command to generate them

If the customer wants *custom* prometheus rules (because the expressions, duration or severity does not fit for their infra), it does not make sense to generate them using CLI with namespace parameter. The CLI command would have to provide too many flags.

miguelsorianod:
> > An alternative instead of committing the rules would be being able to provide the namespace in the CLI command. That means they would need to run that command to generate them
> 
> If the customer wants _custom_ prometheus rules (because the expressions, duration or severity does not fit for their infra), it does not make sense to generate them using CLI with namespace parameter. The CLI command would have to provide too many flags.

I was not proposing  custom prometheus rules. I was proposing adding a way to generate them with the namespace they want. In the commited rules I see multiple references to `__NAMESPACE__` which I understand is a placeholder and the user will have to do a find and replace of all the occurrences to their desired namespace.

miguelsorianod:
The doc looks good to me :+1:

eguzki:
added `--namespace` flag

miguelsorianod (approved):
:+1:  great work

PR #625, opened 2021-06-25
apicast custom policies

eguzki:
https://issues.redhat.com/browse/THREESCALE-7032

* Reconciliation of apicast production and staging volumes and volumeMounts introduced. Only for volumes resulting from custom policies configuration in the APIManager CR. Manually mounted volumes are not reconciled. 

Tracking of volumes (and volumemounts) resulting from the custom policies configuration is done using DC annotations
The following APIManager CR:
```
spec:
  apicast:
    productionSpec:
      customPolicies:
      - name: Example
        secretRef:
          name: cp-1
        version: "0.1"
      replicas: 1
    stagingSpec:
      customPolicies:
      - name: Exampleother
        secretRef:
          name: cp-3
        version: "0.1"
      - name: Exampleother2
        secretRef:
          name: cp-1
        version: "0.1"
```

Will result in the following custom policies annotations in the apicast-staging DC:

```
apiVersion: apps.openshift.io/v1
kind: DeploymentConfig
metadata:
  annotations:
    apps.3scale.net/apicast-policy-volume-550ce71bf15f7ab80906b5e066d2fc30: policy-01-exampleother
    apps.3scale.net/apicast-policy-volume-b2629bce1bd345c97d4037889ec37939: policy-01-exampleother2
```

in apicast production DC:
```
apiVersion: apps.openshift.io/v1
kind: DeploymentConfig
metadata:
  annotations:
    apps.3scale.net/apicast-policy-volume-550ce71bf15f7ab80906b5e066d2fc30: policy-01-example
```


eguzki:
@miguelsorianod feedback request before going forward

pkg/3scale/amp/operator/apicast_options_provider.go
```diff
@@ -4,23 +4,31 @@ import (
...
 type ApicastOptionsProvider struct {
 	apimanager     *appsv1alpha1.APIManager
+	namespace      string
 	apicastOptions *component.ApicastOptions
+	client         client.Client
+	secretSource   *helper.SecretSource
 }
 
-func NewApicastOptionsProvider(apimanager *appsv1alpha1.APIManager) *ApicastOptionsProvider {
+func NewApicastOptionsProvider(apimanager *appsv1alpha1.APIManager, namespace string, client client.Client) *ApicastOptionsProvider {
```
miguelsorianod, line 25: I think `namespace`  namespace might not be needed. We can use the `APIManager` namespace as I understand they will always be in the same namespace

pkg/3scale/amp/operator/apicast_reconciler.go
```diff
@@ -151,8 +157,30 @@ func (r *ApicastReconciler) apicastLogLevelEnvVarMutator(desired, existing *apps
 	return reconcilers.DeploymentConfigEnvVarReconciler(desired, existing, "APICAST_LOG_LEVEL")
 }
 
-func Apicast(apimanager *appsv1alpha1.APIManager) (*component.Apicast, error) {
-	optsProvider := NewApicastOptionsProvider(apimanager)
+func (r *ApicastReconciler) volumeMountsMutator(desired, existing *appsv1.DeploymentConfig) bool {
+	changed := false
+
+	if !reflect.DeepEqual(existing.Spec.Template.Spec.Containers[0].VolumeMounts, desired.Spec.Template.Spec.Containers[0].VolumeMounts) {
```
miguelsorianod, line 163: Have you checked if K8s initializes some uninitialized field that could led to some issue with the comparison? Like the case with the K8s `Service`'s `ClusterIP` attribute

pkg/3scale/amp/operator/apicast_reconciler.go
```diff
@@ -151,8 +157,30 @@ func (r *ApicastReconciler) apicastLogLevelEnvVarMutator(desired, existing *apps
...
+
+	return changed
+}
+
+func (r *ApicastReconciler) volumesMutator(desired, existing *appsv1.DeploymentConfig) bool {
+	changed := false
+
+	if !reflect.DeepEqual(existing.Spec.Template.Spec.Volumes, desired.Spec.Template.Spec.Volumes) {
+		changed = true
+		existing.Spec.Template.Spec.Volumes = desired.Spec.Template.Spec.Volumes
```
miguelsorianod, line 176: Adding a note here: we need to be aware that this might have an impact on existing users installations. I have the feeling some users might be manually creating volumes for it, and adding this will revert those changes.
I am aware we've never told it's supported but still.

miguelsorianod:
The approach looks good to me

pkg/3scale/amp/operator/apicast_reconciler.go
eguzki: The apicast operator is reconciling the volumes and volumemounts in the same way: https://github.com/3scale/apicast-operator/blob/a787307dabeef1f9b90100e55068cb25132feb9e/controllers/apps/apicast_logic_reconciler.go#L238

There is a difference, though, in the apicast operator the resource is Deployment and the 3scale operator deploys DeploymentConfigs. I have deployed and it does not seem to be anything that can ruin DeepEqual.
```
volumeMounts:
        - mountPath: /opt/app-root/src/policies/Example/0.1
          name: policy-01-example
          readOnly: true
```
```
volumes:
      - name: policy-01-example
        secret:
          defaultMode: 420
          secretName: cp-1
```

pkg/3scale/amp/operator/apicast_reconciler.go
eguzki: Yes, that is why I underlined this very clearly in the description. IMO, this is something the operator can do and should do. The CRD should be the public interface to add volumes, like implemented in this PR for custom policies. 

Anyway, I do not have a strong position on this. If you think this can be a source of issues, I can remove the volume generic reconciliation. I would have to add specific reconciliation for the expected volumes in order to support adding and removing custom policies in the CR.

pkg/3scale/amp/operator/apicast_options_provider.go
eguzki: updated

eguzki:

pkg/3scale/amp/operator/apicast_reconciler.go
miguelsorianod: :+1:  thanks for looking at it

pkg/3scale/amp/operator/apicast_reconciler.go
miguelsorianod: I think this might led to issues. Although I agree that is something the operator can and should do. I don't have a strong position to this either. I was just adding a note so we are aware that if we go ahead with this we might receive some feedback about it

pkg/3scale/amp/operator/apicast_reconciler.go
eguzki: Ok, this is not entirely true. It is true for the volumemounts, however, for the volumes, the `defaultMode` is not set in the desired, hence, DeepEqual returns false:
```
 	Name: "policy-01-exampleother",
  	VolumeSource: v1.VolumeSource{
  		... // 3 identical fields
  		AWSElasticBlockStore: nil,
  		GitRepo:              nil,
  		Secret: &v1.SecretVolumeSource{
  			SecretName:  "cp-3",
  			Items:       nil,
- 			DefaultMode: nil,
+ 			DefaultMode: &420,
  			Optional:    nil,
  		},
  		NFS:   nil,
  		ISCSI: nil,
  		... // 20 identical fields
  	},
  }
```

I will write custom comparator to make it work regardless DefaultMode

pkg/3scale/amp/operator/apicast_reconciler.go
eguzki: Implemented specific reconciliation only for volumes (and volumemounts) derived from the custom policy conf. It allows manually added volumes and will not remove them in the reconcilliation loop.

eguzki:

pkg/3scale/amp/component/apicast.go
```diff
@@ -16,6 +18,9 @@ import (
 const (
 	ApicastStagingName    = "apicast-staging"
 	ApicastProductionName = "apicast-production"
+
+	CustomPoliciesMountBasePath    = "/opt/app-root/src/policies"
+	CustomPoliciesAnnotationPrefix = "policy.apicast.3scale.redhat.com/"
```
miguelsorianod, line 23: Maybe we should use `apps.3scale.net` as the prefix as in other annotations already set for this operator?
https://github.com/3scale/3scale-operator/blob/master/apis/apps/v1alpha1/apimanager_types.go#L39

pkg/3scale/amp/component/apicast.go
```diff
@@ -90,8 +95,9 @@ func (apicast *Apicast) StagingDeploymentConfig() *appsv1.DeploymentConfig {
 	return &appsv1.DeploymentConfig{
 		TypeMeta: metav1.TypeMeta{APIVersion: "apps.openshift.io/v1", Kind: "DeploymentConfig"},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:   ApicastStagingName,
-			Labels: apicast.Options.CommonStagingLabels,
+			Name:        ApicastStagingName,
```
miguelsorianod, line 98: I see the annotations have been set in the `DeploymentConfig` annotations, and not in the PodTemplateSpec annotations of the DeploymentConfig. Is this intentional? The difference is that if the annotation is changed here there will be no redeployment of the pods of the DeploymentConfig, whereas if we put them in the PodTemplateSpec the pods would be redeployed

eguzki: Yes it is intentional. Those annotations are for operator tracking purposes. Adding/updating/removing those annotations should not be the only reason to respin a deployment. An also, there is not a good reason to have them in the pod. IMO, this is data to operate at DC level.

pkg/3scale/amp/operator/apicast_reconciler.go
```diff
@@ -151,8 +161,112 @@ func (r *ApicastReconciler) apicastLogLevelEnvVarMutator(desired, existing *apps
...
+	return changed
+}
+
+func apicastCustomPolicyAnnotationsMutator(desired, existing *appsv1.DeploymentConfig) bool {
+	// It is expected that APIManagerMutator has already added desired annotations to the existing annotations
+	// find existing custom policy annotations not in desired and delete them
+	updated := false
+	existingCustomPolicyVolumeNames := component.ApicastVolumeNamesFromAnnotations(existing.Annotations)
+	desiredCustomPolicyVolumeNames := component.ApicastVolumeNamesFromAnnotations(desired.Annotations)
+	if !helper.StringSliceEqualWithoutOrder(existingCustomPolicyVolumeNames, desiredCustomPolicyVolumeNames) {
```
miguelsorianod, line 250: Instead of doing the "string slice equal without order" wouldn't sorting the strings work too?

eguzki: It will also work, sort and then reflect.DeepEqual. That is another implementation to compare slices regardless order.

pkg/3scale/amp/component/apicast.go
eguzki: Updated implementation. Now it is 

PREFIX-HASH: VOLUME_NAME

```
apps.3scale.net/apicast-policy-volume-550ce71bf15f7ab80906b5e066d2fc30: policy-01-exampleother
```

pkg/3scale/amp/component/apicast.go
```diff
@@ -16,6 +18,9 @@ import (
 const (
 	ApicastStagingName    = "apicast-staging"
 	ApicastProductionName = "apicast-production"
+
+	CustomPoliciesMountBasePath    = "/opt/app-root/src/policies"
+	CustomPoliciesAnnotationPrefix = "apps.3scale.net/apicast-policy-volume"
```
miguelsorianod, line 23: The name of the constant is kind of confusing. I understand why it's been named with "prefix", but prefix is the part before the `/` https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/.
So this is really a partial annotation key and not a prefix

pkg/3scale/amp/component/apicast.go
miguelsorianod: :+1:

pkg/3scale/amp/component/apicast.go
eguzki: Fair enough. I have updated. It is the prefix of the key of the annotation, hence: `CustomPoliciesAnnotationKeyPrefix`

eguzki:

miguelsorianod (changes requested):
I think this https://github.com/3scale/3scale-operator/pull/625#discussion_r674912391 has still to be addressed right?

eguzki:
> I think this [#625 (comment)](https://github.com/3scale/3scale-operator/pull/625#discussion_r674912391) has still to be addressed right?

That comment does not make sense anymore. That had a meaning for when *hard* volume reconciliation was done. The current implementation, more flexible, allows manually mounted volumes and only reconciles volume mounts for apicast custom policies.

pkg/3scale/amp/component/apicast.go
miguelsorianod: :+1:

miguelsorianod (approved):
looks good to me. great work :+1:

PR #626, opened 2021-06-28
[WIP] Update zync ruby base image to ruby-27 ubi8

miguelsorianod:
Zync component has recently updated its base image to ruby-27 ubi8. This fixes the zync nightly builds lately failing

### DIFF
--- a/pkg/3scale/amp/manual-templates/amp/build.yml
+++ b/pkg/3scale/amp/manual-templates/amp/build.yml
@@ -11,12 +11,12 @@ objects:
     annotations:
     labels:
       app: zync
-    name: ruby-25-ubi7
+    name: ruby-27-ubi8
   spec:
     tags:
     - from:
         kind: DockerImage
-        name: registry.redhat.io/ubi7/ruby-25
+        name: registry.access.redhat.com/ubi8/ruby-27
       name: latest
 
 - apiVersion: v1
@@ -42,7 +42,7 @@ objects:
           name: "${PULLSECRET_NAME}"
         from:
           kind: ImageStreamTag
-          name: ruby-25-ubi7:latest
+          name: ruby-27-ubi8:latest
       type: Docker
 
 - apiVersion: v1


pkg/3scale/amp/manual-templates/amp/build.yml
```diff
@@ -11,12 +11,12 @@ objects:
...
     labels:
       app: zync
-    name: ruby-25-ubi7
+    name: ruby-27-ubi8
   spec:
     tags:
     - from:
         kind: DockerImage
-        name: registry.redhat.io/ubi7/ruby-25
+        name: registry.access.redhat.com/ubi8/ruby-27
```
miguelsorianod, line 19: An interesting note about this PR is that it seems that the URL for it is now in `registry.access.redhat.com` domain instead of `registry.redhat.io`. It seems that depending on the image they've changed it to `registry.access.redhat.com`

eguzki:
I would not merge this PR. 

There is a pending PR to stop building 3scale components nightly images. Instead, each component team is responsible for the create `latest` image and so far it is being done. 

The only question is what do we do with stable releases.. what images do we use for testing? We ask them to also create latest stable images for 2.11, 2.12...??

miguelsorianod:
> There is a pending PR to stop building 3scale components nightly images. Instead, each component team is responsible for the creation of the latest image and so far it is being done.

Do you know if zync is already doing that? if we don't merge this our nightly CircleCI tests will be failing and no nightly images will be built at all until they do it.

Regarding stable images, yes. They should build stable candidate images and stable images.

miguelsorianod:
Additionally, in any case we should update `build.yml` to be able to be successfully executed even if we don't do nightly builds anymore. Otherwise we should then open a PR removing `build.yml` but I'm not sure if we can do that without warning as I seem to remember other people might be relying on it

eguzki:
There is this issue https://issues.redhat.com/browse/THREESCALE-7142 which seems is still under development. 

When done, we should open PR removing the build.yaml and circleci nightl build jobs. 

Then think about what we do with stable releases in upstream. Maybe make them generate latest images from release branches as well.

miguelsorianod:
:+1:  then until that is done we should fix this to be able to have nightly builds, unless we want to live without nightly images until that task is done.
Regarding removing `build.yaml` file, I think we should at least comment this to @3scale/qe team because they might be relying on it and the removal might impact them.
Regarding stable releases, see my previous comment. I think all components should generate candidate images and stable images on the community side. Otherwise we would be testing nightly component images (it's what we are doing now) on candidate / stable branches which does not make sense.

eguzki:
Ok, let's merge this.

eguzki (approved):

miguelsorianod:
Some tests are failing due to routes not being created. It seems the issue might be some recent change in zync component. I've observed the following logs in the `zync-que` pod:
```
E, [2021-06-28T12:23:11.365748 #1] ERROR -- : {"lib":"que","hostname":"zync-que-1-bxlpw","pid":1,"thread":13980,"event":"job_errored","job_id":23,"elapsed":0.084466909,"error":"NoMethodError: undefined method `[]=' for nil:NilClass"}
E, [2021-06-28T12:23:11.382571 #1] ERROR -- : {"lib":"que","hostname":"zync-que-1-bxlpw","pid":1,"thread":13700,"event":"job_errored","job_id":32,"elapsed":0.082535315,"error":"NoMethodError: undefined method `[]=' for nil:NilClass"}
E, [2021-06-28T12:23:11.383638 #1] ERROR -- : {"lib":"que","hostname":"zync-que-1-bxlpw","pid":1,"thread":13560,"event":"job_errored","job_id":49,"elapsed":0.082126243,"error":"NoMethodError: undefined method `[]=' for nil:NilClass"}
```

I've pinged System team to take a look at it.
Let's wait for them to try to get the tests working again.

pkg/3scale/amp/manual-templates/amp/build.yml
guicassolato: `registry.redhat.io/ubi8/ruby-27` also exists.

guicassolato (approved):

miguelsorianod:
This issue does not apply anymore after the changes made in #636. Closing it

PR #627, opened 2021-06-30
apimanager: update status regardless spec error

eguzki:
Status was not updated when spec reconciliation returned error. The only way to debug errors was looking at the logs of the operator. 

With this change, on spec reconciliation error, the status is being updated with `Available` condition set to `false`. 

This reminds me that we still have a TODO which is try to set the error root cause in the status section

### DIFF
--- a/controllers/apps/apimanager_controller.go
+++ b/controllers/apps/apimanager_controller.go
@@ -112,21 +112,22 @@ func (r *APIManagerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) 
 		return ctrl.Result{Requeue: true}, nil
 	}
 
-	result, err := r.reconcileAPIManagerLogic(instance)
-	if err != nil {
-		logger.Error(err, "Error during reconciliation")
-		return result, err
-	}
-	if result.Requeue {
+	specResult, specErr := r.reconcileAPIManagerLogic(instance)
+	if specErr != nil && specResult.Requeue {
 		logger.Info("Reconciling not finished. Requeueing.")
-		return result, nil
+		return specResult, nil
 	}
 
+	// reconcile status regardless specErr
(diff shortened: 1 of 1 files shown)


miguelsorianod:
> With this change, on spec reconciliation error, the status is being updated with Available condition set to false.

How is that being done? specError is not being used when reconciling status

Have you ckecked if there is some risk that if spec reconcile fails, statusReconcile could crash because of some assumption?

eguzki:
> if there is some risk that if spec reconcile fails

The status reconciliation does not depend on the spec reconciliation (as we do in other controllers like backend controller). I think it is ok and less coupled. The status reconciliation checks current existing state (specifically the deployments). If the deployments were correct, the status `Available` would be true. So there is a potential use case where the status is Available but still the spec reconciliation fails. In that case, the operator would still re-queue and try to reconcile again the spec.

I do not see any issue.

miguelsorianod:
:+1: about the previous comment. I still have this question from my previous comment. Could you tell me about it?

> With this change, on spec reconciliation error, the status is being updated with Available condition set to false.

How is that being done? specError is not being used when reconciling status

eguzki:
> How is that being done? specError is not being used when reconciling status

The statement `on spec reconciliation error, the status is being updated with Available condition set to false.` is not accurate enough. I meant that before this change, on spec reconciliation error, status was not updated. With this change, the status is being updated. The status update outcome is based on status reconciliation. This status reconciliation logic does not take into account spec error. Just reads existing state and updates status according to the checks implemented. Thus, it could be that:
* on spec error -> status available condition set to false (very likely
* on spec error -> status available condition set to true (less likely, but still possible)

miguelsorianod (approved):
:+1:  thanks. Looks good to me

PR #630, opened 2021-07-09
Fix link and typos

mledoze:


eguzki (approved):

eguzki:
thanks @mledoze

eguzki:
Templates check is expected to fail. Nothing to do with the 3scale operator. It is one of the components that is not working as expected. We are already working on that.

PR #632, opened 2021-07-12
apicast enable tls at pod level

eguzki:
3scale managed APIcast pods can be deployed to serve HTTP endpoints over TLS. Each APIcast instance, i.e. `production` and `staging`, can be individually configured to serve, or not, HTTP over TLS.

```
kind: APIManager
metadata:
  name: apimanager-apicast-custom-environment
spec:
  wildcardDomain: <desired-domain>
  apicast:
    productionSpec:
      httpsPort: 8443
      httpsCertificateSecretRef:
        name: mycertsecret
    stagtingSpec:
      httpsPort: 8443
      httpsCertificateSecretRef:
        name: mycertsecret
```

https://issues.redhat.com/browse/THREESCALE-7297

Depends on #625. Should not be merged before that.


apis/apps/v1alpha1/apimanager_types.go
```diff
@@ -184,6 +205,19 @@ type ApicastProductionSpec struct {
 	// +optional
 	// +kubebuilder:validation:Enum=debug;info;notice;warn;error;crit;alert;emerg
 	LogLevel *string `json:"logLevel,omitempty"` // APICAST_LOG_LEVEL
+	// CustomPolicies specifies an array of defined custome policies to be loaded
+	// +optional
+	CustomPolicies []CustomPolicySpec `json:"customPolicies,omitempty"`
+	// HttpsPort controls on which port APIcast should start listening for HTTPS connections. If this clashes with HTTP port it will be used only for HTTPS.
+	// +optional
+	HTTPSPort *int32 `json:"httpsPort,omitempty"` // APICAST_HTTPS_PORT
```
miguelsorianod, line 213: Which one of this parameters is the one that controls whether TLS is enabled? `httpsPort`? It would be good to explain that behavior documented somewhere, or reorganize a little bit the structure in the CR to make it clear that "you are enabling tls"

doc/apimanager-reference.md
```diff
@@ -106,6 +107,10 @@ Generated using [github-markdown-toc](https://github.com/ekalinin/github-markdow
 | Resources | `resources` | [v1.ResourceRequirements](https://v1-17.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#resourcerequirements-v1-core) | No | `nil` | Resources describes the compute resource requirements. Takes precedence over `spec.resourceRequirementsEnabled` with replace behavior |
 | Workers | `workers` | integer | No | Automatically computed. Check [apicast doc](https://github.com/3scale/APIcast/blob/master/doc/parameters.md#apicast_workers) for further info. | Defines the number of worker processes |
 | LogLevel | `logLevel` | string | No | N/A | Log level for the OpenResty logs  (see [docs](https://github.com/3scale/APIcast/blob/master/doc/parameters.md#apicast_log_level)) |
+| CustomPolicies | `customPolicies` | [][CustomPolicySpec](#CustomPolicySpec) | No | N/A | List of custom policies |
+| HTTPSPort | `httpsPort` | int | No | **8443** only when `httpsCertificateSecretRef` is provided | Controls on which port APIcast should start listening for HTTPS connections. Do not use `8080` as HTTPS port (see [docs](https://github.com/3scale/APIcast/blob/master/doc/parameters.md#apicast_https_port)) |
```
miguelsorianod, line 111: Have we checked if there is some other component that is referencing the APIcast port or doing some assumption about its value?

eguzki: Apicast is not, as we are using the same in the apicast operator and it is working.

It does not have any sidecar container in the pod, so far. Thus it should be available.

apis/apps/v1alpha1/apimanager_types.go
eguzki: It is in the doc: https://github.com/3scale/3scale-operator/pull/632/files#diff-72dab7f3126dc7dbd85ab6398f36d48315960176cddc220c7fd9bdb36da97fb6R5

`Enable TLS at APIcast pod level setting either `httpsPort` or `httpsCertificateSecretRef` fields or both.`

doc/apimanager-reference.md
miguelsorianod: Yes, I understand that.
My question is if some other component relies on the communication with apicast through http and we would need to change that too. For example, system to apicast communication

apis/apps/v1alpha1/apimanager_types.go
miguelsorianod: Maybe add it in the apimanager_type too then? :thinking:

eguzki: I will add it to the apimanager_type

doc/apimanager-reference.md
eguzki: the HTTP port (called `gateway` in the service) is still enabled. So other components should not be affected. Unless there is a conflict, in which case only TLS will be served [link](https://github.com/3scale/APIcast/blob/master/doc/parameters.md#apicast_https_port).

The doc explicitly states `Do not use 8080 as HTTPS port` to avoid conflicts.

apis/apps/v1alpha1/apimanager_types.go
eguzki: added

eguzki:

pkg/3scale/amp/operator/apicast_reconciler.go
```diff
@@ -86,29 +88,31 @@ func (r *ApicastReconciler) Reconcile() (reconcile.Result, error) {
 		reconcilers.DeploymentConfigContainerResourcesMutator,
 		reconcilers.DeploymentConfigAffinityMutator,
 		reconcilers.DeploymentConfigTolerationsMutator,
-		r.apicastProductionWorkersEnvVarMutator,
-		r.apicastLogLevelEnvVarMutator,
-		r.apicastTracingConfigEnvVarsMutator,
-		r.apicastEnvironmentEnvVarMutator,
+		apicastProductionWorkersEnvVarMutator,
```
miguelsorianod, line 91: Why has it been changed from APIcastReconciler methods to generic functions?

I understand the decision for the `portsMutator` function, which I guess has been done this way because it could be reusable for other reconcilers, but the other ones seem pretty specific to the APIcastReconciler

doc/apimanager-reference.md
miguelsorianod: I see. Got you. I still have some doubts about this though.
Ok, we have the 'gateway' port and the 'httpsproxy' port.
When we enable TLS the httpsproxy port is created and used in both the APIcast pod and APIcast service, which means that someone could use the Service to access APIcast through that pot in the `Service`.
However, the components inter-communication would still be through HTTP and not through HTTPS right? is that the intended behavior that we want? If that's the intended behavior I think we should clarify this in the documentation because a user might think/interpret that when they set TLS on APIcast that means that automatically all communication to it would be encrypted, which would not be the case.

pkg/3scale/amp/operator/apicast_reconciler.go
eguzki: Just because those functions are global, they do not use anything from the ApicastReconciler struct. All required input is in the input parameters.

doc/apimanager-reference.md
eguzki: Ok, The operator does not provide a way to "disable" the HTTP when TLS is enabled. That is wanted?

eguzki: Added doc  to make this point clear

eguzki: https://github.com/3scale/3scale-operator/pull/632/files?file-filters%5B%5D=.go&file-filters%5B%5D=.md#diff-72dab7f3126dc7dbd85ab6398f36d48315960176cddc220c7fd9bdb36da97fb6R6

miguelsorianod: :+1:

pkg/3scale/amp/operator/apicast_reconciler.go
miguelsorianod: I understand it from that point of view. I think I would still make them ApicastReconciler methods even in that case because they are specific to it and because they are namespaced to that type instead of being namespaced at the package level, but I am ok with both in any case, not a big deal

miguelsorianod (approved):

PR #640, opened 2021-07-19
[backport 628] Rename operator components images to RELATED_IMAGES to support digest pinning

miguelsorianod:


miguelsorianod:
test-e2e failing is expected because we need to do the change in OpenShift CI to use the new envvar names https://github.com/openshift/release/pull/20166, which will not work until we merge this. So we are going to intentionally merge this with test-e2e failing and once that other PR works it should fix the test-e2e

PR #657, opened 2021-09-01
Bump release 2.12

eguzki:
* operator version bump: 0.9.0
* upgrade procedure reset

### DIFF
--- a/bundle/manifests/3scale-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/3scale-operator.clusterserviceversion.yaml
@@ -342,7 +342,7 @@ spec:
               labels:
                 com.redhat.component-name: 3scale-operator
                 com.redhat.component-type: infrastructure
-                com.redhat.component-version: 0.8.0
+                com.redhat.component-version: 0.9.0
                 com.redhat.product-name: 3scale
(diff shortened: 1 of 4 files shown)


miguelsorianod:
The following lines should be removed as they were a procedure specific to the 2.11 release that had to be done at the same time as the backend-redis DeploymentConfig update:
https://github.com/3scale/3scale-operator/blob/d0f1f208b3cc039398a0a61026057233a8ae2de3/pkg/3scale/amp/operator/upgrade.go#L280-L290

Same applies to:
https://github.com/3scale/3scale-operator/blob/d0f1f208b3cc039398a0a61026057233a8ae2de3/pkg/3scale/amp/operator/upgrade.go#L331-L341

The following methods could be removed too as they were specific to 2.11:
* `ensureRedisCommand`
* `ensureRedisPodTemplateLabels`
* `upgradeBackendCronDeploymentConfigMemoryLimits`
* `ensureBackendCronMemoryLimit`
* `backendCronResourcesOverriden`

miguelsorianod (changes requested):
See previous comment

eguzki:
requested changes applied

miguelsorianod (approved):
:+1:

PR #660, opened 2021-09-13
fix docker image version parser

eguzki:
* Keep backward compat. New unittests added, not updated any existing unittests.
* Full check for valid label value, not only length.

### DIFF
--- a/pkg/helper/version_parser.go
+++ b/pkg/helper/version_parser.go
@@ -1,7 +1,10 @@
 package helper
 
 import (
+	"fmt"
+	"net/url"
 	"regexp"
+	"strings"
 )
 
 type versionParser func(string) string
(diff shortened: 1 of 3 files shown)


pkg/helper/version_parser.go
```diff
@@ -19,11 +22,32 @@ var (
...
 )
 
+func normalizeDockerImage(image string) string {
+	newImage := image
+
+	if strings.Index(image, "/") < 0 {
+		newImage = fmt.Sprintf("docker.io/%s", newImage)
+	}
+
+	if !strings.HasPrefix(newImage, "http://") && !strings.HasPrefix(newImage, "https://") {
```
miguelsorianod, line 32: I am not sure we should take into account or assume anything about the protocol:

Looking at the docker documentation https://docs.docker.com/engine/reference/commandline/pull/#pull-from-a-different-registry there does not seem to be a protocol part available to specify.

The only way that seems to be available is to configure the corresponding registry as an "insecure" (https://docs.docker.com/engine/reference/commandline/dockerd/#insecure-registries) one to force the connection to http. Unless that is done https is used. In any case the protocol does not seem to be part of the image specification

Also trying to pull specifying a protocol gives me an invalid reference format error:
```
msoriano@localhost:~/go/src/github.com/3scale/3scale-operator (master)$ docker pull https://registry.redhat.io/ubi8/go-toolset
invalid reference format
```

miguelsorianod: :+1: I agree we should not do more work on this at the moment as it could be removed in the near future.
Could we put some comments around the `normalizeDockerImage` function to clearly scope that should / is only used on parsing and that the protocol is just to make the parser work properly but not really used to set the image?

eguzki: added comment and refactored a bit to make clear that normalizing is juas a mean to get the `path` of the docker image

miguelsorianod: :+1: thanks!

miguelsorianod (approved):
looks good to me.
Are you thinking on backporting this to 2.11?

PR #667, opened 2021-10-01
enable operator metric service and servicemonitor

eguzki:
https://issues.redhat.com/browse/THREESCALE-7644

TODO
- [ ] test upgrade procedure. There was a bug in OLM where upgrading a CSV fails when providing a Service object as part of the bundle manifests. We should reenable this once that bug is fixed
- [ ] deploy with olm and check service and servicemonitor


eguzki:
The upgrade procedure worked. New service for the metrics was created and old (2.10) metrics service was deleted. Same for the servicemonitor

For few seconds, they co-existed.
```
$ k get servicemonitors
NAME                                                     AGE
threescale-operator-controller-manager-metrics-monitor   1s
threescale-operator-metrics                              109s
```
Maybe it helped that the name changed, and if the name is the same, the upgrade breaks :) but after the operator-sdk version changes, the name of the service changed and the upgrade succeed.

Same for services

```
$ k get services
NAME                                                     TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)             AGE
threescale-operator-controller-manager-metrics-service   ClusterIP   172.30.40.66   <none>        8443/TCP            4s
threescale-operator-metrics                              ClusterIP   172.30.92.90   <none>        8383/TCP,8686/TCP   2m
```

After few seconds, old service and servicemonitor was deleted

```
k get services
NAME                                                     TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)    AGE
threescale-operator-controller-manager-metrics-service   ClusterIP   172.30.40.66   <none>        8443/TCP   9s
```

```
k get servicemonitors
NAME                                                     AGE
threescale-operator-controller-manager-metrics-monitor   16s
```

miguelsorianod:
:+1:. Could you test an upgrade with the same name?
The originally reported issue happened when the name is the same https://bugzilla.redhat.com/show_bug.cgi?id=1898500.

I see a fix for it was backported to 4.7 looking at that issue but I am not sure if it was backported to 4.6

miguelsorianod:
I think I found the fix for 4.6 too. It seems it was included in 4.6.16: https://bugzilla.redhat.com/show_bug.cgi?id=1920873
It would be good to verify it

PR #681, opened 2021-11-03
http client respects http proxy env vars

eguzki:
Currently the http client used for application capabilities was not respecting the http proxy env variables. 

https://pkg.go.dev/net/http#ProxyFromEnvironment


miguelsorianod:
looks good to me

miguelsorianod (approved):

PR #682, opened 2021-11-03
Update system-mysql internal DB from MySQL 5.7 to 8.0

miguelsorianod:
The change of image also implies an automated upgrade from
MySQL 5.7 to 8.0 when coming from an older 3scale installation
where 5.7 is used. There is no need for a custom upgrade procedure.

The data migration is performed automatically when changing the image.

On the product side, the image to be set will be:
`registry.redhat.io/rhscl/mysql-80-rhel7:8.0`

JIRA: https://issues.redhat.com/browse/THREESCALE-7786
which depends on https://issues.redhat.com/browse/THREESCALE-6825

### Verification steps

Deploy 3scale with the APIManager
```yaml
k apply -f - <<EOF
kind: APIManager
metadata:
  name: apimanager
spec:
  wildcardDomain: ns.apps.example.com
```

Wait for the deployment to be ready

```
oc wait --for=condition=available apimanager/apimanager --timeout=-1s
```
Check mysql 8 is running

```
oc exec -t dc/system-mysql -- /opt/rh/rh-mysql80/root/usr/libexec/mysqld --version
/opt/rh/rh-mysql80/root/usr/libexec/mysqld  Ver 8.0.21 for Linux on x86_64 (Source distribution)
```

### DIFF
--- a/doc/template-user-guide.md
+++ b/doc/template-user-guide.md
@@ -65,7 +65,7 @@ oc new-app --file pkg/3scale/amp/auto-generated-templates/amp/amp.yml \
 | **ZYNC_DATABASE_IMAGE** | Zync's PostgreSQL image to use | centos/postgresql-10-centos7 |
 | **MEMCACHED_IMAGE** | Memcached image to use | memcached:1.5 |
 | **IMAGESTREAM_TAG_IMPORT_INSECURE** | the server may bypass certificate verification | false |
-| **SYSTEM_DATABASE_IMAGE** | System MySQL image URL | centos/mysql-57-centos7 |
+| **SYSTEM_DATABASE_IMAGE** | System MySQL image URL | centos/mysql-80-centos7 |
 | **REDIS_IMAGE** | Redis image to use | centos/redis-5-centos7 |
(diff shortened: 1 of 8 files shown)


miguelsorianod:
At the moment this cannot be merged due to system's pre-hook pod is failing to start after this changes.
This issue has been identified in https://issues.redhat.com/browse/THREESCALE-6825 and this issue cannot be merged until that issue is resolved and this PR is tested again after those changes to verify that all works. The PR will need to be tested with both the upstream and product image, including the upgrade between mysql 5.7 and 8

eguzki (dismissed):
this was a very easy one :)

eguzki:
starting command line or env vars... all remain the same from 5.7 to 8? Seems like 5.8 more than 8 :)

miguelsorianod:
> starting command line or env vars... all remain the same from 5.7 to 8? Seems like 5.8 more than 8.0 :)

It should remain the same. It's 8.0. MySQL 5.8 does not exist.

eguzki:
@MStokluska  added verification steps

MStokluska (approved):
Looks good to me, verified changes on cluster and code looks good as well

PR #686, opened 2021-11-05
Remove usage of system redis messagebus from System

miguelsorianod:
This PR removes the usage of System's Redis message bus database from System DCs (DeploymentConfigs).

* Removes all redis message-bus related environment variable references from the following DCs:
  * system-app
  * system-sidekiq
  * system-sphinx
* In case databases deployment mode is internal it removes message-bus related attributes from the system-redis K8s Secret. In case databases deployment mode is external then this attributes remain unchanged.
* Implements an upgrade procedure that performs the previously explained changes

This is done for both operator and templates.

Templates will need an upgrade procedure for this change.

### DIFF
--- a/pkg/3scale/amp/operator/upgrade.go
+++ b/pkg/3scale/amp/operator/upgrade.go
@@ -45,6 +45,14 @@ func (u *UpgradeApiManager) Upgrade() (reconcile.Result, error) {
 		return res, nil
 	}
 
+	res, err = u.deleteMessageBusConfigurations()
+	if err != nil {
+		return res, fmt.Errorf("Upgrade: delete message bus configurations: %w", err)
+	}
+	if res.Requeue {
+		return res, nil
+	}
+
 	return reconcile.Result{}, nil
 }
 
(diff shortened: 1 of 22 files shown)


miguelsorianod:
cc @akostadinov too to review that conceptual changes are correct and that we are not accidentally removing elements needed by Zync, which we understand will still be using message bus.

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -524,6 +532,214 @@ func (u *UpgradeApiManager) upgradeSystemPostgreSQLImageStream() (reconcile.Resu
...
+			}
+		}
+	}
+
+	// post-hook pod
+	var postHookPod *appsv1.ExecNewPodHook
+	if existing.Spec.Strategy.RollingParams != nil && existing.Spec.Strategy.RollingParams.Post != nil && existing.Spec.Strategy.RollingParams.Post.ExecNewPod != nil {
+		postHookPod = existing.Spec.Strategy.RollingParams.Post.ExecNewPod
+	}
+	if existing.Spec.Strategy.RecreateParams != nil && existing.Spec.Strategy.RecreateParams.Post != nil && existing.Spec.Strategy.RecreateParams.Post.ExecNewPod != nil {
```
eguzki, line 703: AFAIK `systrem-app`, `system-sphix` and `system-sidekiq` do not use `Spec.Strategy.RecreateParams `

eguzki (changes requested):

pkg/3scale/amp/operator/upgrade.go
miguelsorianod: :+1: I was aware. The method was intended to be generic/reusable so I added that part too

eguzki: Is it being used by any other 3scale deploymentconfig? Not that I am aware of, But it could be. 

One more (small) thing, the code is replacing the `postHookPod` (or `preHookPod` earlier) in case there are recreateParams. I guess the strategy has to be either `rollingParams` (exclusive) OR `recreateParams`. Then, shouldn't it be `else if`?

eguzki:
> cc @akostadinov too to review that conceptual changes are correct and that we are not accidentally removing elements needed by Zync, which we understand will still be using message bus.

Looking at the code, Zync never used message bus (redis queue). Am I right?

miguelsorianod:
> Looking at the code, Zync never used message bus (redis queue). Not even redis at all. Am I right?

Zync never used redis message bus DB but apparently it uses some kind of message bus if I understood correctly.

pkg/3scale/amp/operator/upgrade.go
miguelsorianod: > Is it being used by any other 3scale deploymentconfig? Not that I am aware of, But it could be.
It isn't.

The strategy is exclusive. It could be `else if`. This should never happen though so having just another `if` is another issue. Even if it happens only one is used due to the type field so it would not be an issue

akostadinov:
Not very familiar with operator but I can't spot anything zync would be concerned with.

eguzki (approved):
rebase before merge, please

miguelsorianod:
Done :+1:

PR #692, opened 2021-11-22
Delete kube-rbac-proxy container from controller-manager

eguzki:
kube-rbac-proxy is used by default on operator-sdk to protect operator metrics path, in case you don't want anyone in the cluster but only k8s authenticated resources can access to them.

However there are some issues with the kube-rbac-proxy:
* The proxy serves metrics in HTTPS and prometheus needs to be conffigured with TLS skipVerify enabled. The following fix needs to be applied:
```yaml
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      tlsConfig:
+        insecureSkipVerify: true
```
However **OpenShift User Workload Monitoring** (OpenShift official monitoring stack) does not support ServiceMonitor with bearerTokenFile field [Prometheus ServiceMonitor failing to scrape operator metrics served though kube-proxy HTTPS 8443 port operator-framework/operator-sdk#4764 (comment)](https://github.com/operator-framework/operator-sdk/issues/4764#issuecomment-821264658) (which is needed to scrape metrics behind kube-rbac-proxy) , so it seems there is no way of having operator metrics with auth if using OCP UWM. We are looking at leveraging OCP UWM for a better integration with openshift.

* Prometheus instances require extra permissions to scrape, this is the required cluster scope role to scrape operator metrics:
```yaml
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: threescale-operator-metrics-reader
rules:
- nonResourceURLs:
  - /metrics
  verbs:
  - get
```

* In addition, Openshift monitoring team is considering a different way of securying metrics. Bear in mind that using bearer tokens for metrics authn puts additional load on the API server and they are looking at replacing this by client TLS auth in the future (it's being discussed here: [openshift/enhancements#701](https://github.com/openshift/enhancements/pull/701))

The downside is that the operator metrics could be read by anyone within the cluster. Two comments about this:
* 3scale components metrics are currently, "open". Does not make much sense to have operator metrics secured and other metrics open.
* ServiceMonitor needs to be configured with TLS cert verification disabled. So, kube-rbac-proxy enforces security, but the connection is not verified. 

Having into account all this information, we have decided to remove the kube-rbac-proxy for now. The alternative approach would be the fix in https://github.com/3scale/3scale-operator/pull/694. We have decided to go for disabling the auth proxy.


config/default/manager_metrics_patch.yaml
```diff
@@ -0,0 +1,18 @@
+# This patch exposes metrics endpoint in plain HTTP 8080 port
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: controller-manager
+  namespace: system
```
miguelsorianod, line 7: I understand the namespace should not be set

config/manager/metrics_service.yaml
```diff
@@ -0,0 +1,15 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    control-plane: controller-manager
+  name: controller-manager-metrics-service
+  namespace: system
```
miguelsorianod, line 8: I understand the namespace should not be set

config/default/manager_metrics_patch.yaml
eguzki: At least, we know it does not have any effect.

The auth proxy patch has the same namespace https://github.com/3scale/3scale-operator/blob/1b7b317fda2d4e0c1022d7e053dfc1ecb6956945/config/default/manager_auth_proxy_patch.yaml#L7

```yaml

apiVersion: apps/v1
kind: Deployment
metadata:
  name: controller-manager
  namespace: system
spec:
```

miguelsorianod: :+1:

miguelsorianod (approved):

PR #694, opened 2021-11-22
Fix operator metrics servicemonitor

eguzki:
kube-rbac-proxy is used by default on operator-sdk to protect operator metrics path, in case you don't want anyone in the cluster but only k8s authenticated resources can access to them.

Currently, the service monitor is incorrectly configured and prometheus cannot scrape metrics. The service monitor needs to be fixed to specify:
* use HTTPS scheme 
* Use bearer token for authenticated calls (or kube-rbac-proxy will return 403 Forbidden)

Besides, the documentation has been enhanced to ensure prometheus instances have required RBAC permission to scrape operator metrics. 

Related issues:
* https://github.com/operator-framework/operator-sdk/issues/4764

**Warning**: with this fix, the prometheus instance from the openshift user-workload-monitoring stack does not scrape metrics from the operator. Apparently, a security policy configured for the user workload prometheus instance prevents from reading bearer token from  local file system. Thus, the service monitor is being discarded. The error shown in the prometheus operator:
```
level=warn ts=2021-11-21T14:41:30.167698607Z caller=operator.go:1753 component=prometheusoperator msg="skipping servicemonitor" error="it accesses file system via bearer token file which Prometheus specification prohibits" servicemonitor=eguzki/threescale-operator-controller-manager-metrics-monitor namespace=openshift-user-workload-monitoring prometheus=user-workload
```

### DIFF
--- a/bundle/manifests/threescale-operator-controller-manager-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml
+++ b/bundle/manifests/threescale-operator-controller-manager-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml
@@ -6,8 +6,12 @@ metadata:
   name: threescale-operator-controller-manager-metrics-monitor
 spec:
   endpoints:
-  - path: /metrics
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    path: /metrics
     port: https
+    scheme: https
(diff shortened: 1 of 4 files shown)


eguzki:
Closed in favor of https://github.com/3scale/3scale-operator/pull/692

WE have decided it is a better option to for for #692. Reason explained in the other PR

PR #712, opened 2022-02-01
disable openapi schema format validation

eguzki:
As per the OAS: https://spec.openapis.org/oas/v3.0.0#data-types

```
the format property is an open string-valued property, and can have any value. Formats such as "email", "uuid", and so on, MAY be used even though undefined by this specification.
```

The following openapi Doc schema is not successfully validated:
```yaml
parameters:
   - name: customer_uuid
          in: query
          description: Unique customer identifier to be used for experience customization.
          schema:
            type: string
            format: uuid
```

The error from the operator:
```
spec.openapiRef.secretRef: Invalid value: v1.ObjectReference{Kind:"",
        Namespace:"3scale-cand-3", Name:"uberoas3", UID:"", APIVersion:"",
        ResourceVersion:"", FieldPath:""}: invalid paths: parameter
        "customer_uuid" schema is invalid: Unsupported 'format' value 'uuid'
```

Issue: https://issues.redhat.com/browse/THREESCALE-7642

### DIFF
--- a/main.go
+++ b/main.go
@@ -24,6 +24,7 @@ import (
 	"runtime"
 
 	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/getkin/kin-openapi/openapi3"
 	grafanav1alpha1 "github.com/integr8ly/grafana-operator/v3/pkg/apis/integreatly/v1alpha1"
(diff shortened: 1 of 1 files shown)


KevFan (approved):
Changes looks good to me :+1:

PR #715, opened 2022-02-11
MGDAPI-3467 tenant deletion support

MStokluska:
- adds support of deleting tenants via CR
  * https://issues.redhat.com/browse/MGDAPI-3467
  * https://issues.redhat.com/browse/THREESCALE-6051

## Verification:
For RHOAM:
- install RHOAM locally from this branch: https://github.com/integr8ly/integreatly-operator/pull/2409 but replace the manifests/3scale/0.9.0 values for containerImage and image from `quay.io/austincunningham/3scale-operator:cluster-scoped` to `quay.io/mstoklus/3scale-operator:v0.12.0`
- wait for install to complete 
- create tenant via CR
- confirm tenant got created in 3scale 
- delete tenant CR
- confirm tenant has been marked for deletion in 3scale


makslion:
Installed RHOAM using the `quay.io/mstoklus/3scale-operator:v0.9.0` 3scale image:
Screenshot from 2022-02-15 11-33-57
Created a tenant CR:
Screenshot from 2022-02-15 11-34-46
And watched a tenant being created in the 3scale:
Screenshot from 2022-02-15 11-43-43
After deleting a tenant CR the tenant was scheduled for a deletion:
Screenshot from 2022-02-15 11-45-16

controllers/capabilities/tenant_internal_reconciler.go
```diff
@@ -90,6 +137,14 @@ func (r *TenantInternalReconciler) reconcileTenant() (*porta_client_pkg.Tenant,
 	return tenantDef, nil
 }
 
+func (r *TenantInternalReconciler) confirmTenantDeleted(tenant *porta_client_pkg.Tenant) (bool, error) {
+	if tenant.Signup.Account.State == "scheduled_for_deletion" {
+		return true, nil
+	} else {
```
makslion, line 143: The only complaint from a [CodeClimate](https://codeclimate.com/github/3scale/3scale-operator/pull/715) that is related to this change is this `else` statement not being on a new line, but I don't think it matters that much :thinking:

makslion:
Other failed checks are related to the `k8s.io/utils` licence approval :thinking: Anyway, the code looks good and performs well, hence

/lgtm

eguzki (changes requested):
IMO the tenant deletion support should be done at the `tenant_controller.go` on the entrypoint of reconciler.

The finalizer pattern used across other controllers we are working on is:

```go
func (r *TenantReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {                                  
    _ = context.Background()                                                                                   
    reqLogger := r.Log.WithValues("tenant", req.NamespacedName)                                                
                                                                                                               
    // Fetch the Tenant instance                                                                               
    tenantR := &capabilitiesv1alpha1.Tenant{}                                                                  
    err := r.Client.Get(context.TODO(), req.NamespacedName, tenantR)                                           
    if err != nil {                                                                                            
        if errors.IsNotFound(err) {                                                                            
            // Request object not found, could have been deleted after reconcile request.                      
            // Owned objects are automatically garbage collected. For additional cleanup logic use finalizers. 
            // Return and don't requeue                                                                        
            reqLogger.Info("Tenant resource not found")                                                        
            return ctrl.Result{}, nil                                                                          
        }                                                                                                      
        // Error reading the object - requeue the request.                                                     
        return ctrl.Result{}, err                                                                              
    }       
    // Tenant has been marked for deletion                                                      
    if tenantR.GetDeletionTimestamp() != nil && controllerutil.ContainsFinalizer(tenantR, tenantFinalizer) {
         // your deletion handling
         .....
         
         // 
         //Remove finalizer and update the object.             
         controllerutil.RemoveFinalizer(tenantR, tenantFinalizer)   
         err = r.Client.Update(ctx, tenantR)                     
         logger.Info("Removing finalizer", "error", err)       
         return ctrl.Result{Requeue: true}, err                
     }

       // Ignore deleted resources, this can happen when foregroundDeletion is enabled                                     
 // https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#foreground-cascading-deletion      
        if tenantR.GetDeletionTimestamp() != nil {                                                                             
              return ctrl.Result{}, nil                                                                                       
         } 
 
          if !controllerutil.ContainsFinalizer(tenantR, tenantFinalizer) {                    
                   controllerutil.AddFinalizer(tenantR, tenantFinalizer)                           
                    err := r.Client.Update(ctx, tenantR)                                         
                    logger.Info("Adding finalizer", "error", err)                              
                    return ctrl.Result{Requeue: true}, err                                     
            }                                                                              
            ...
                                                                                                                           
```


                                                                                                       

```

File: pkg/controller/helper/controllerUtil.go [hunk 2]
```diff
@@ -0,0 +1,305 @@
+/*
```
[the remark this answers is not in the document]
MStokluska: Good point, reason why I've used it is because it's slightly extended version of it, but I did refactored to use the one from controllerutil

controllers/capabilities/tenant_internal_reconciler.go
```diff
@@ -17,6 +17,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	tenantFinalizer = "tenant.3scale.net/finalizer"
```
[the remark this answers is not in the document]
MStokluska: agree, updated

controllers/capabilities/tenant_controller.go
```diff
@@ -98,6 +102,39 @@ func (r *TenantReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{}, err
 	}
 
+	// Tenant has been marked for deletion
+	if tenantR.GetDeletionTimestamp() != nil && controllerutil.ContainsFinalizer(tenantR, tenantFinalizer) {
+		// your deletion handling
+		err := controllerhelper.DeleteTenant(tenantR, portaClient)
```
eguzki, line 108: This is the most critical part of the code. It should be idempotent and expect errors at any operation.

For the initial implementation, I would say that when this op `DeleteTenant` does not return error, it means that the 3scale API confirmed that eventually the tenant will be deleted (or has been deleted). 

What is the idea to check confirmation?

Let's say this is what happens:
- Customer deletes the CR
- 3scale operator sends HTTP request to delete the tenant
- When checking to confirm, 3scale API says it is not deleted. 
 
The 3scale operator will not retry and the CR will be "tagged" to delete but actually never deleted because the finalizer was not removed

My logic proposal is:
- Customer deletes the CR
- 3scale operator sends HTTP request to delete the tenant
- If the response is DONE (200) or DOES NOT EXIST(404) or whatever http status code is 3scale API returning when trying to delete a tenant already in "scheduled for deletetion", then remove the finalizer
- Any other issue, connection issue or http 5xx response, then retry.

eguzki (changes requested):

controllers/capabilities/tenant_controller.go
```diff
@@ -73,17 +77,6 @@ func (r *TenantReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{}, err
 	}
 
-	changed := tenantR.SetDefaults()
```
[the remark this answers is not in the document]
MStokluska: this is from previous commit. That's already back in.

controllers/capabilities/tenant_controller.go
```diff
@@ -98,6 +91,39 @@ func (r *TenantReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{}, err
 	}
 
+	// Tenant has been marked for deletion
+	if tenantR.GetDeletionTimestamp() != nil && controllerutil.ContainsFinalizer(tenantR, tenantFinalizer) {
```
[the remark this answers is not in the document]
MStokluska: good point. Thanks.

controllers/capabilities/tenant_internal_reconciler.go
```diff
@@ -67,12 +68,12 @@ func (r *TenantInternalReconciler) Run() error {
...
 // * tenant's attributes will be updated if required
 func (r *TenantInternalReconciler) reconcileTenant() (*porta_client_pkg.Tenant, error) {
-	tenantDef, err := r.fetchTenant()
+	tenantDef, err := controllerhelper.FetchTenant(r.tenantR, r.portaClient)
 	if err != nil {
 		return nil, err
 	}
 
-	if tenantDef == nil {
+	if tenantDef == nil && controllerutil.ContainsFinalizer(r.tenantR, tenantFinalizer){
```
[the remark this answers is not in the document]
MStokluska: My idea with this was, when this is called https://github.com/3scale/3scale-operator/blob/569e81471367de84c856b1563a4814ff329e25de/controllers/capabilities/tenant_controller.go#L130 I was hoping to remove the check of whether the finalizer is present and call the reconcileFinalizers regardless, where the finalizers would be updated only if not present. But I guess if we want to wrap the reconcileFinalizers there's no point in having it as createOrUPdate.

pkg/controller/helper/finalizers.go
```diff
@@ -0,0 +1,36 @@
...
+- k8client
+- finalizer
+If the deletion timestamp is found, the finalizer will be removed.
+If the deletion timestamp is not present, a finalizer will be reconciled
+*/
+func ReconcileFinalizers(object controllerutil.Object, client client.Client, finalizer string) error {
+	var err error
+
+	if object.GetDeletionTimestamp() == nil {
+		_, err = controllerutil.CreateOrUpdate(context.TODO(), client, object, func() error {
```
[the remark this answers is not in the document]
MStokluska: My idea with this was, when this is called

[3scale-operator/controllers/capabilities/tenant_controller.go](https://github.com/3scale/3scale-operator/blob/569e81471367de84c856b1563a4814ff329e25de/controllers/capabilities/tenant_controller.go#L130)

Line 130 in [569e814](https://github.com/3scale/3scale-operator/commit/569e81471367de84c856b1563a4814ff329e25de)
 if !controllerutil.ContainsFinalizer(tenantR, tenantFinalizer) { 
I was hoping to remove the check of whether the finalizer is present and call the reconcileFinalizers regardless, where the finalizers would be updated only if not present. But I guess if we want to wrap the reconcileFinalizers there's no point in having it as createOrUPdate.

pkg/controller/helper/finalizers.go
```diff
@@ -0,0 +1,36 @@
...
+			controllerutil.AddFinalizer(object, finalizer)
+			return nil
+		})
+	} else {
+		_, err = controllerutil.CreateOrUpdate(context.TODO(), client, object, func() error {
```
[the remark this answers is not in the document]
MStokluska: as above

controllers/capabilities/tenant_controller.go
MStokluska: Following the scenario above, I've spotted where the issue could be. 
Think this would work better than checking the error codes?
```
		// delete tenant
		err := controllerhelper.DeleteTenant(tenantR, portaClient)
		if err != nil {
			return ctrl.Result{}, err
		}

		// confirm tenant deletion
		isTenantDeleted, err := controllerhelper.ConfirmTenantDeleted(tenantR, portaClient)
		if err != nil {
			return ctrl.Result{}, err
		}

		if !isTenantDeleted {
			return ctrl.Result{Requeue: true}, nil
		}
		
		// add or remove finalizer
		err = controllerhelper.ReconcileFinalizers(tenantR, r.Client, tenantFinalizer)
		if err != nil {
			return ctrl.Result{}, err
		}

		return ctrl.Result{}, nil
		```

File: controllers/capabilities/tenant_controller.go [hunk 10]
```diff
@@ -98,6 +91,39 @@ func (r *TenantReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+			// add or remove finalizer
+			err = controllerhelper.ReconcileFinalizers(tenantR, r.Client, tenantFinalizer)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+		}
+
+		return ctrl.Result{}, nil
+	}
+
```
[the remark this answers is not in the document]
MStokluska: Although, there might an issue with backward compatibility, with CRs where the finalizers are not present. Adding it in.

MStokluska:
@eguzki I've addressed all your comments and explained the reasoning behind having it done in a certain way in the first place, I would appreciate a review again. 
The only thing that I question is around the requeue if by any chance the 3scale api call fails on removing the tenant. But it might be easier to go through the scenarios on call. 
Thanks for the reviews :)

controllers/capabilities/tenant_internal_reconciler.go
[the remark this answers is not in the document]
MStokluska: Done, thanks

pkg/controller/helper/tenant.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+	return tenantDef, nil
+}
+
+/*
+DeleteTenant deletes tenant from 3scale
+- tenant
+- portaClient
+*/
+func DeleteTenant(tenant *capabilitiesv1alpha1.Tenant, portaClient *porta_client_pkg.ThreeScaleClient) error {
```
eguzki, line 34: nitpick: this method does not add anything, does it?

controllers/capabilities/tenant_controller.go
```diff
@@ -98,6 +97,56 @@ func (r *TenantReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 
+	// Tenant has been marked for deletion
+	if tenantR.GetDeletionTimestamp() != nil && controllerutil.ContainsFinalizer(tenantR, tenantFinalizer) {
+		existingTenant, err := controllerhelper.FetchTenant(tenantR, portaClient)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
+		// do not attempt to delete tenant that is already scheduled for deletion
+		if existingTenant.Signup.Account.State != scheduledForDeletionState {
```
eguzki, line 108: What if `existingTenant` is `nil`?

eguzki (changes requested):
Overall looks good

I left some comments

controllers/capabilities/tenant_controller.go
```diff
@@ -73,17 +83,6 @@ func (r *TenantReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
-		err = r.Client.Update(context.TODO(), tenantR)
-		if err != nil {
-			return ctrl.Result{}, err
-		}
-		reqLogger.Info("Tenant resource updated with defaults")
-		// Expect for re-trigger
-		return ctrl.Result{}, nil
-	}
-
 	masterAccessToken, err := r.FetchMasterCredentials(r.Client, tenantR)
```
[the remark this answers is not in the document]
MStokluska: Reason why it's done that way is because the deletion required the portaClient and finalizers can only be deleted if the portaClient tenant delete succeeds. If we are to do it the other way around, we are risking deleting finalizers before tenant is removed from 3scale. How do you think we could resolve this?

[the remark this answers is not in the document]
eguzki: Right, I changed my mind. It is correct as it is now. 

If there is an error building the portaClient (involves reading a secret), it will retry. 

The only "edge" case here is that is some other controller adds its own finalizer, our controller will still read secrets and build portaClient even when the delete job is done and our finalizer is removed. But a small issue IMO.

controllers/capabilities/tenant_controller.go
```diff
@@ -98,6 +97,61 @@ func (r *TenantReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+		return ctrl.Result{}, nil
+	}
+
+	// Ignore deleted resources, this can happen when foregroundDeletion is enabled
+	// https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#foreground-cascading-deletion
+	if tenantR.GetDeletionTimestamp() != nil {
+		return ctrl.Result{}, nil
+	}
+
+	changed := tenantR.SetDefaults()
```
eguzki, line 136: After all the finalizer logic?

controllers/capabilities/tenant_controller.go
```diff
@@ -98,6 +97,61 @@ func (r *TenantReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+		// Expect for re-trigger
+		return ctrl.Result{}, nil
+	}
+
+	if !controllerutil.ContainsFinalizer(tenantR, tenantFinalizer) {
+		err = controllerhelper.ReconcileFinalizers(tenantR, r.Client, tenantFinalizer)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
```
eguzki, line 152: should be `return ctrl.Result{Requeue: true}, nil` or it will stop reconciling and the tenant will not be created

eguzki (changes requested):

controllers/capabilities/tenant_controller.go
MStokluska: That's not the case, the tenants are created and I think (correct me if this is wrong) it's because by editing the tenant CR the reconciler is triggered again, but setting requeue: true, we are going to have 2 runs, one after finalizer is added, and one coming from the update done to tenant CR

pkg/controller/helper/tenant.go
MStokluska: Potentially it could be moved, although I was thinking it would be nice to have it in a single file, I was thinking of possibility moving the tenant create here too, but happy to remove it directly if that's what you preffer.

controllers/capabilities/tenant_controller.go
eguzki: When I create the tenant CR, the operator will get a reconciliation event. The controller will add the finalizer and stop the reconciliation job. No new reconciliation will be done until there is a update or delete event.  So, effectively, the tenant will not be created in 3scale until you edit the CR after having created that. 

Maybe you can test and tell me if I am wrong

controllers/capabilities/tenant_controller.go
eguzki: it is about step order:
- finalizer logic
- setting defaults
- reconciliate 3scale tenants

Setting the finalizer should be first thing to do, otherwise while the operator is doing "things", the user could remove the CR and the operator would not be able to handle the deletion.

controllers/capabilities/tenant_controller.go
MStokluska: I did test it and it works as expect. The way I see it adding finalizer is also an update to CR

pkg/controller/helper/tenant.go
MStokluska: Updated

eguzki:
Looks good

I think we should tackle the issue of cascade deletion of 3scale tenant objects (DeveloperUser, DeveloperAccount, Products, Backends, ActiveDocs) in another PR. 

I am aware that the code coverage in the capabilites code is near to zero, but it is prepared to have tests (https://github.com/3scale/3scale-operator/blob/62f241c29eb4136e055269cf31ab7062c16e6bfa/controllers/capabilities/suite_test.go). If you are willing to add some tests, feel free.

Regarding upstream doc, I think it would be nice to have a note in the capabilities doc about the new feature. Also will be used by tech writers team for the official doc.

eguzki (approved):

eguzki:
good job :medal_military:

PR #720, opened 2022-02-23
Reconcile apicast environment

eguzki:
### what

The following APIManager fields are not reconciled. 
* `spec.apicast.managementAPI`
* `spec.apicast.openSSLVerify`
* `spec.apicast.responseCodes`

The root reason is that those fields in the APIMamanger CR are reconciled in the `apicast-environment` config map. Then the apicast production and staging deploymentconfigs have env vars as references to fields from that config map. When the APIManager CR fields are changed, the config map is changed, but the apicast deployments are not rolled out, so changes do not take effect. 

This PR makes apicast deployments being rolled out when those APIManager CR fields change their value.

### how

The ideal fix would be to simplify the deployment. The env var defined in the deploymentconfig object should be set with the value of the APIManager CR. This fix makes the rolling out of the deployment when the CR values changes automatically. However, this solution implies removing the configmap `apicast-environment`. This removal would imply upgrade procedures in the operator based deployments and template based deployments. In order to avoid upgrade procedures, the fix is otherwise. The apicast pods will have one annotation with a hash value computed from the content of the configmap. If the fields of the APIManager CR change, the hash will change and the deployment will rollout. 

### Verification Steps

```yaml
k apply -f - <<EOF
---
kind: APIManager
metadata:
  name: apimanager1
spec:
  wildcardDomain: example.com
EOF
```
Wait for deployment to be available

```
$ oc wait --for=condition=available apimanager/apimanager1 --timeout=-1s
apimanager.apps.3scale.net/apimanager1 condition met
```

Patch apicast field

```
k patch apimanager apimanager1 --type='merge' -p '{"spec":{"apicast":{"managementAPI": "debug"}}}'
```

Apicast deployments should be rolling out

```
$ k get pods
NAME                          READY   STATUS              RESTARTS      AGE
apicast-production-8-5zj92    1/1     Running             0             5m18s
apicast-production-9-deploy   0/1     ContainerCreating   0             2s
apicast-staging-8-vpjcs       1/1     Running             0             5m19s
apicast-staging-9-deploy      0/1     ContainerCreating   0             3s
```

### DIFF
--- a/pkg/3scale/amp/component/apicast.go
+++ b/pkg/3scale/amp/component/apicast.go
@@ -31,6 +31,8 @@ const (
 
 	HTTPSCertificatesMountPath  = "/var/run/secrets/tls"
 	HTTPSCertificatesVolumeName = "https-certificates"
+
+	APIcastEnvironmentConfigMapName = "apicast-environment"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


KevFan (approved):
Verified :+1:  
image

Code looks good to me :+1:

PR #723, opened 2022-03-03
Mgdapi 3472 3

MStokluska:
- Enables support of backend && product deletion (product deletion done by Patryk Stefanski) 
- Enables cascading deletion of product and backend CRs when a tenant CR is deleted

Jira issues:
* https://issues.redhat.com/browse/THREESCALE-6580
* https://issues.redhat.com/browse/THREESCALE-6581


controllers/capabilities/backend_controller.go
```diff
@@ -74,12 +76,57 @@ func (r *BackendReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+
+		// Retrieve ownersReference of tenant CR that owns the Backend CR
+		tenantCR, err := controllerhelper.RetrieveTenantCR(providerAccount, r.Client())
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
+		// If tenant CR is found, set it's ownersReference as ownerReference in the BackendCR CR
+		if tenantCR != nil {
+			err := controllerhelper.SetOwnersReference(backend, r.Client(), tenantCR)
```
eguzki, line 123: Let's add owner references in another PR. IMO out of scope.

Besides:
* Product controller does not add them
* controllerhelper.SetOwnersReference does not need to be implemented. You can call `r.SetOwnerReference(tenantCR, backend)`
* The method `RetrieveTenantCR` has several issues
  * the tenant name parsed from the URL and the `OrganizationName` are case sensitive and might not match. 
  * `OrganizationName`  is not unique and the selected tenant owner might not be the actual tenant owning the backend
  * What I would do is list all the tenants in the same namespace as the backend (or product) and read the secret from the `tenantSecretRef` field. When a secret is found which`adminURL` matches the backend's provider account URL, then you have the tenant owning the backend resource.

controllers/capabilities/backend_controller.go
```diff
@@ -74,12 +76,57 @@ func (r *BackendReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		reqLogger.V(1).Info(string(jsonData))
 	}
 
+	// Ignore deleted Backends, this can happen when foregroundDeletion is enabled
+	// https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#foreground-cascading-deletion
+	if backend.GetDeletionTimestamp() != nil && controllerutil.ContainsFinalizer(backend, backendFinalizer) {
+		err = controllerhelper.ReconcileFinalizers(backend, r.Client(), backendFinalizer)
```
eguzki, line 82: First delete backend from 3scale, if that works, remove the finalizer. 

Have always in mind that the implementation needs to be idempotent and you can expect many reconciliation loop calls to delete the backend.

eguzki (changes requested):
The `removeBackend` is the key here. 

What I would do is to list all the products (in the same namespace as the backend) and remove the current backend from the `spec.BackendUsage` map where it exists. 

I find quite confusing the current implementation. Some clarification would help. Comparing to how I would implement, what is the benefit from this implementation?

I would never update the remote 3scale product entity. The 3scale will do when deleting the backend. 

If the 3scale API call to delete the backend fails, the operator should create a openshift event for visibility. 

The backend usages is not the only reference from the product to the backend being deleted. Application plan limits and pricing rules also have references to backends. Check out my description in the [issue](https://issues.redhat.com/browse/THREESCALE-6580). You are here implementing option **a**

controllers/capabilities/backend_controller.go
```diff
@@ -21,24 +21,26 @@ import (
...
-	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-
 	capabilitiesv1beta1 "github.com/3scale/3scale-operator/apis/capabilities/v1beta1"
 	controllerhelper "github.com/3scale/3scale-operator/pkg/controller/helper"
 	"github.com/3scale/3scale-operator/pkg/helper"
 	"github.com/3scale/3scale-operator/pkg/reconcilers"
 	"github.com/3scale/3scale-operator/version"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
```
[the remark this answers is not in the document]
MStokluska: Thanks! I will update, the refactor comes from running go fmt.

controllers/capabilities/backend_controller.go
MStokluska: good catch. Thank you

controllers/capabilities/backend_controller.go
```diff
@@ -74,12 +76,57 @@ func (r *BackendReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 
+	if !controllerutil.ContainsFinalizer(backend, backendFinalizer) {
+		err = controllerhelper.ReconcileFinalizers(backend, r.Client(), backendFinalizer)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	if backend.Spec.ProviderAccountRef != nil {
```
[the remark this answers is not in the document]
MStokluska: I've added it in because it seems that the spec.ProviderAccountRef is not a required field, so it might be missing. Unless I'm reading it wrong, will have a look again. Thank you

controllers/capabilities/backend_controller.go
MStokluska: Sorry, also, good point in pushing it in another PR. Going to move it to another ticket.

MStokluska:
Hey @eguzki 
So following our conversation here's the list of updates I did:
1. Imports should be in correct order now
2. Product reconciler - removed the ownerReference bits - it will be done in another PR
3. Backend reconciler:
- removed the ownerRef logic - will be done in another PR
- removed calls to get the products from 3scale
- fetching product CRs now works as follows:
a) we first get all the product CRs from the ns where the backend is, as proven during our conversation, they will have to be under the same ns
b) we then loop through products CR to find only the ones that match the providerAcc URL with backend providerAcc URL
c) return only the ones that have the backendUsage mentioned 
- update productCR (removal of backendUsage)
- wait for product reconciler to update product in 3scale
- come back to backend reconciler to remove the backend from 3scale

The issue here however, was that the reconcile finazilers had to be moved only after backend is removed, this means that
backend CR will not be removed if backend fails to remove from 3scale, although to me, this is not an issue and makes sense, but I might be missing something.

controllers/capabilities/backend_controller.go
```diff
@@ -174,3 +202,71 @@ func (r *BackendReconciler) validateSpec(backendResource *capabilitiesv1beta1.Ba
...
+	// update backendUsages for each product retrieved
+	for _, productCR := range tenantProductCRs {
+		delete(productCR.Spec.BackendUsages, backend.Spec.SystemName)
+		err = r.Client().Update(context.TODO(), &productCR)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Attempt to remove backendAPI - expect error on first attempt as the backendUsage has not been removed yet from 3scale
```
eguzki, line 239: We know that this call very likely will fail just after updating product CRs. The product reconciler did not have time to delete backend resources. 

As an idea, what if we tell the controller to retry in, let's say, 2 secs? It would be only when product CRs have been updated.  wdyt?

controllers/capabilities/backend_controller.go
```diff
@@ -174,3 +202,71 @@ func (r *BackendReconciler) validateSpec(backendResource *capabilitiesv1beta1.Ba
...
+	if err != nil {
+		return err
+	}
+
+	// fetch CRs that belong to a tenant
+	tenantProductCRs, err := r.fetchTenantProductCRs(productCRsList, backend)
+
+	// update backendUsages for each product retrieved
+	for _, productCR := range tenantProductCRs {
+		delete(productCR.Spec.BackendUsages, backend.Spec.SystemName)
```
eguzki, line 232: Product CR has more references to the backend than in the backendUsages. I propose a method of the Product struct, to do that. Something like:
```go
func (product *Product) RemoveBackendRefs(backendSystemName string) bool
```

That method will remove, if there exists, references from:
* Application plan pricing rules
* Application plan limits
* Backend Usages

Will return `true` when a change has been made, `fatse` otherwise.

controllers/capabilities/product_controller.go
```diff
@@ -329,3 +355,30 @@ func computeBackendUsageList(list []capabilitiesv1beta1.Backend, backendUsageMap
...
+	if err != nil {
+		return err
+	}
+
+	threescaleAPIClient, err := controllerhelper.PortaClient(providerAccount)
+	if err != nil {
+		return err
+	}
+
+	err = threescaleAPIClient.DeleteProduct(*productResource.Status.ID)
```
eguzki, line 370: Is there a use case where the user can delete a product CR and the status ID is `nil`? I would say it could happen, so I would add a guard at the beginning, if the ID is nil, just return.

eguzki: Same comment applies to the backend CR

controllers/capabilities/backend_controller.go
MStokluska: hey, so I've practiced this a couple of times. 
Looks like the flow already is that the product reconciliation gets thrown into a queue before the next backed reconciler run. 
Looks like the queueing happens in order.
But saying this, I'm happy to implement time based requeue

controllers/capabilities/backend_controller.go
MStokluska: Hey, so it looks to me that the application plan limits and pricing mentions of the backend do not block the backend from deletion and get cleaned up by 3scale. Would you still have it deleted manually anyway?

controllers/capabilities/backend_controller.go
```diff
@@ -174,3 +202,71 @@ func (r *BackendReconciler) validateSpec(backendResource *capabilitiesv1beta1.Ba
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (r *BackendReconciler) fetchTenantProductCRs(productsCRsList *capabilitiesv1beta1.ProductList, backend *capabilitiesv1beta1.Backend) ([]capabilitiesv1beta1.Product, error) {
+	var productsList []capabilitiesv1beta1.Product
+	backendProviderAccount, err := controllerhelper.LookupProviderAccount(r.Client(), backend.Namespace, backend.Spec.ProviderAccountRef, r.Logger())
```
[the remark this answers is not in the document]
MStokluska: I was hoping to achieve this on my first try, but I also need spec.SystemName

controllers/capabilities/backend_controller.go
eguzki: My proposal is to trying to avoid error in the 3scale API call. Increase probability that the 3scale API will work,

controllers/capabilities/backend_controller.go
eguzki: Yes, but the product reconciler will put the Product CR into inconsistent state because it has references to not existing backend resources

controllers/capabilities/backend_controller.go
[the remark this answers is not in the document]
eguzki: then, add the systemName too

controllers/capabilities/backend_controller.go
MStokluska: Cool, thanks Eguzki, I've set it to 2sec.

controllers/capabilities/backend_controller.go
```diff
@@ -74,12 +81,46 @@ func (r *BackendReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+		if backend.Status.ID != nil {
+			requeue, err := r.removeBackend(backend.Spec.ProviderAccountRef, *backend.Status.ID, backend.Namespace, backend.Spec.SystemName)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+			if requeue {
+				return ctrl.Result{RequeueAfter: requeueTime}, nil
+			}
+		} else {
+			return ctrl.Result{}, fmt.Errorf("backend %s .status.ID is missing, cannot remove backend", backend.Spec.Name)
```
eguzki, line 97: This will start some endless loop. 

What about logging internally that the backend could not be removed because ID was missing?

controllers/capabilities/backend_controller.go
```diff
@@ -74,12 +81,46 @@ func (r *BackendReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		reqLogger.V(1).Info(string(jsonData))
 	}
 
+	// Ignore deleted Backends, this can happen when foregroundDeletion is enabled
+	// https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#foreground-cascading-deletion
+	if backend.GetDeletionTimestamp() != nil && controllerutil.ContainsFinalizer(backend, backendFinalizer) {
+		// Attempt to remove backend only if backend.Status.ID is present
+		if backend.Status.ID != nil {
+			requeue, err := r.removeBackend(backend.Spec.ProviderAccountRef, *backend.Status.ID, backend.Namespace, backend.Spec.SystemName)
```
eguzki, line 89: The workflow should be:

* Remove product CR's backend references. Update product CR's if backend references were updated
* If at least one product CR was updated, requeue 
* if no product CR was updated, call 3scale api to delete the backend

controllers/capabilities/product_controller.go
```diff
@@ -76,7 +79,37 @@ func (r *ProductReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+		}
+
+		if product.Status.ID != nil {
+			err = r.removeProduct(product)
+			if err != nil {
+				r.EventRecorder().Eventf(product, corev1.EventTypeWarning, "Failed to delete product", "%v", err)
+				return ctrl.Result{}, err
+			}
+		} else {
+			return ctrl.Result{}, fmt.Errorf("product %s .status.ID is missing, cannot remove product", product.Spec.Name)
```
eguzki, line 96: same comment as for the backend controller

controllers/capabilities/product_controller.go
```diff
@@ -76,7 +79,37 @@ func (r *ProductReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 	// Ignore deleted Products, this can happen when foregroundDeletion is enabled
 	// https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#foreground-cascading-deletion
-	if product.DeletionTimestamp != nil {
+	if product.GetDeletionTimestamp() != nil && controllerutil.ContainsFinalizer(product, productFinalizer) {
+		controllerutil.RemoveFinalizer(product, productFinalizer)
```
eguzki, line 83: To be consistent with the backend controller, the finalizer should be removed after removing the product in 3scale

controllers/capabilities/product_controller.go
```diff
@@ -329,3 +362,30 @@ func computeBackendUsageList(list []capabilitiesv1beta1.Backend, backendUsageMap
...
+		return err
+	}
+
+	err = threescaleAPIClient.DeleteProduct(*productResource.Status.ID)
+	if err != nil {
+		return err
+	}
+
+	//Confirm that product has been removed
+	products, err := threescaleAPIClient.ListProducts()
```
eguzki, line 383: no need to check. If the 3scale API does not return an error, the controller can assume it has been deleted.

eguzki (changes requested):

controllers/capabilities/backend_controller.go
MStokluska: Good catch, thanks, logging internally seems like a way to go.

controllers/capabilities/backend_controller.go
MStokluska: Hey, sorry, I seem to be missing the purpose of this comment. The current flow is:
- retrieve product CRs that belong to the tenant
- remove any mentions of backend in product CR, one by one where if updated - we are doing requeue.
- if there is no updates to product CR - we are removing backend 

What am I missing?

controllers/capabilities/product_controller.go
MStokluska: Thanks

controllers/capabilities/product_controller.go
MStokluska: true, thanks Eguzki

controllers/capabilities/backend_controller.go
eguzki: the current implementation, returns with requeue order every time a product CR is updated. If there are N products being updated, N requeues.

What I suggest is to update all product CRs (when backend references were removed) in a single run. After that, requeue if at least one product CR was updated. If no product CR was updated, then remove backend. If there are N products being updated, 1 requeue. Besides, IMO, code is much simpler to understand (no early return in https://github.com/3scale/3scale-operator/pull/723/files#diff-e35c4c9b422525ea6dec6faafdc3acd7f0e376fd2c8a180c233eb7d0cbf19041R256)

MStokluska: I get it, good point. Thanks, done!

controllers/capabilities/product_controller.go
MStokluska: done, thanks

eguzki (approved):

eguzki:
good job :medal_military:

PR #724, opened 2022-03-08
tenant controller: secrets default namespace is the tenant's one

eguzki:
### what

The Tenant CRD has three fields of SecretReference type. If the `Namespace` is not set in the custom resource, the tenant controller should default that value to the custom resource's namespace.

### verification steps

Deploy 3scale with APIManager CR

```yaml
kind: APIManager
metadata:
  name: apimanager1
spec:
  wildcardDomain: example.com
  system:
    fileStorage:
      simpleStorageService:
        configurationSecretRef:
          name: aws-auth
```

Then create one tenant with secret references without namespace

```yaml
kind: Tenant
metadata:
  name: ecorp-tenant
spec:
  username: admin
  email: admin@example.com
  organizationName: ECorp
  masterCredentialsRef:
    name: system-seed
  passwordCredentialsRef:
    name: ecorp-admin-secret
  tenantSecretRef:
    name: ecorp-tenant-secret
``` 

Without the fix, the controller would fail and tenant CR status would not have IDs from 3scale

If it works, the tenant status should look like this:

```yaml
k get tenants  ecorp-tenant -o jsonpath='{.status}' | yq e -P
adminId: 5
tenantId: 4
```

### DIFF
--- a/apis/capabilities/v1alpha1/tenant_types.go
+++ b/apis/capabilities/v1alpha1/tenant_types.go
@@ -22,6 +22,7 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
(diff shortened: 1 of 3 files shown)


MStokluska:
code changes look good - going to verify it on a cluster

MStokluska:
code changes look good - going to verify it on a cluster

MStokluska:
Verified on cluster, looks good to me
/lgtm

PR #726, opened 2022-03-08
tenant controller: on cr deletion first remove finalizers

eguzki:
### what 		
On tenant deletion, first remove the finalizer. Then try to delete the tenant.
If the tenant cannot be deleted for any reason, the controller cannot know whether it is transient error or not.

If finalizer was removed after actual tenant deletion in 3scale, on not transient error, the controller would keep trying (with exponential backoff retry intervals) and the CR would not be deleted until the error is fixed externally.

The "con" of this implementation is that the operator has only one chance to delete the tenant from 3scale.

### Verification steps

Deploy 3scale with APIManager CR

```yaml
k apply -f - <<EOF
kind: APIManager
metadata:
  name: apimanager1
spec:
  wildcardDomain: example.com
  system:
    fileStorage:
      simpleStorageService:
        configurationSecretRef:
          name: aws-auth
EOF
```
Create a new tenant:

```yaml
k apply -f - <<EOF
kind: Tenant
metadata:
  name: ecorp-tenant
spec:
  username: admin
  email: admin@example.com
  organizationName: ECorp
  masterCredentialsRef:
    name: system-seed
  passwordCredentialsRef:
    name: ecorp-admin-secret
  tenantSecretRef:
    name: ecorp-tenant-secret
EOF
```

Remove 3scale

```
k delete apimanager apimanager1
```

Then, delete the tenant

```
k delete tenant ecorp-tenant
```

The tenant object should be deleted

### DIFF
--- a/controllers/capabilities/tenant_controller.go
+++ b/controllers/capabilities/tenant_controller.go
@@ -107,6 +107,19 @@ func (r *TenantReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 	// Tenant has been marked for deletion
 	if tenantR.GetDeletionTimestamp() != nil && controllerutil.ContainsFinalizer(tenantR, tenantFinalizer) {
+		// First remove the finalizer
+		// Then try to delete the tenant.
+		// If the tenant cannot be deleted for any reason,
+		// the controller cannot know whether it is transient error or not.
+		// If finalizer was removed after actual tenant deletion in 3scale, on not transient error
+		// the controller would keep trying (with exponential backoff retry intervals)
+		// and the CR would not be deleted until the error is fixed externally.
+		controllerutil.RemoveFinalizer(tenantR, tenantFinalizer)
+		err = r.UpdateResource(tenantR)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
 		existingTenant, err := controllerhelper.FetchTenant(tenantR.Status.TenantId, portaClient)
 		if err != nil {
 			return ctrl.Result{}, err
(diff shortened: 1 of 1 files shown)


controllers/capabilities/tenant_controller.go
```diff
@@ -117,12 +129,6 @@ func (r *TenantReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 			}
 		}
 
-		// add or remove finalizer, tenant CR will be deleted if tenant in 3scale does not exists, if tenant is deleted or if it's already marked for deletion
-		err = controllerhelper.ReconcileFinalizers(tenantR, r.Client, tenantFinalizer)
```
MStokluska, line 121: Hey Eguzki. 
I understand how this fixes the CR being stuck on the cluster, maybe I am missing something but would this not mean that:
- if a tenant CR is deleted the CR will get deleted regardless of the tenant being removed from 3scale or not?
- if so do we really want this behavior? Wouldn't it be too misleading for users?

eguzki: The current implementation will stuck the k8s resource in "deleting" state and you cannot tell if that will eventually change or it is permanent and the CR will exist for ever even if the user tried to delete it. 

A k8s API call may fail, they fail, but eventually will success. With a 3scale API call, we cannot tell. You may try to fetch a tenant that does not exist internally and you will get error for ever. So, I consider safer, less problematic, even though not ideal, I agree, make sure the CR is deleted and have only one change to delete from 3scale.

eguzki:
There are two scenarios.

* The operator tries to delete 3scale tenant first and then removes the finalizers.
  * If the delettion of 3scale tenant fails, of the tenant does not exist in 3scale, the finalizer will never be deleted and the deletion of the CR will be stuck.  
* The oeprator removes the finalizers first and then tries to remove the 3scale tenant
  * The CR will always be deleted, but there is only one chance to remove the tenant in 3scale.  If something goes wrong, there will be inconsistencies (orphan tenants in 3scale)

We have decided to favor CR delettion in stuck if something goes wrong instead of leaving "orphan" tenants in 3scale.

PR #727, opened 2022-03-09
add owners reference to backend and product CRs

MStokluska:


pkg/controller/helper/shared.go
```diff
@@ -0,0 +1,88 @@
...
+RetrieveTenantCR retrieves tenantCR of a tenant that matches the provider account org name
+- providerAccount
+- k8client
+- logger
+If the tenantList is empty it will return nil, nil
+If tenantCR for given providerAccount org is not present, it will return nil, nil
+*/
+func RetrieveTenantCR(providerAccount *ProviderAccount, client k8sclient.Client, logger logr.Logger) (*capabilitiesv1alpha1.Tenant, error) {
+	tenantList := &capabilitiesv1alpha1.TenantList{}
+	err := client.List(context.TODO(), tenantList)
```
eguzki, line 25: Namespace scope? Owner references can only be set in the same namespace.

eguzki (changes requested):

pkg/controller/helper/shared.go
MStokluska: But tenant CR can essentially be created under any namespace, this means to me that if a tenant is created via CR, the secret coming from the tenant creation can have a different ns that the tenant CR.
The product and backend must be under same ns as the secret but we still need to loop through all the tenants CRs.

pkg/controller/helper/shared.go
```diff
@@ -0,0 +1,88 @@
...
+
+	return nil
+}
+
+/*
+retrieveTenantSecret retrieves tenants secret
+- k8client
+- tenantCR
+*/
+func retrieveTenantSecret(client k8sclient.Client, tenantCR *capabilitiesv1alpha1.Tenant) (corev1.Secret, error) {
```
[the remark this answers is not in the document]
MStokluska: thank you

controllers/capabilities/backend_controller.go
```diff
@@ -80,6 +80,27 @@ func (r *BackendReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{}, nil
 	}
 
+	if backend.Spec.ProviderAccountRef != nil {
```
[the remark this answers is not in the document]
MStokluska: thanks Eguzki!

pkg/controller/helper/shared.go
MStokluska: Hey, @eguzki as we have agreed on our recent call - going to get the tenant's CRs that are under the same ns as the product or backend. Although this is bringing a bit of limitation on the cascading deletion it looks like the best way to go with it.

Thank you.

MStokluska: Hey, @eguzki as we have agreed on our recent call - going to get the tenant's CRs that are under the same ns as the product or backend. Although this is bringing a bit of limitation on the cascading deletion it looks like the best way to go with it.

Thank you.

pkg/controller/helper/shared.go
```diff
@@ -0,0 +1,93 @@
...
+	ownerReference := []metav1.OwnerReference{
+		{
+			APIVersion: tenantCR.APIVersion,
+			Kind:       tenantCR.Kind,
+			Name:       tenantCR.Name,
+			UID:        tenantCR.UID,
+		},
+	}
+
+	object.SetOwnerReferences(ownerReference)
```
eguzki, line 70: When the object has already the owner reference to the parent tenant, how does this method behave?

In the controller reconciliation pattern, you always need to check if the action to be taken has already been done.

MStokluska: Thanks, from the notes on this function:
```
SetOwnerReference is a helper method to make sure the given object contains an object reference to the object provided. This allows you to declare that owner has a dependency on the object without specifying it as a controller. If a reference to the same object already exists, it'll be overwritten with the newly provided version. 
```
So it will be overwritten. 
Would you rather see this check if present and only run when not present?

MStokluska: IMO, I would leave it as it is not harming us in any way.

controllers/capabilities/product_controller.go
```diff
@@ -112,6 +112,25 @@ func (r *ProductReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+
+	// Retrieve ownersReference of tenant CR that owns the Backend CR
+	tenantCR, err := controllerhelper.RetrieveTenantCR(providerAccount, r.Client(), r.Logger(), product.Namespace)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// If tenant CR is found, set it's ownersReference as ownerReference in the BackendCR CR
+	if tenantCR != nil {
+		err := controllerhelper.SetOwnersReference(product, r.Client(), tenantCR)
```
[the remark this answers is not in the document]
MStokluska: Thanks Eguzki

pkg/controller/helper/shared.go
MStokluska: Thanks, going to work on an update asap

eguzki (approved):

MStokluska:
@eguzki done, thanks!

PR #728, opened 2022-03-18
fix tenant reconciliation

eguzki:
Jira Issue: https://issues.redhat.com/browse/THREESCALE-8301

### what

When a tenant is created declaratively, using the TenantCR, and afterwards it’s deleted from the 3scale admin portal UI, the operator starts creating tenants in an infinite loop.

### how

The issue was caused by a wrong reconciliation logic that did not take into account the NOT FOUND response when reading an admin user. That returned an error, which turned up into a new reconciliation loop before the tenantID and AdminID was stored in the status section. Every new reconciliation loop ended up creatiing a new tenant.

The reconciliation loop logic has been refactored. Basically:
* Read tenant from status.TenantId
* If not found, create it and store in status new tenant ID together with 0 in the status.AdminId
* Read tenant user list looking for one user with Email and Username as specified in the tenant's spec.
* If not found any, create one
* reconcile the admin user -> make admin role and activated state in 3scale

### Verification steps

Deploy 3scale with the APIManager

```yaml
k apply -f - <<EOF
kind: APIManager
metadata:
  name: apimanager
spec:
  wildcardDomain: example.com
EOF
```

Wait for the deployment to be ready

```
oc wait --for=condition=available apimanager/apimanager --timeout=-1s
```

Deploy new tenant CR

```yaml
k apply -f - <<EOF
kind: Tenant
metadata:
  name: ecorp-tenant
spec:
  username: admin
  email: admin@example.com
  organizationName: ECorp
  masterCredentialsRef:
    name: system-seed
  passwordCredentialsRef:
    name: ecorp-admin-secret
  tenantSecretRef:
    name: ecorp-tenant-secret
EOF
```

From the 3scale admin portal delete the tenant created (it will not be deleted, just scheduled for deletion)

Fully delete the tenant with the rails console:

open console
```
oc rsh -c system-master "$(oc get pods --selector deploymentconfig=system-app -o name)"
bundle exec rails console
```
delete with lines (takes few seconds)

```
tenant = Account.find(PROVIDER_ID)
# run several times
DeleteAccountHierarchyWorker.perform_later(tenant)
```

the tenant controller should not have notified the deletion of the tenant. We need to force tenant reconciliation for our tenant CR. Either stop and start the operator or update something in the TenantCR. I usually add harmless annotation

```
k annotate tenant ecorp-tenant newannotation=yes
```

The tenant controller should create (only) one new tenant with the CR spec.

### DIFF
--- a/controllers/capabilities/tenant_internal_reconciler.go
+++ b/controllers/capabilities/tenant_internal_reconciler.go
@@ -3,10 +3,13 @@ package controllers
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
+	"reflect"
 
 	porta_client_pkg "github.com/3scale/3scale-porta-go-client/client"
 	"github.com/go-logr/logr"
+	"github.com/google/go-cmp/cmp"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -50,9 +53,13 @@ func (r *TenantInternalReconciler) Run() (ctrl.Result, error) {
 		return res, nil
 	}
 
-	err = r.reconcileAdminUser()
+	res, err = r.reconcileAdminUser()
 	if err != nil {
 		return ctrl.Result{}, err
+	}
+
+	if res.Requeue {
+		return res, nil
 	}
 
 	return ctrl.Result{}, err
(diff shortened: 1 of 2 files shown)


MStokluska:
:eyes:

MStokluska:
:eyes:

MStokluska:
Verified on the cluster, all looks good. Only one tenant is created afterward.

MStokluska (approved):
looks good, verified on cluster and the code looks good to me.

PR #731, opened 2022-03-31
Fix product delete

eguzki:
### What 

The product deletion was broken. When deleting a product CR, the operator started retrying repeatedly without removing the finalizers.

### How

Two issues were involved.
* The  `github.com/3scale/3scale-porta-go-client/client` v0.5.0 had a bug and returned error on product deletion command even if the 3scale API returned `200 OK`.
* After first error, the operator tried to delete some non existing product, so the 3scale API returned `404 Not Found`. This error was not handled correctly by the product controller resulting in repeated retries. 

Both issues have been fixed

### Verification Steps

Run the operator

```
make run
```
Create a product

```yaml
k apply -f - <<EOF
apiVersion: v1
kind: Secret
metadata:
  name: threescale-provider-account
type: Opaque
stringData:
  token: MY_TOKEN
  adminURL: https://3scale-admin.example.com
EOF
```

```yaml
k apply -f - <<EOF
kind: Product
metadata:
  name: product1
spec:
  name: "OperatedProduct 1"
EOF
```

Delete the product should return "quickly", meaning that the finalizer has been correctly deleted

```
$ k delete product product1
product.capabilities.3scale.net "product1" deleted
```

Checking the 3scale tenant, there should not be any product called `OperatedProduct 1`

### DIFF
--- a/controllers/capabilities/backend_controller.go
+++ b/controllers/capabilities/backend_controller.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"time"
 
+	threescaleapi "github.com/3scale/3scale-porta-go-client/client"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/validation/field"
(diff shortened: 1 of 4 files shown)


Patryk-Stefanski:
Created secret in 3scale namespace with auth token and used as account ref in CR for creation of the product. Deleting the product CR also deleted the product in 3scale.
/lgtm

PR #733, opened 2022-04-07
THREESCALE-6836: Combining external/internal dependencies

sergioifg94:
Deprecate the `.Spec.HighAvailability` field in favour of a new `.Spec.ExternalComponents` fields, which allows to specify a combination of external/internal dependencies for components.

Refactor the logic that manages the reconciliation of internal and external dependencies to allow a fine grained reconciliation.


pkg/3scale/amp/operator/upgrade.go
```diff
@@ -191,21 +192,25 @@ func (u *UpgradeApiManager) upgradeImages() (reconcile.Result, error) {
...
+		when(
 
-		res, err = u.upgradeSystemRedisImageStream()
-		if res.Requeue || err != nil {
-			return res, err
-		}
+			notExternal(v1alpha1.SystemDatabase),
+			(*UpgradeApiManager).upgradeSystemDatabaseImageStream,
+		),
+	)
```
sergioifg94, line 209: Hopefully this is not too overcomplicated. The goal is to avoid the alternative chain of `if` blocks, while also making the code more readable (`sequentially, when Backend redis is not external, upgrade backend redis ImageStream, ...`)

pkg/3scale/amp/operator/highavailability_reconciler.go
```diff
@@ -54,3 +55,48 @@ func HighAvailability(apimanager *appsv1alpha1.APIManager, client client.Client)
 	}
 	return component.NewHighAvailability(opts), nil
 }
+
+func NewBackendExternalRedisReconciler(baseAPIManagerLogicReconciler *BaseAPIManagerLogicReconciler) DependencyReconciler {
```
eguzki, line 59: With this reconcilers, `HighAvailabilityReconciler` is not being used, right? can it be removed?

pkg/3scale/amp/operator/redis_reconciler.go
```diff
@@ -19,6 +22,95 @@ func NewRedisReconciler(baseAPIManagerLogicReconciler *BaseAPIManagerLogicReconc
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	dcMutator := reconcilers.DeploymentConfigMutator(
+		reconcilers.DeploymentConfigContainerResourcesMutator,
+		reconcilers.DeploymentConfigAffinityMutator,
+		reconcilers.DeploymentConfigTolerationsMutator,
+	)
+	err = r.ReconcileDeploymentConfig(r.DeploymentConfig(redis), dcMutator)
```
eguzki, line 77: err not checked

pkg/3scale/amp/operator/redis_reconciler.go
```diff
@@ -19,6 +22,95 @@ func NewRedisReconciler(baseAPIManagerLogicReconciler *BaseAPIManagerLogicReconc
 	}
 }
 
+// RedisDependencyReconciler is a generic DependencyReconciler that reconciles
+// an internal Redis instance using the Redis options
+type RedisDependencyReconciler struct {
```
eguzki, line 27: The existing `RedisReconciler` is not being used. Maybe remove that unused one and use that name for this struct?

pkg/3scale/amp/operator/upgrade.go
```diff
@@ -191,21 +192,25 @@ func (u *UpgradeApiManager) upgradeImages() (reconcile.Result, error) {
 		return res, err
 	}
 
-	if !u.apiManager.IsExternalDatabaseEnabled() {
-		res, err = u.upgradeBackendRedisImageStream()
-		if res.Requeue || err != nil {
-			return res, err
-		}
+	upgradeImages := sequentially(
+		when(
```
eguzki, line 196: This is actually not needed because in the upgrade procedure, customers could not specify external dependencies. It was all or nothing (expect for the zync database). But it should also work

eguzki (changes requested):
Overall looks good

Let's go with this approach

sergioifg94:
Thanks for the review @eguzki I just pushed a few commits addressing the comments

eguzki (approved):

PR #740, opened 2022-04-26
THREESCALE-7961: Use sum_irate instead of sum_rate

sergioifg94:
## Description

> Link to JIRA: https://issues.redhat.com/browse/THREESCALE-7961

Add logic to use `sum_irate` in monitoring queries when the Openshift version is greater or equal than 4.9. Keep using `sum_rate` when the Openshift version is less than 4.9.

## Verification steps

1. **On an Openshift 4.9 (or above) cluster**
1. Install 3scale operator from this PR and the monitoring stack
2. Create an APIManager with monitoring enabled
3. Verify that the Grafana dashboards that use `sum_rate` use now `sum_irate` and show data
4. **On an Openshift 4.8 (or below) cluster**
5. Repeat steps 2 and 3
6. Verify that the Grafana dashboards continue using `sum_rate` and show data


pkg/helper/openshift.go
```diff
@@ -0,0 +1,126 @@
...
+		return nil, err
+	}
+
+	return AddSumRateField(sumRate), nil
+}
+
+// AddSumRateField creates a templateDataMutation that constructs a struct
+// identical to the original with the added `SumRate` field, containing the
+// value of sumRate
+func AddSumRateField(sumRate string) TemplateDataMutation {
```
eguzki, line 73: I understand the motivation of this implementation based on the `reflect` package. 

I am ok with it, but I would like to share that IMHO the complexity added to solve a relatively simple use case does not pay off. 

Some (if not all) the dashboard templates have a new parameter called `SumRate`. I would add them to the data struct and the value passed to the `SumRate` will be computed based on the logic you implemented. 

For example, in `pkg/3scale/amp/component/generic_monitoring.go`
```go

func KubernetesResourcesByNamespaceGrafanaDashboard(sumRate string, ns string, appLabel string) *grafanav1alpha1.GrafanaDashboard {
	data := &struct {
		Namespace string
                SumRate string
	}{
		ns, sumRate,
	}
	return &grafanav1alpha1.GrafanaDashboard{
		ObjectMeta: metav1.ObjectMeta{
			Name: "kubernetes-resources-by-namespace",
			Labels: map[string]string{
				"monitoring-key": common.MonitoringKey,
				"app":            appLabel,
			},
		},
		Spec: grafanav1alpha1.GrafanaDashboardSpec{
			Json: assets.TemplateAsset("monitoring/kubernetes-resources-by-namespace-grafana-dashboard-1.json.tpl", data),
			Name: fmt.Sprintf("%s/kubernetes-resources-by-namespace-grafana-dashboard-1.json", ns),
		},
	}
}
```

The caller would be:
```go
sumRate, err := helper.SumRateForOpenshiftVersion(r.Context(), r.Client())  
if err != nil {                                                             
}                  

grafanaDashboard := component.KubernetesResourcesByNamespaceGrafanaDashboard(sumRate, r.apiManager.Namespace, *r.apiManager.Spec.AppLabel)
err = r.ReconcileGrafanaDashboard(grafanaDashboard, reconcilers.GenericGrafanaDashboardsMutator)   
if err != nil {                                                                                                  
}                                                                                                                
                                                

```

sergioifg94: Thanks @eguzki I followed your suggestion and simplify the inclusion of the `SumRate` parameter into the templates

pkg/helper/openshift.go
```diff
@@ -0,0 +1,60 @@
...
+
+	// If the version is less than 4.9, use sum_rate
+	if comparison < 0 {
+		sumRate = "sum_rate"
+	}
+
+	return sumRate, nil
+}
+
+type TemplateDataMutation func(interface{}) interface{}
```
eguzki, line 60: is it being used?

eguzki (approved):

pkg/helper/openshift.go
sergioifg94: Good catch, I missed it when removing the stuff from the initial implementation. Fixed it now

PR #744, opened 2022-05-04
change sum_rate to sum_irate in dashboards and prometheusrules

austincunningham:
# What 
After deploying the monitoring stack I notices a number of panels with no data 
This may be due to me using the latest version of grafana and prometheus in Operatorhub
- grafana 7.5.15
- prometheus 0.47.0

found a lot of the queries that failed were using
`node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate`

found they panels work again if switched to 
`node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate`

And found an issue regarding this change regarding a breaking change from rate to irate
https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/670

Guessing this can't merge because of the older versions of prometheus and grafana don't support it.


austincunningham:
Looks like this has already updated. must be working off an older branch.

eguzki:
Fixed in https://github.com/3scale/3scale-operator/pull/744

PR #745, opened 2022-05-05
grant clusterversions.config.openshift.io resource add and list verbs

eguzki:
### what

The 3scale operator role assigned to the service account running the 3scale operator pod missed permission to `get`, `list` `clusterversions.config.openshift.io`. It was complaining with the following error:

```
E0505 12:54:28.484088       1 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.6/tools/cache/reflector.go:125: Failed to list *v1.ClusterVersion: clusterversions.config.openshift.io is forbidden: User "system:serviceaccount:eguzki:3scale-operator" cannot list resource "clusterversions" in API group "config.openshift.io" at the cluster scope
```

### Verification

Deploy the 3scale operator from OLM: follow our [user guide to deploy using OLM](https://github.com/3scale/3scale-operator/blob/master/doc/development.md#deploy-custom-3scale-operator-using-olm)

Create APIManager object, check all 3scale deployment objects are created. When the operator does not have permission to do something, it stops creating new deployment objects and retries.

### DIFF
--- a/controllers/apps/apimanager_controller.go
+++ b/controllers/apps/apimanager_controller.go
@@ -64,7 +64,7 @@ var _ reconcile.Reconciler = &APIManagerReconciler{}
 // +kubebuilder:rbac:groups=policy,namespace=placeholder,resources=poddisruptionbudgets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=monitoring.coreos.com,namespace=placeholder,resources=podmonitors;servicemonitors;prometheusrules,verbs=get;list;watch;create;update;delete
 // +kubebuilder:rbac:groups=integreatly.org,namespace=placeholder,resources=grafanadashboards,verbs=get;list;watch;create;update;delete
-// +kubebuilder:rbac:groups=config.openshift.io,namespace=placeholder,resources=clusterversions,verbs=get;list
+// +kubebuilder:rbac:groups=config.openshift.io,resources=clusterversions,verbs=get;list
 
 func (r *APIManagerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	_ = context.Background()
--- a/controllers/apps/apimanager_controller.go
+++ b/controllers/apps/apimanager_controller.go
@@ -64,6 +64,7 @@ var _ reconcile.Reconciler = &APIManagerReconciler{}
 // +kubebuilder:rbac:groups=policy,namespace=placeholder,resources=poddisruptionbudgets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=monitoring.coreos.com,namespace=placeholder,resources=podmonitors;servicemonitors;prometheusrules,verbs=get;list;watch;create;update;delete
 // +kubebuilder:rbac:groups=integreatly.org,namespace=placeholder,resources=grafanadashboards,verbs=get;list;watch;create;update;delete
+// +kubebuilder:rbac:groups=config.openshift.io,namespace=placeholder,resources=clusterversions,verbs=get;list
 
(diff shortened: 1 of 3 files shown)


controllers/apps/apimanager_controller.go
```diff
@@ -64,6 +64,7 @@ var _ reconcile.Reconciler = &APIManagerReconciler{}
 // +kubebuilder:rbac:groups=policy,namespace=placeholder,resources=poddisruptionbudgets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=monitoring.coreos.com,namespace=placeholder,resources=podmonitors;servicemonitors;prometheusrules,verbs=get;list;watch;create;update;delete
 // +kubebuilder:rbac:groups=integreatly.org,namespace=placeholder,resources=grafanadashboards,verbs=get;list;watch;create;update;delete
+// +kubebuilder:rbac:groups=config.openshift.io,namespace=placeholder,resources=clusterversions,verbs=get;list
```
sergioifg94, line 67: I haven't been able to verify the fix. Followed the verification steps and the error continues showing up. Perhaps the permission should be added in the ClusterRole? 

```suggestion
// +kubebuilder:rbac:groups=config.openshift.io,resources=clusterversions,verbs=get;list
```

eguzki: If I am not wrong, the `namespace=placeholder` is used for the namespaced deployment of the operator. Otherwise, what is it's purpose?

If  that would be deleted, why not from the other rbac deficinitions? I think we should keep it. 

If you deployed the operator in its own deployment and pod and still get the errors, we should do pair reviewing to see what you did differently

sergioifg94: The reason to remove the `namespace` part of the marker is that when the namespace is not specified, Kubebuilder will set this permission in the `ClusterRole` (instead of the `Role`)

I deployed the operator with OLM, generated the manifests from this PR and even with the permission set in the role, the error was still showing up. I manually added the permission to the ClusterRole and it seemed to work shortly after

eguzki: Oh, I see. The clusterversion is not namespaced resource, you need cluster role. 

When I run the verification steps, I might have deployed the operator cluster wide, which effectively turns permissions in clusterroles and the permission issue was gone. 

yes, indeed... we need to remove that, good catch.

eguzki:
@sergioifg94 ready for a new review

controllers/apps/apimanager_controller.go
```diff
@@ -64,6 +64,7 @@ var _ reconcile.Reconciler = &APIManagerReconciler{}
 // +kubebuilder:rbac:groups=policy,namespace=placeholder,resources=poddisruptionbudgets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=monitoring.coreos.com,namespace=placeholder,resources=podmonitors;servicemonitors;prometheusrules,verbs=get;list;watch;create;update;delete
 // +kubebuilder:rbac:groups=integreatly.org,namespace=placeholder,resources=grafanadashboards,verbs=get;list;watch;create;update;delete
+// +kubebuilder:rbac:groups=config.openshift.io,resources=clusterversions,verbs=get;list
```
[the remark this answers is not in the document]
eguzki: This is unexpected. The controller only runs `get` operation. Why is `watch` needed?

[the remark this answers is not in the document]
sergioifg94: I might be wrong but I think the client populates the cache in the background by watching changes in the objects

sergioifg94 (approved):
Thanks @eguzki looks good now, and verified it on cluster. Happy to have this merged

PR #746, opened 2022-05-06
fix cascading deletion

MStokluska:
Fix cascading deletion issue when provider account secret is not found

Verification:
Create tenant
Create backend associated with that tenant
Create product associated with that tenant
Remove tenant
Confirm product and backend CRs are deleted as well

### DIFF
--- a/controllers/capabilities/backend_controller.go
+++ b/controllers/capabilities/backend_controller.go
@@ -20,9 +20,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


eguzki:
I was looking going to implement something very similar. Then I realized that in the backend controller, if the secret is not found, then the product CR references are not deleted. 

I refactored a bit to decouple backend deletion from 3scale and product reference cleaning. 

Let me know what you think: https://github.com/3scale/3scale-operator/pull/747

MStokluska:
closing in favor of https://github.com/3scale/3scale-operator/pull/747

PR #747, opened 2022-05-06
Fix tenant cascade deletion

eguzki:
### what

Fix cascading deletion issue when provider account secret is not found

### Verification steps
* Create tenant and admin pass secret
```yaml
k apply -f - <<EOF
---
apiVersion: v1
kind: Secret
metadata:
  name: example-admin-secret
type: Opaque
stringData:
  admin_password: "123456"
---
kind: Tenant
metadata:
  name: example-tenant
spec:
  username: admin
  systemMasterUrl: https://master.3scale.example.com
  email: admin@example.com
  organizationName: example
  masterCredentialsRef:
    name: system-seed
  passwordCredentialsRef:
    name: example-admin-secret
  tenantSecretRef:
    name: example-tenant-secret
EOF
```

* Create backend associated with that tenant
```yaml
k apply -f - <<EOF
---
kind: Backend
metadata:
  name: backend-a
spec:
  name: "Operated Backend A"
  systemName: "backenda"
  privateBaseURL: "https://exampleA.com"
  providerAccountRef:
    name: example-tenant-secret
```

* Create product associated with that tenant
```yaml
k apply -f - <<EOF
---
kind: Product
metadata:
  name: product1
spec:
  name: "OperatedProduct 1"
  backendUsages:
    backenda:
      path: /v1
  providerAccountRef:
    name: example-tenant-secret
```

* Remove tenant

``` 
k delete tenant  example-tenant
```

* Confirm product and backend CRs are deleted as well. If deletion fails, the objects deletion process will be stuck and objects will remain available but marked for deletion

### DIFF
--- a/controllers/capabilities/backend_controller.go
+++ b/controllers/capabilities/backend_controller.go
@@ -25,8 +25,10 @@ import (
 	threescaleapi "github.com/3scale/3scale-porta-go-client/client"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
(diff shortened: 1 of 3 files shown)


guicassolato (approved):
This looks correct to me.

Sometimes the deletion of the API Backend will be accepted in 3scale regardless of existing associations with the API Product – i.e., when the parent Tenant has been marked for destruction as well –, while some other times it won't. This is subject to concurrency and cannot be predicted in cases of cascade deletion.

Ref: https://github.com/3scale/porta/blob/33cc4d5bcb5910295bfa28d84416ff05e1a606f3/app/models/backend_api.rb#L114

controllers/capabilities/backend_controller.go
```diff
@@ -240,71 +238,95 @@ func (r *BackendReconciler) validateSpec(backendResource *capabilitiesv1beta1.Ba
...
+
+	return ctrl.Result{}, nil
+}
+
+func (r *BackendReconciler) removeBackendFrom3scale(backend *capabilitiesv1beta1.Backend) error {
+	logger := r.Logger().WithValues("backend", client.ObjectKey{Name: backend.Name, Namespace: backend.Namespace})
+
+	// Attempt to remove backend only if backend.Status.ID is present
+	if backend.Status.ID == nil {
+		logger.Info("could not remove backend because backend ID is missing for backend name")
```
MStokluska, line 286: if the ID is nil I think 305 will be throwing nil pointer?

controllers/capabilities/backend_controller.go
```diff
@@ -240,71 +238,95 @@ func (r *BackendReconciler) validateSpec(backendResource *capabilitiesv1beta1.Ba
...
+	}
+
+	providerAccount, err := controllerhelper.LookupProviderAccount(r.Client(), backend.Namespace, backend.Spec.ProviderAccountRef, logger)
+	if apierrors.IsNotFound(err) {
```
MStokluska, line 290: would it be cleaner that way:
```
if err != nil {
   if apierrors.IsNotFound(err) {
       		logger.Info("backend not deleted from 3scale, provider account not found")
		return nil
   }
   return err
 }

controllers/capabilities/backend_controller.go
eguzki: good catch! I forgot the early `return nil`

MStokluska:
Code looks good to me,
I've verified on cluster with following scenarios:
1.
Tenant creates 2 backends and 3 products where 2 products are associated with a single backend and 3rd product associated with the second backend
Single backend deletion works as expcted, backend mentions are removed from associated products while 3rd product still points to the correct backend
When tenant gets deleted all CRs are removed correctly.

Good job!

/lgtm

PR #748, opened 2022-05-09
Fix reqlogger panic that caused operator to crash

Patryk-Stefanski:
# What
reqLogger.Info was causing a panic which could cause the operator to crash in some scenarios. Added extra field to prevent it from panicking.

# Verification

1. Create a tenant  through CR
2. Delete tenant through ui
3. Delete tenant CR
4. Operator shouldn't panic and crash

### DIFF
--- a/controllers/capabilities/tenant_controller.go
+++ b/controllers/capabilities/tenant_controller.go
@@ -122,7 +122,7 @@ func (r *TenantReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 					return ctrl.Result{}, err
 				}
 			} else {
-				reqLogger.Info("Removing tenant CR - tenant is already scheduled for deletion %v", existingTenant.Signup.Account.ID)
+				reqLogger.Info("Removing tenant CR - tenant is already scheduled for deletion", "tenantID", existingTenant.Signup.Account.ID)
 			}
 		}
 


Patryk-Stefanski:
@eguzki  would you mind reviewing this small fix?

controllers/capabilities/tenant_controller.go
```diff
@@ -122,7 +122,7 @@ func (r *TenantReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 					return ctrl.Result{}, err
 				}
 			} else {
-				reqLogger.Info("Removing tenant CR - tenant is already scheduled for deletion %v", existingTenant.Signup.Account.ID)
+				reqLogger.Info("Error", "Removing tenant CR - tenant is already scheduled for deletion %v", existingTenant.Signup.Account.ID)
```
eguzki, line 125: Good catch!

In logr, format strings are not allowed [link](https://github.com/go-logr/logr#why-not-allow-format-strings-too)

It should be:
```go
reqLogger.Info("Removing tenant CR - tenant is already scheduled for deletion", "tenantID", existingTenant.Signup.Account.ID)
```

eguzki (changes requested):

controllers/capabilities/tenant_controller.go
Patryk-Stefanski: Ah, okay. Made the change.

eguzki (approved):

PR #749, opened 2022-05-12
update the grafana-datasource secret name

austincunningham:
# What 
grafana-datasources secret change to grafana-datasources-v2 from Openshift 4.10

image

Updating the monitoring doc


eguzki:
Ok to update this. 

FYI, we are considering updating the doc to use the  OCP's user defined workload monitoring. The grafana data source would be the Thanos querier which aggregates platform and user defined metrics. https://issues.redhat.com/browse/THREESCALE-8389

PR #750, opened 2022-05-16
fix product ref doc: application plan published

eguzki:


KevFan (approved):
Looks good to me :+1:

PR #751, opened 2022-05-19
THREESCALE-7055 - feat: delete developer user

KevFan:
## Description
Delete developer user when cr is deleted

## Verification
* Install CRDs and run 3scale operator
```
make install
oc new-project 3scale
make run
```
* Create S3 secret for APIManager
```
oc apply -f - <<EOF   
---
apiVersion: v1
kind: Secret
metadata:
  creationTimestamp: null
  name: aws-auth
stringData:
  AWS_ACCESS_KEY_ID: something
  AWS_BUCKET: something
  AWS_REGION: us-east-1
type: Opaque
EOF
```
* Create `APIManager`
```
oc apply -f - <<EOF                                                             
---             
kind: APIManager
metadata:
  name: apimanager1
spec:
  wildcardDomain: apps.some-valid-domain
  system:
    fileStorage:
      simpleStorageService:
        configurationSecretRef:
          name: aws-auth
EOF
```
* Create secret for tenant admin
```
oc apply -f - <<EOF                                                             
---             
apiVersion: v1
kind: Secret
metadata:
  name: ecorp-admin-secret
type: Opaque
stringData:
  admin_password: password
EOF
```
* Create `Tenant`
```
oc apply -f - <<EOF                                                             
---             
kind: Tenant
metadata:
  name: tenant1
spec:
  email: test1@test.com
  masterCredentialsRef:
    name: system-seed
  organizationName: test1Org
  passwordCredentialsRef:
    name: ecorp-admin-secret
  systemMasterUrl: 'https://master.apps.some-valid-domain'
  tenantSecretRef:
    name: tenant-secret
  username: test1
EOF
```
* Create `DeveloperAccount` on the created tenant
```
oc apply -f - <<EOF                                                         
---
kind: DeveloperAccount
metadata:
  name: devaccount
  namespace: 3scale
spec:
  orgName: "devaccount"
  monthlyBillingEnabled: false
  monthlyChargingEnabled: false
  providerAccountRef:
    name: "tenant-secret"
EOF
```
* Create secret for `DeveleperUser`
```
oc apply -f - <<EOF                                                             
---             
apiVersion: v1
kind: Secret
metadata:
  name: my-user-password
type: Opaque
stringData:
  password: password
EOF
```
* Create `DeveloperUser` on the created tenant
```
oc apply -f - <<EOF                                                                       
---
kind: DeveloperUser
metadata:
  name: devuser1
  namespace: 3scale
spec:
  username: devuser
  email: devuser@example.com
  passwordCredentialsRef:
    name: my-user-password
  developerAccountRef:
    name: devaccount
  providerAccountRef:
    name: "tenant-secret"
  role: admin
EOF
```
* Create second `DeveloperUser` on the created tenant
```
oc apply -f - <<EOF                                                                       
---
kind: DeveloperUser
metadata:
  name: devuser2
  namespace: 3scale
spec:
  username: devuser2
  email: devuser2@example.com
  passwordCredentialsRef:
    name: my-user-password
  developerAccountRef:
    name: devaccount
  providerAccountRef:
    name: "tenant-secret"
  role: admin
EOF
```
* Sign into the created tenant and verify that `DeveloperUsers` is present
* Delete 1 `DeveloperUser` CR
```
oc delete developeruser devuser1
```
* Refresh the UI and verify developer user is deleted from the UI
* Delete last developer user 
```
oc delete developeruser devuser2
```
* Verify that last admin error is on the CR
```
oc delete developeraccount devaccount  devuser2 -o yaml
```
* Delete `DeveloperAccount`
```
oc delete developeraccount devaccount 
```
* Verify last developer account cr has been removed 
```
oc get developerusers
```
* Create a developer user and developer user again
* Delete tenant 
```
oc delete tenant tenant1
```
* Verify `DeveloperUser` and `DeveloperAccount` CR is removed along with the `Tenant` CR

### DIFF
--- a/controllers/capabilities/developeruser_controller.go
+++ b/controllers/capabilities/developeruser_controller.go
@@ -20,6 +20,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
+	threescaleapi "github.com/3scale/3scale-porta-go-client/client"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	capabilitiesv1beta1 "github.com/3scale/3scale-operator/apis/capabilities/v1beta1"
 	controllerhelper "github.com/3scale/3scale-operator/pkg/controller/helper"
@@ -29,12 +34,13 @@ import (
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
+
+const developerUserFinalizer = "developeruser.capabilities.3scale.net/finalizer"
 
 // DeveloperUserReconciler reconciles a DeveloperUser object
 type DeveloperUserReconciler struct {
@@ -55,7 +61,7 @@ func (r *DeveloperUserReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 	developerUserCR := &capabilitiesv1beta1.DeveloperUser{}
 	err := r.Client().Get(context.TODO(), req.NamespacedName, developerUserCR)
 	if err != nil {
-		if errors.IsNotFound(err) {
+		if apierrors.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			// Return and don't requeue
@@ -74,10 +80,65 @@ func (r *DeveloperUserReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 		reqLogger.V(1).Info(string(jsonData))
 	}
 
+	// DeveloperUser has been marked for deletion
+	if developerUserCR.GetDeletionTimestamp() != nil && controllerutil.ContainsFinalizer(developerUserCR, developerUserFinalizer) {
+		err = r.removeDeveloperUserFrom3scale(developerUserCR)
+		if err != nil {
+			r.EventRecorder().Eventf(developerUserCR, corev1.EventTypeWarning, "Failed to delete developer user", "%v", err)
+
+			// Update status with err
+			statusResult, statusUpdateErr := NewDeveloperUserStatusReconciler(r.BaseReconciler, developerUserCR, nil, "", nil, err).Reconcile()
+			if statusUpdateErr != nil {
+				return ctrl.Result{}, fmt.Errorf("Failed to update developers user status: %w", statusUpdateErr)
+			}
+
+			if statusResult.Requeue {
+				return statusResult, nil
+			}
+
+			return ctrl.Result{}, err
+		}
+
+		controllerutil.RemoveFinalizer(developerUserCR, developerUserFinalizer)
+		err = r.UpdateResource(developerUserCR)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
+		return ctrl.Result{}, nil
+	}
+
 	// Ignore deleted resource, this can happen when foregroundDeletion is enabled
 	// https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#foreground-cascading-deletion
 	if developerUserCR.DeletionTimestamp != nil {
 		return ctrl.Result{}, nil
+	}
+
+	if !controllerutil.ContainsFinalizer(developerUserCR, developerUserFinalizer) {
+		controllerutil.AddFinalizer(developerUserCR, developerUserFinalizer)
+		err = r.UpdateResource(developerUserCR)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	// Retrieve ownersReference of DeveloperAccount CR that owns the DeveloperUser CR
+	developerAccountCR, err := r.retrieveDevelopAccountCR(developerUserCR)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// If DeveloperAccount CR is found, set it's ownersReference as ownerReference in the DeveloperUser CR
+	if developerAccountCR != nil {
+		updated, err := controllerhelper.SetOwnersReferenceByObjectAndMeta(developerUserCR, r.Client(), developerAccountCR.ObjectMeta, developerAccountCR.TypeMeta)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
+		if updated {
+			return ctrl.Result{Requeue: true}, nil
+		}
 	}
 
 	statusReconciler, reconcileErr := r.reconcileSpec(developerUserCR, reqLogger)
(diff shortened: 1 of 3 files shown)


controllers/capabilities/developeruser_controller.go
```diff
@@ -74,12 +81,61 @@ func (r *DeveloperUserReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
...
+		updated, err := controllerhelper.SetOwnersReference(developerUserCR, r.Client(), tenantCR)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
+		if updated {
+			return ctrl.Result{Requeue: true}, nil
+		}
+	}
+
```
KevFan, line 138: As `DeveloperUsers` are linked to `DeveloperAccounts`, should `DeveloperUsers` be cascade deleted to when the associated `DeveloperAccount` is deleted also ? :thinking:

controllers/capabilities/developeruser_controller.go
```diff
@@ -208,6 +264,37 @@ func (r *DeveloperUserReconciler) findParentAccount(userCR *capabilitiesv1beta1.
...
 }
 
+func (r *DeveloperUserReconciler) removeDeveloperUserFrom3scale(developerUser *capabilitiesv1beta1.DeveloperUser) error {
+	logger := r.Logger().WithValues("developerUser", client.ObjectKey{Name: developerUser.Name, Namespace: developerUser.Namespace})
+
+	// Attempt to remove developerUser only if developerUser.Status.ID is present
+	if developerUser.Status.ID == nil {
+		logger.Info("could not remove developerUser because ID is missing in status")
+		return nil
+	}
```
KevFan, line 274: Should `developerUser.Status.AccountID` be checked also or can we assume `AccountID` is not `nil` when `ID` is not `nil`  ?

controllers/capabilities/developeruser_controller.go
eguzki: I think it makes all the sense. The `DeveloperUsers` have ownerrefs to `DeveloperAccounts`. And `DeveloperAccounts` have ownerrefs to `Tenant`.

The operator should make sure that when deleting a DevUser that if the devAccount does not exist in 3scale, it can be removed.

controllers/capabilities/developeruser_controller.go
eguzki: I think it should be checked as well. If any field is missing, the operator should consider to create a new one.

controllers/capabilities/developeruser_controller.go
```diff
@@ -74,12 +81,61 @@ func (r *DeveloperUserReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
...
+
+	// Retrieve ownersReference of tenant CR that owns the DeveloperUser CR
+	tenantCR, err := controllerhelper.RetrieveTenantCR(providerAccount, r.Client(), r.Logger(), developerUserCR.Namespace)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// If tenant CR is found, set it's ownersReference as ownerReference in the DeveloperUser CR
+	if tenantCR != nil {
+		updated, err := controllerhelper.SetOwnersReference(developerUserCR, r.Client(), tenantCR)
```
eguzki, line 129: As said, maybe it is better to have the developer account CR as owner instead of the Tenant. wdyt?

KevFan: I was thinking of having both `Tenant` and `DeveloperAccount` as owner refs if available :thinking: 

I think there'll be an edge case of orphaned resources on cluster if we only have either or as owner ref.

The main example I'm thinking of is where `DeveloperUser` with an `providerAccountRef` to a `Tenant`, but is waiting for the linked `DeveloperAccount` to be available, and so do not have `ownerRef` to `DeveloperAccount` :thinking: . If we delete the `Tenant` now, this resouce will be left on cluster due to not being marked for deletion.

Probably not a big deal but, I think trying to have both `ownerRefs` if available should minimize orphaned resources of clusters :thinking: 

What do you think ?

eguzki: LGTM.

Both ownerrefs can be `Controller: false`. (Kubernetes does not allow multiple controller ownerreferences)

controllers/capabilities/developeruser_controller.go
```diff
@@ -74,12 +80,67 @@ func (r *DeveloperUserReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
...
+			statusResult, statusUpdateErr := NewDeveloperUserStatusReconciler(r.BaseReconciler, developerUserCR, nil, "", nil, err).Reconcile()
+			if statusUpdateErr != nil {
+				return ctrl.Result{}, fmt.Errorf("Failed to update developers user status: %w", statusUpdateErr)
+			}
+
+			if statusResult.Requeue {
+				return statusResult, nil
+			}
+
+			return ctrl.Result{}, err
```
KevFan, line 99: Screenshot from 2022-05-31 14-59-02

This will add the last admin error to the CR condition field

controllers/capabilities/developeruser_controller.go
eguzki: good to know. Never explored the multi ownerref option. 

Fair enough

doc/developeruser-reference.md
```diff
@@ -132,3 +133,10 @@ Each element of the Condition array has the following fields:
 | Reason | `reason` | string | Condition state reason |
 | Message | `message` | string | Condition state description |
 | LastTransitionTime | `lastTransitionTime` | timestamp | Last transition timestap |
+
+
+## Supported Actions
+* Create - creating the CR will create the developer user in the associated account and tenant
+* Delete - deleting the CR will delete the developer account in the associated account and tenant
```
eguzki, line 140:
```suggestion
* Delete - deleting the CR will delete the developer user in the associated account and tenant
```

eguzki:
looking good!

ready for final review @KevFan ?

pkg/controller/helper/shared.go
```diff
@@ -52,18 +51,29 @@ func RetrieveTenantCR(providerAccount *ProviderAccount, client k8sclient.Client,
...
+}
+
+/*
+SetOwnersReferenceByObjectAndMeta sets ownersReference in given object using ObjectMeta and TypeMeta
+- object
+- k8client
+- objectMeta
+- typeMeta
+*/
+func SetOwnersReferenceByObjectAndMeta(object controllerutil.Object, client k8sclient.Client, objectMeta metav1.ObjectMeta, typeMeta metav1.TypeMeta) (bool, error) {
```
KevFan, line 70: Not sure is this intended from the function but just want to note the current logic for these functions will overwrite any existing ownerReferences on the object.

It will also skip updating if the ownerReference lengths are the same. For example, if object had an ownerRef to `Tenant` but wanted to be updated to `DeveloperAccount` (don't believe there is this case yet), this will be skipped because the old and new ownerRef length of the same :thinking:

We might want to improve the latter at least :thinking: 

https://github.com/3scale/3scale-operator/blob/bd3283cffe3b04c0c57f7592a9d5dbb13c378cfb/pkg/controller/helper/shared.go#L70-L93

KevFan:
@eguzki Yes, ready for final review whenever you find the time

pkg/controller/helper/shared.go
KevFan: Oh cool, didn't know that method existed :+1: 

Eh I think it's fine to continue using the current overwrite method for now as it does not cause any problems if the intention is to always just have the singular ownerReference :thinking: 

Yes, comparing lengths I think is valid if we're using the control runtime library method.

However I think a deeper comparison is needed for this one that is overwriting the ownerRefs, possibly by comparing the owner uid :thinking:

PR #757, opened 2022-06-10
THREESCALE-8467 (2): Make Redis ConfigMap common

sergioifg94:
Currently, the Backend Redis reconciler creates a `redis-config` ConfigMap that is used by the deployments, but the System Redis also depends on that ConfigMap. This results in the failure to deploy the System Redis when the Backend Redis is configured as external (which is now possible since making this configuration fine grained).

Fix this by making the ConfigMap reconciliation common among both Redis reconcilers, which will result in the ConfigMap being created regardless of what combination of internal/external Redis is configured.

Fixes https://issues.redhat.com/browse/THREESCALE-8467

### DIFF
--- a/pkg/3scale/amp/component/redis.go
+++ b/pkg/3scale/amp/component/redis.go
@@ -47,10 +47,11 @@ func (redis *Redis) buildDeploymentConfigObjectMeta() metav1.ObjectMeta {
 }
 
 const (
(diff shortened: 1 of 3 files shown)


eguzki:
Tested with 

```yaml
externalComponents: 
  system: 
    database: true 
    redis: true
```

And working as expected.

PR #742, opened 2022-05-04
Threescale 7135 - Product Promotion

Patryk-Stefanski:
[Threescale-7135](https://issues.redhat.com/secure/RapidBoard.jspa?rapidView=6585&view=detail&selectedIssue=THREESCALE-7135)

# What

Addition of proxyconfigpromote CRD that allows for single-use product promotion to stage or production.

# Verification

1. Deploy 3scale-operator locally 

2.Create dummy smtp secret
```sh
kubectl apply -f - <<EOF
---
kind: Secret
apiVersion: v1
metadata: 
  name: s3-credentials
  namespace: 3scale-test
data: 
  AWS_ACCESS_KEY_ID: UkVQTEFDRV9NRQ==
  AWS_BUCKET: UkVQTEFDRV9NRQ==
  AWS_REGION: UkVQTEFDRV9NRQ==
  AWS_SECRET_ACCESS_KEY: UkVQTEFDRV9NRQ==
type: Opaque
EOF
```

3.Create apimanager resource
```sh
kubectl apply -f - <<EOF
---
apiVersion: apps.3scale.net/v1alpha1
kind: APIManager
metadata: 
  name: apimanager-sample
  namespace: 3scale-test
spec: 
  system: 
    fileStorage: 
      simpleStorageService: 
        configurationSecretRef: 
          name: s3-credentials
  wildcardDomain: <cluster_domain>
EOF
```
cluster domain example : `apps.pstefans.uq32.s1.devshift.org`

4. Create Backend resource
```sh
kubectl apply -f - <<EOF
---
apiVersion: capabilities.3scale.net/v1beta1
kind: Backend
metadata:
  name: backend1-cr
  namespace: 3scale-test
spec:
  mappingRules:
    - httpMethod: GET
      increment: 1
      last: true
      metricMethodRef: hits
      pattern: /
  name: backend1
  privateBaseURL: 'https://api.example.com'
  systemName: backend1
EOF
```

5. Create product resources with the backend referenced
```sh
kubectl apply -f - <<EOF
---
apiVersion: capabilities.3scale.net/v1beta1
kind: Product
metadata:
  name: product1-cr
  namespace: 3scale-test
spec:
  name: product1
  backendUsages:
    backend1:
      path: /
EOF
```

6. Create proxyconfigpromote resources  to promote product to staging.
```sh
kubectl apply -f - <<EOF
---
apiVersion: capabilities.3scale.net/v1beta1
kind: ProxyConfigPromote
metadata:
[...]

PR #765, opened 2022-06-21
fix circleci tests

eguzki:
Unittest memory footpring increased up to 7.5 GB and requires large resource class. 

https://github.com/3scale/3scale-operator/pull/742#issuecomment-1160226259

Opened issue to try to fix the memory footpring: https://issues.redhat.com/browse/THREESCALE-8493

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -290,6 +290,7 @@ jobs:
   unit-tests-coverage:
     docker:
       - image: circleci/golang:1.13.7
+    resource_class: large
     steps:
       - unit-tests
       - run:


MStokluska:
code changes looks good to me, as I understand it it's a temp solution while the https://issues.redhat.com/browse/THREESCALE-8493 is going to be investigated.
/lgtm

eguzki:
> code changes looks good to me, as I understand it it's a temp solution while the https://issues.redhat.com/browse/THREESCALE-8493 is going to be investigated. /lgtm

exactly. If we can reduce the memory footpring, large resource class will not be necessary. I am pretty confident we can. 7.5 Gb does not make sense.

eguzki:
@MStokluska hit "approve" in "files changed"

MStokluska (approved):

PR #758, opened 2022-06-10
Fix backup restore

eguzki:
### what

Fixes: https://issues.redhat.com/browse/THREESCALE-8464

* Restore apimanager restore controller. On operator-sdk upgrade, the migration missed some code https://github.com/3scale/3scale-operator/commit/dd3a3fa9209c1255b02ad7154fdb2ba218a06abf
  * later, not used code (because it was missed in the migration process) was deleted https://github.com/3scale/3scale-operator/commit/b5b1d7cac1d425b9cd64cd819fd739e9194c33b5
 * Add service account, roles and rolebinding to backup and restore jobs

# Verification steps

* Deploy the operator **in cluster wide mode** using OLM from this PR branch [guide](https://github.com/3scale/3scale-operator/blob/master/doc/development.md#deploy-custom-3scale-operator-using-olm)  

* Prepare external databases

<details>

```yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: backend-redis-queues
  labels:
    app: backend-redis-queues
spec:
  selector:
    matchLabels:
      app: backend-redis-queues
  template:
    metadata:
      labels:
        app: backend-redis-queues
    spec:
      containers:
        - name: backend-redis-queues
          args:
            - "/etc/redis.d/redis.conf"
          image: redis
          ports:
            - containerPort: 6379
              name: redis
          volumeMounts:
          - mountPath: /etc/redis.d/
            name: redis-config
      volumes:
        - configMap:
            items:
            - key: redis.conf
              path: redis.conf
            name: redis-config
          name: redis-config
  replicas: 1
---
apiVersion: v1
kind: Service
metadata:
  name: backend-redis-queues
spec:
  selector:
    app: backend-redis-queues
  ports:
    - name: redis
      port: 6379
      protocol: TCP
      targetPort: 6379
---
apiVersion: v1
kind: Secret
metadata:
  name: backend-redis
stringData:
  REDIS_STORAGE_URL: "redis://backend-redis-storage"
  REDIS_STORAGE_SENTINEL_HOSTS: ""
  REDIS_STORAGE_SENTINEL_ROLE: ""
  REDIS_QUEUES_URL: "redis://backend-redis-queues"
  REDIS_QUEUES_SENTINEL_HOSTS: ""
  REDIS_QUEUES_SENTINEL_ROLE: ""
type: Opaque
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: backend-redis-storage
  labels:
    app: backend-redis-storage
spec:
[...]

PR #766, opened 2022-06-22
fix restore resync job

eguzki:
### What
The APIManagerRestore controller, creates a job to resync zync domains. That job needs `system-sidekiq` pod up&running. The controller  creates the mentioned job with default [Pod backoff failure policy](https://issues.redhat.com/browse/THREESCALE-8476#pod-backoff-failure-policy].)

That means that the job will try 6 times before giving up. That last try happens 6 minutes after the job was created. If the system-sidekiq is not up&running in 6 minutes, the restore process would fail, as [Marian Ganisin](https://issues.redhat.com/secure/ViewProfile.jspa?name=mganisin%40redhat.com) found.

This PR implements a waiting step for the `system-sidekiq` deployment before running the job for resyncing zync domains.

Fixes https://issues.redhat.com/browse/THREESCALE-8476

### Verification Steps

Those defined in https://github.com/3scale/3scale-operator/pull/758

Additionally, the operator logs should report a log line while waiting for the system-sidekiq deployment to be ready.

```
2022-06-22T15:09:29.795+0200	INFO	controllers.APIManagerRestore	system-sidekiq deployments not ready. Waiting	{"APIManagerRestore Controller": "example-apimanagerrestore-pvc", "APIManager": "apimanager1"}
```

### DIFF
--- a/controllers/apps/apimanagerrestore_logic_reconciler.go
+++ b/controllers/apps/apimanagerrestore_logic_reconciler.go
@@ -4,16 +4,7 @@ import (
 	"fmt"
 	"time"
 
-	appsv1alpha1 "github.com/3scale/3scale-operator/apis/apps/v1alpha1"
-	"github.com/3scale/3scale-operator/pkg/3scale/amp/component"
-	"github.com/3scale/3scale-operator/pkg/backup"
-	"github.com/3scale/3scale-operator/pkg/common"
-	"github.com/3scale/3scale-operator/pkg/reconcilers"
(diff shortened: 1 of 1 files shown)


MStokluska:
I have reviewed the code and it all looks good to me, however, was unable to verify due to the limitations on the clusters we are using.

MStokluska (approved):
Verified it on one of the 3scale clusters, code, and functionality work as expected.
Unit tests would be useful but in general looks good to me!
/lgtm

PR #771, opened 2022-07-25
remove unused job

eguzki:


sergioifg94 (approved):
I suppose this is related to templates having being removed recently. In that case lgtm

PR #775, opened 2022-08-19
Makefile go install

eguzki:
Additionally, fix some go lints


eguzki:
go 1.13 does not support this change. Closing until go is upgraded

PR #791, opened 2022-10-31
MGDAPI-4815 - Investigate adding APICast metrics to RHOAM prometheus

valerymo:
https://issues.redhat.com/browse/MGDAPI-4815 - Investigate adding APICast metrics to RHOAM prometheus
Use together with RHOAM PR https://github.com/integr8ly/integreatly-operator/pull/2981

### DIFF
--- a/pkg/3scale/amp/component/apicast.go
+++ b/pkg/3scale/amp/component/apicast.go
@@ -765,6 +765,7 @@ func (apicast *Apicast) productionServicePorts() []v1.ServicePort {
 	ports := []v1.ServicePort{
 		v1.ServicePort{Name: "gateway", Protocol: v1.ProtocolTCP, Port: 8080, TargetPort: intstr.FromInt(8080)},
 		v1.ServicePort{Name: "management", Protocol: v1.ProtocolTCP, Port: 8090, TargetPort: intstr.FromInt(8090)},
+		v1.ServicePort{Name: "metrics", Protocol: v1.ProtocolTCP, Port: 9421, TargetPort: intstr.FromInt(9421)},
 	}
 
 	if apicast.Options.ProductionHTTPSPort != nil {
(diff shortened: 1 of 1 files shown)


eguzki:
why do you need to expose metrics at the service if there is already a podmonitor object https://github.com/3scale/3scale-operator/blob/6bfa8e4f5429579320212948601b2bd9a564479a/pkg/3scale/amp/component/apicast_monitoring.go#L16?

valerymo:
Hi @eguzki ,  thank you very much for your review and reply.
Regarding PodMonitors - yes, I thought about this, but I saw that currently looks like in Rhoam is ServiceMonitors are used for metrics. And no any podmonitor in the cluster. So I thought that maybe I could use servicemonitor.
```
[~]$ oc get podmonitor --all-namespaces
No resources found
[~]$ oc get servicemonitor --all-namespaces |wc -l
101
[~]$ oc get servicemonitor --all-namespaces |grep rhoam |wc -l
16
$ oc whoami --show-server
https://api.valerym01.pf40.s1.devshift.org:6443
```

valerymo:
Hi @eguzki , last week we discussed several ways of make ApiCast metrics available in Rhoam/Observability Prometheus,
and looks like agree to add the metrics port to the services.

I copied your last notes here "I would  not implement reconciling podmonitors in your operator
let's add the metrics port to the services.
But the cleanest way is to have your Prometheus instance watch 3scale namespace's podmonitors"

Could you please approve this PR, if it looks ok for you.
Thank you!

eguzki:
We need this for RHOAM monitoring. Reasons explained [here](https://issues.redhat.com/browse/MGDAPI-4815?focusedCommentId=21240377&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-21240377)

eguzki (approved):

PR #781, opened 2022-08-23
THREESCALE-8502 3scale operator to have an ability to deal with multiple release paths

eguzki:
### what

[THREESCALE-8502](https://issues.redhat.com/browse/THREESCALE-8502) 3scale operator to have an ability to deal with multiple release paths

### how

Upgrade procedure in reconciliation loop. Actually, most of it was already in the normal reconciliation loop. Only DC triggers and pod template labels were moved

custom release upgrade procedure has been removed.

### verification steps

Deploy 2.12

```
git checkout 3scale-2.12.0-GA
```
```
make run
```

```yaml
k apply -f - <<EOF
apiVersion: v1
kind: Secret
metadata:
  creationTimestamp: null
  name: aws-auth
stringData:
  AWS_ACCESS_KEY_ID: testID
  AWS_SECRET_ACCESS_KEY: testkey
  AWS_BUCKET: testbucket
  AWS_REGION: us-east-1
type: Opaque
EOF
```
```yaml
k apply -f - <<EOF
---
apiVersion: apps.3scale.net/v1alpha1
kind: APIManager
metadata:
  name: apimanager1
spec:
  wildcardDomain: eguzki.apps.dev-eng-ocp4-operator.dev.3sca.net
  resourceRequirementsEnabled: false
  system:
    fileStorage:
      simpleStorageService:
        configurationSecretRef:
          name: aws-auth
EOF
```
wait for deployment to be ready
```
oc wait --for=condition=available apimanager/apimanager1 --timeout=-1s
apimanager.apps.3scale.net/apimanager1 condition met
```
Check apimanager version
```
k get apimanager apimanager1 -o yaml | yq e '.metadata.annotations' -
apps.3scale.net/apimanager-threescale-version: "2.12"
apps.3scale.net/threescale-operator-version: 0.9.0
```
Check imagestreams, for instance amp-apicast
```yaml
k get is amp-apicast -o yaml | yq e '.spec.tags' -
- annotations:
    openshift.io/display-name: AMP APIcast 2.12
  from:
    kind: DockerImage
    name: quay.io/3scale/3scale212:apicast-3scale-2.12.0-GA
  generation: 2
  importPolicy: {}
  name: "2.12"
  referencePolicy:
    type: Source
```
Check deployment config's triggers, for instance, apicast-production
```yaml
k get dc apicast-production -o yaml | yq e '.spec.triggers' -
- type: ConfigChange
- imageChangeParams:
    automatic: true
    containerNames:
      - system-master-svc
[...]

PR #795, opened 2022-11-18
bump to 3scale 2.14 and version 0.11.0

eguzki:
* bump operator version to 0.11.0
* Add CHANGELOG for 0.10.0
* ~reset upgrade procedure~ (not needed) :dancers:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,47 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- Bumped Go to 1.17 [#785](https://github.com/3scale/3scale-operator/pull/785)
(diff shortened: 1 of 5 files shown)


thomasmaas:
Care to explain why reset upgrade proceed is not needed @eguzki?

eguzki:
> Care to explain why reset upgrade proceed is not needed @eguzki?

Due to #781 

There is some cleaning up to be done "from time to time". For example, the content in the [upgrade package](https://github.com/3scale/3scale-operator/tree/019d10565614ec6091f6f8c0389a21ce8ed15336/pkg/3scale/amp/upgrade).  But only when old release upgrades get stacked and the stack grow too much to become a maintenance issue. From 2.12 to 2.13, there is only one specific upgrade step (besides common upgrading objects like images, tags and so on..). It is safe to leave in 2.14. But probably in 2.15 or 2.16 there is no need to keep it and we can safely remove it.

PR #797, opened 2022-12-05
Fix StsSpec table rendering

mijaros:
The table is not rendered due to invalid number of cells in separator row.


KevFan (approved):
Looks good to me ! 👍

PR #811, opened 2023-04-11
MGDAPI-5152 - remove duplicate FILE_UPLOAD_STORAGE env var

adam-cattermole:
`FILE_UPLOAD_STORAGE` env var is added as part of the config map, and added separately after resulting in duplicate env vars on some DeploymentConfigs.

On v4.11 when updating any of the deployment configs a new replication controller is created and there are warnings about a duplicate `FILE_UPLOAD_STORAGE` env var in the deploy logs. However on v4.13 these no longer show as warnings but as errors and the rollout is marked as failed. This duplicate env var is present as it is added from the base config map and then appended again.

*v4.11*
```
--> pre: Running hook pod ...
NOTE: Gem::Specification#has_rdoc= is deprecated with no replacement. It will be removed on or after 2018-12-01.
Gem::Specification#has_rdoc= called from /opt/deps/rubygems/github.com/3scale/prawn/prawn-external-gitcommit-88aead0d97e230d08e9f51a18711d01317965bfe/app/prawn-core.gemspec:36.
fatal: not a git repository (or any of the parent directories): .git
Removing :environment prerequisite from db:create
I, [2023-03-31T10:45:53.839087 #1] INFO -- : ActiveMerchant MODE set to 'production'
W, [2023-03-31T10:45:53.854215 #1] WARN -- [Bugsnag]: No valid API key has been set, notifications will not be sent
I, [2023-03-31T10:45:54.026795 #1] INFO -- : [Core] Using http://backend-listener:3000/internal/ as URL
OpenIdAuthentication.store is nil. Using in-memory store.
Creating scope :admins. Overwriting existing method User.admins.
Creating scope :by_name. Overwriting existing method Cinstance.by_name.
[core] non-native log levels verbose, notice, critical emulated using UNKNOWN severity
Backend Internal API version 3.4.3 status: ok
Connected to mysql2://root@system-mysql/system
Connected to redis://system-redis:6379/1
--> pre: Success
[...]

PR #816, opened 2023-04-14
[Backport #811 MGDAPI-5152] remove duplicate env

eguzki:
#### What

Provides support for OCP 4.13 for 3scale 2.13

Fix in master https://github.com/3scale/3scale-operator/pull/811

### DIFF
--- a/pkg/reconcilers/deploymentconfig_test.go
+++ b/pkg/reconcilers/deploymentconfig_test.go
@@ -572,19 +572,17 @@ func TestDeploymentConfigRemoveDuplicateEnvVarMutator(t *testing.T) {
 	cases := []struct {
 		testName        string
 		existingEnvs    []corev1.EnvVar
-		desiredEnvs     []corev1.EnvVar
 		expectedResult  bool
 		expectedNewEnvs []corev1.EnvVar
 	}{
-		{"NothingToReconcile", envsA, envsA, false, envsA},
-		{"EnvsReconciled", envsB, envsA, true, envsA},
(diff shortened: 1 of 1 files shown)


pkg/helper/envvarutils.go
```diff
@@ -107,3 +107,18 @@ func EnsureEnvVar(desired v1.EnvVar, existingEnvVars *[]v1.EnvVar) bool {
...
+	var result []v1.EnvVar
+
+	for idx := range envVars {
+		if _, ok := set[envVars[idx].Name]; !ok {
+			set[envVars[idx].Name] = envVars[idx]
+			result = append(result, envVars[idx])
+		}
+	}
+
+	return result
```
thomasmaas, line 123: Help me understand where in this function the env vars are actually removed.

thomasmaas (approved):

pkg/helper/envvarutils.go
eguzki: The map has the property that the map keys are unique. The method will iterate over all the env vars and try to create a map of env var key to the same env var.
```
{env var key A} -> {env var A}
{env var key B} -> {env var B}
{env var key C} -> {env var C}
```
On each iteration of the loop, the map can tell if the key has been inserted before or not (line 17) (in O(1) time). When there are duplicated env vars, the loop will try to insert the same key multiple times in the map. Only the first time will succeed. After that, when a duplicated key is found, the conditional (line 17) will evaluate to false and the duplicated env var will not be added to the list of env vars that is returned by the method.

PR #841, opened 2023-07-11
Update controller-runtime to v0.15.0 & client-go to v0.28.0-alpha.2 &…

austincunningham:
… double the requests and limits for the operator


eguzki:
As of today, latest productized Golang image comes with 1.19. We do not know yet the estimated time of arrival for Golang 1.20. It should be in few months. We have seen RHEL9 based images shipped with Golang 1.20.

austincunningham:
unit tests failing looks like its this bug https://github.com/kubernetes-sigs/controller-runtime/issues/2362

PR #845, opened 2023-07-24
[WIP] Threescale 6735 - Reconcile Secrets

Patryk-Stefanski:
### What 
[THREESCALE-6735](https://issues.redhat.com/browse/THREESCALE-6735) - Addition of label reconciliation for Apicast Custom Policies

### How

- The APIManager referencing the secret will have labels with the UID of the secrets being referenced.
- The controller will watch secrets with the `apimanager.apps.3scale.net/watched-by=apimanager` label.  The operator will get notified of the changes in the secret only when this label is present.
- Controller implements a custom mapper that:
  - Lists all the APIManager CRs with the secrets referencing the label 
  -  Creates a reconciliation request for each element in the list
- The deployment will be notified of the update to the secret through annotations in the pod template. The annotations in the pod template will specify the resource version of the secret so each time the secret is updated the resource version changes and it then updates the annotation in the pod template causing a redeployment of the pod. When the pod is redeployed it will read the updated content of the secret.

### Verification 

1.  Run the following commands to install the 3scale operator

```bash
go mod vendor 
go mod tidy 
make install
export NAMESPACE=3scale-test
oc new-project $NAMESPACE
make download
make run
```

2.  For the operator to create routes we need to mock the s3 credentials secret, run the following command.
```bash
kubectl apply -f - <<EOF
---
kind: Secret
apiVersion: v1
metadata: 
  name: s3-credentials
  namespace: 3scale-test
data: 
  AWS_BUCKET: UkVQTEFDRV9NRQ==
  AWS_REGION: UkVQTEFDRV9NRQ==
type: Opaque
EOF
```

3. Save the following to files to create custom policy secrets.

- apicast-policy.json 
```bash 
{
  "$schema": "http://apicast.io/policy-v1/schema#manifest#",
  "name": "APIcast Example Policy",
  "summary": "This is just an example.",
  "description": "This policy is just an example how to write your custom policy.",
  "version": "0.1",
  "configuration": {
    "type": "object",
    "properties": { }
  }
}
```

- example.lua
```bash
local setmetatable = setmetatable

local _M = require('apicast.policy').new('Example', '0.1')
local mt = { __index = _M }

function _M.new()
  return setmetatable({}, mt)
end

function _M:init()
  -- do work when nginx master process starts
end

function _M:init_worker()
  -- do work when nginx worker process is forked from master
end

function _M:rewrite()
  -- change the request before it reaches upstream
    ngx.req.set_header('X-CustomPolicy', 'customValue')
end

function _M:access()
  -- ability to deny the request before it is sent upstream
end

function _M:content()
  -- can create content instead of connecting to upstream
end

function _M:post_action()
  -- do something after the response was sent to the client
end

function _M:header_filter()
  -- can change response headers
end

function _M:body_filter()
  -- can read and change response body
  -- https://github.com/openresty/lua-nginx-module/blob/master/README.markdown#body_filter_by_lua
end

function _M:log()
  -- can do extra logging
end

function _M:balancer()
  -- use for example require('resty.balancer.round_robin').call to do load balancing
end

return _M
```

- init.lua
```bash 
return require('example')
```

4. Create apicast customPolicy secrets 
```bash
 oc create secret generic custom-policy2-secret \
  --from-file=./apicast-policy.json \
  --from-file=./init.lua \
  --from-file=./example.lua
  
   oc create secret generic custom-policy1-secret \
  --from-file=./apicast-policy.json \
  --from-file=./init.lua \
  --from-file=./example.lua
```

5. Add this label to the secrets `apimanager.apps.3scale.net/watched-by: apimanager`

6. Deploy the APIManager CR , specifying the secrets and wildcardDomain.
```bash
oc apply -f - <<EOF
kind: APIManager
metadata:
  name: 3scale-test
spec:
  system: 
    fileStorage: 
      simpleStorageService: 
        configurationSecretRef: 
          name: s3-credentials
  wildcardDomain: apps.pstefans.giq5.s1.devshift.org
  apicast:
    stagingSpec:
      customPolicies:
        - name: custom-policy1
          version: "0.1"
          secretRef:
             name: custom-policy1-secret
        - name: custom-policy2
          version: "0.1"
          secretRef:
              name: custom-policy2-secret
    productionSpec:
      customPolicies:
        - name: custom-policy1
          version: "0.1"
          secretRef:
             name: custom-policy1-secret
        - name: custom-policy2
          version: "0.1"
          secretRef:
              name: custom-policy2-secret
              
EOF
```

7. APIManager should have a label with the UIDs of the two secrets
8. Apicast production DC should have annotations with the two secret file names
9.  The pod template in the apicast production CR should have two annotations for the labels along with their resourceVersion number.
20. Add a new label to one of the secrets the new resourceVersion of the secret should update the pod template annotation in the apicast production DC

### DIFF
--- a/controllers/apps/apimanager_controller.go
+++ b/controllers/apps/apimanager_controller.go
@@ -19,6 +19,9 @@ package controllers
 import (
 	"context"
 	"fmt"
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	appsv1 "github.com/openshift/api/apps/v1"
 	routev1 "github.com/openshift/api/route/v1"
@@ -39,11 +42,15 @@ import (
 	"github.com/3scale/3scale-operator/pkg/helper"
 	"github.com/3scale/3scale-operator/pkg/reconcilers"
 	"github.com/3scale/3scale-operator/version"
+
+	apimachinerymetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // APIManagerReconciler reconciles a APIManager object
 type APIManagerReconciler struct {
 	*reconcilers.BaseReconciler
+	SecretLabelSelector apimachinerymetav1.LabelSelector
+	WatchedNamespace    string
 }
 
 // blank assignment to verify that APIManagerReconciler implements reconcile.Reconciler
@@ -123,13 +130,29 @@ func (r *APIManagerReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 }
 
 func (r *APIManagerReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	secretToApimanagerEventMapper := &SecretToApimanagerEventMapper{
+		K8sClient: r.Client(),
+		Logger:    r.Logger().WithName("secretToApimanagerEventMapper"),
+		Namespace: r.WatchedNamespace,
+	}
+
 	handlers := &handlers.APIManagerRoutesEventMapper{
 		K8sClient: r.Client(),
 		Logger:    r.Logger().WithName("APIManagerRoutesHandler"),
 	}
 
+	labelSelectorPredicate, err := predicate.LabelSelectorPredicate(r.SecretLabelSelector)
+	if err != nil {
+		return nil
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&appsv1alpha1.APIManager{}).
+		Watches(
+			&source.Kind{Type: &v1.Secret{}},
+			handler.EnqueueRequestsFromMapFunc(secretToApimanagerEventMapper.Map),
+			builder.WithPredicates(labelSelectorPredicate),
+		).
 		Owns(&appsv1.DeploymentConfig{}).
 		Watches(&source.Kind{Type: &routev1.Route{}}, handler.EnqueueRequestsFromMapFunc(handlers.Map)).
 		Complete(r)
--- a/controllers/apps/secret_to_apimanager_event_mapper.go
+++ b/controllers/apps/secret_to_apimanager_event_mapper.go
@@ -1,0 +1,58 @@
+package controllers
+
+import (
+	"context"
+	"fmt"
+	appsv1alpha1 "github.com/3scale/3scale-operator/apis/apps/v1alpha1"
+
+	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	APImanagerSecretLabelPrefix = "secret.apimanager.apps.3scale.net/"
+)
+
+// SecretToApimanagerEventMapper is an EventHandler that maps secret object to apimanager CR's
+type SecretToApimanagerEventMapper struct {
+	K8sClient client.Client
+	Logger    logr.Logger
+	Namespace string
+}
+
+func apimanagerSecretLabelKey(uid string) string {
+	return fmt.Sprintf("%s%s", APImanagerSecretLabelPrefix, uid)
+}
+
+func (s *SecretToApimanagerEventMapper) Map(obj client.Object) []reconcile.Request {
+
+	apimanagerList := &appsv1alpha1.APIManagerList{}
+
+	// filter by Secret UID
+	opts := []client.ListOption{client.HasLabels{apimanagerSecretLabelKey(string(obj.GetUID()))}}
+
+	// Support namespace scope or cluster scoped
+	if s.Namespace != "" {
+		opts = append(opts, client.InNamespace(s.Namespace))
+	}
+
+	err := s.K8sClient.List(context.Background(), apimanagerList, opts...)
+	if err != nil {
+		s.Logger.Error(err, "reading apimanager list")
+		return nil
+	}
+
+	s.Logger.V(1).Info("Processing object", "key", client.ObjectKeyFromObject(obj), "accepted", len(apimanagerList.Items) > 0)
+
+	requests := []reconcile.Request{}
+	for idx := range apimanagerList.Items {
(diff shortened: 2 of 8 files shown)


valerymo:
last step #10, not sure if it's working as expected:
```
$ oc describe secret/custom-policy1-secret |grep label
              newlabel=newlabel
$ oc get deploymentconfig apicast-production -ojson |grep label
        "labels": {
                "labels": {
$ oc get deploymentconfig apicast-production -ojson |grep new
$ 
```

MStokluska:
@valerymo we are asked to add a  `apimanager.apps.3scale.net/watched-by: apimanager` label to the secrets - this seems to me a bit counter-intuitive - if Operator is meant to reconcile something, why do we need to manually add a label to the resource? 

What I've also noticed that happens is that when I create a new secret without the label, and I add the custom policy pointing to that secret to apim, the apim labels gets annotated with the secret uid but the secret itself is missing the label `apimanager.apps.3scale.net/watched-by: apimanager` - is this expected?

valerymo:
> @valerymo we are asked to add a `apimanager.apps.3scale.net/watched-by: apimanager` label to the secrets - this seems to me a bit counter-intuitive - if Operator is meant to reconcile something, why do we need to manually add a label to the resource?

@MStokluska , it's just sample for testing. Label should be already in secret in real life. We just want to say that it should be done before Apimanager CR creation

valerymo:
> What I've also noticed that happens is that when I create a new secret without the label, and I add the custom policy pointing to that secret to apim, the apim labels gets annotated with the secret uid but the secret itself is missing the label `apimanager.apps.3scale.net/watched-by: apimanager` - is this expected?

@MStokluska  I will recheck. Thank you for comment

MStokluska:
> > @valerymo we are asked to add a `apimanager.apps.3scale.net/watched-by: apimanager` label to the secrets - this seems to me a bit counter-intuitive - if Operator is meant to reconcile something, why do we need to manually add a label to the resource?
> 
> @MStokluska , it's just sample for testing. Label should be already in secret in real life. We just want to say that it should be done before Apimanager CR creation.. Thank you

What is placing that label in the secret in "real life"?

MStokluska:
I've tested it on my cluster and here's a couple of questions/comments

1. When the secret is created, and then referenced in APIM, what currently happens is that the secret UID annotation is added to APIM but any changes to the secret do not trigger rollout of the DCs. 
There are 2 sub-question about this logic:
a) Who/What is to create the label "managed-by-operator" on the secret? IMO - any secret referenced in APIManager should be getting the label and the entity that sets that label would IMO be the 3scale operator. 
b) If a is not true, it's very counter-intuitive to me, but should we remove the UID annotation from the APIM since the updates to the secret only update the UID but it has no effect on the apicasts DCs.

2. When a secret has the label, and the APIM references it etc, but only in lets say - apicast staging, any changes to the secret trigger re-deploy of both, apicast stage and apicast production. IMO this should not be the case.

@austincunningham @eguzki I would appreciate your input/explanation in case point 1 is incorrect

valerymo:
> > > @valerymo we are asked to add a `apimanager.apps.3scale.net/watched-by: apimanager` label to the secrets - this seems to me a bit counter-intuitive - if Operator is meant to reconcile something, why do we need to manually add a label to the resource?
> > 
> > 
> > @MStokluska , it's just sample for testing. Label should be already in secret in real life. We just want to say that it should be done before Apimanager CR creation.. Thank you
> 
> What is placing that label in the secret in "real life"?

@MStokluska we discussed it already seems (?).  Please look at adding ` apimanager.apps.3scale.net/watched-by=apimanager` label as part of secret creation. So any software (I don't know how 300+ , hope number is correct : ) customers /users of 3scale are installing 3scale operator. It could be another operator, as RHOAM, it could be Helm or any other installation/operator-like software, and it will create these secrets, with or without labels, depends to their decision, if they want to use the feature. This is how I'm understand this Design. cc @austincunningham , @eguzki

MStokluska:
Thanks @valerymo as for #1 - thanks for confirming, let me know when #2 is fixed and I can re-review

valerymo:
> Thanks @valerymo as for # 1 - thanks for confirming, let me know when # 2 is fixed and I can re-review

Hi @MStokluska , issue fixed in PR https://github.com/3scale/3scale-operator/pull/861 , as I have no permission for this PR, I will Patric for permission or will decide on Monday, if continue in this PR or new one. 
log
```
[vmogilev@vmogilev 3scale-operator] (THREESCALE-6735)$ oc get dc
NAME                 REVISION   DESIRED   CURRENT   TRIGGERED BY
apicast-production   1          1         1         config,image(amp-apicast:2.14)
apicast-staging      2          1         1         config,image(amp-apicast:2.14)
backend-cron         1          1         1         config,image(amp-backend:2.14)
backend-listener     1          1         1         config,image(amp-backend:2.14)
backend-redis        1          1         1         config,image(backend-redis:2.14)
backend-worker       1          1         1         config,image(amp-backend:2.14)
system-app           1          1         1         config,image(amp-system:2.14)
system-memcache      1          1         1         config,image(system-memcached:2.14)
system-mysql         1          1         1         config,image(system-mysql:2.14)
system-redis         1          1         1         config,image(system-redis:2.14)
system-searchd       1          1         1         config,image(system-searchd:2.14)
system-sidekiq       1          1         1         config,image(amp-system:2.14)
zync                 1          1         1         config,image(amp-zync:2.14)
zync-database        1          1         1         config,image(zync-database-postgresql:2.14)
zync-que             1          1         1         config,image(amp-zync:2.14)

```
Thank you for comment and issue found.

briangallagher:
@Patryk-Stefanski Can you close this in favor of https://github.com/3scale/3scale-operator/pull/861 which should include all these changes and comments

valerymo:
@briangallagher , @Patryk-Stefanski , I closed this PR, hope it's ok. I created new PR (as noted in Descrpiption), that includes this PR + fix. Please notes that fix is not on work that done in current PR. It's fix of legacy issue that caused the problem with DCs update (problem was in AdditionalPodAnnotations that included both apicast staging  and production; the fix  in PR https://github.com/3scale/3scale-operator/pull/861 was for AdditionalPodAnnotations. No changes in code of current PR. Thanks

PR #846, opened 2023-07-26
THREESCALE-9001 / Added Zync Routes Resync Function

den-rgb:
[THREESCALE-9001](https://issues.redhat.com/browse/THREESCALE-9001)

# WHAT

There are instances where the "bundle exec rake zync:resync:domains" needs to run on the system-sidekiq pod to recreate routes.
This was previously done in RHOAM but now it will be managed by the threescale operator.

# TEST

- Checkout this branch
- Run the following commands: 
- - `make docker-build-only IMG=docker.io/<user>/3scale-operator:<tag>`
- - `make operator-image-push IMG=docker.io/<user>/3scale-operator:<tag>`
- - `make bundle-custom-build IMG=docker.io/<user>/3scale-operator:<tag> BUNDLE_IMG=docker.io/<user>/3scale-operator-bundles:<tag>`
- - `make bundle-image-push BUNDLE_IMG=docker.io/<user>/3scale-operator-bundles:<tag>`
- - `make bundle-run BUNDLE_IMG=docker.io/<user>/3scale-operator-bundles:<tag>`

- Once that runs and the operator is installed create the apimanager: 
- -  `oc new-project 3scale-test`
- -  Create the test secret:
``` 
oc apply -f - << EOF
kind: Secret
apiVersion: v1
metadata:
  name: s3-credentials
  namespace: 3scale-test
data:
  AWS_BUCKET: UkVQTEFDRV9NRQ==
  AWS_REGION: UkVQTEFDRV9NRQ==
type: Opaque
EOF
```

- -  Create the api manager (Make sure to change the wildcard domain to your personal cluster , you can find it in the url)
```
oc create -f - << EOF
kind: APIManager
metadata:
  name: apimanager-sample
  namespace: 3scale-test
spec:
  system:
    fileStorage:
      simpleStorageService:
        configurationSecretRef:
          name: s3-credentials
  wildcardDomain: apps.<cluster>.devshift.org

EOF
```
- - Check the progress of the api manager : 
- - `oc get apimanager apimanager-sample -oyaml | yq '.status'`

- Navigate to `Networking -> Routes` on the `3scale-test` namespace once all deployments are in the `Ready` status.
- If a route gets deleted it should be restored.

### Testing multitenancy

- go into Secrets -> system-seed -> get master password
- click the master route and put in (username: master) along with the password you got.
- Find Accounts -> Create account .
- In Networking -> routes there should be new routes created linked to the account you just made.
- Delete one of the `system-provider` or `system-developer` routes  associated with the new tenant they should also recreate.

-------------------------------------------------------------------------------------

# TODO

Implement unit testing for the additional logic in this pr.

### DIFF
--- a/pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
+++ b/pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
@@ -1,7 +1,13 @@
 package operator
 
 import (
+	"context"
+	"encoding/xml"
 	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+	"time"
 
 	appsv1alpha1 "github.com/3scale/3scale-operator/apis/apps/v1alpha1"
 	"github.com/3scale/3scale-operator/pkg/common"
@@ -19,6 +25,7 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type BaseAPIManagerLogicReconciler struct {
@@ -33,6 +40,16 @@ type baseAPIManagerLogicReconcilerCRDAvailabilityCache struct {
 	prometheusRuleCRDAvailable   *bool
 	podMonitorCRDAvailable       *bool
 	serviceMonitorCRDAvailable   *bool
+}
+
+type Accounts struct {
+    XMLName xml.Name `xml:"accounts"`
+    Account []Account `xml:"account"`
+}
+
+type Account struct {
+    AdminBaseURL string `xml:"admin_base_url"`
+    BaseURL      string `xml:"base_url"`
 }
 
 func NewBaseAPIManagerLogicReconciler(b *reconcilers.BaseReconciler, apiManager *appsv1alpha1.APIManager) *BaseAPIManagerLogicReconciler {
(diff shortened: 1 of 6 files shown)


pkg/helper/podHelper.go
```diff
@@ -0,0 +1,135 @@
...
+
+var _ PodExecutorInterface = &PodExecutor{}
+
+func NewPodExecutor(log l.Logger) *PodExecutor {
+	return &PodExecutor{
+		Log: log,
+	}
+}
+
+// ExecuteRemoteCommand exec command on specific pod and wait the command's output.
```
Patryk-Stefanski, line 33: Is this comment needed?

Patryk-Stefanski (commented):
Verified that a root got recreated after being deleted, and left one small comment.

pkg/3scale/amp/operator/base_apimanager_logic_reconciler.go
```diff
@@ -116,6 +133,269 @@ func (r *BaseAPIManagerLogicReconciler) ReconcileGrafanaDashboard(desired *grafa
...
+    }
+
+    if serviceCount >= expectedRoutes {
+        return true, nil
+    }
+
+    return false, fmt.Errorf("base routes not found")
+}
+
+func (r *BaseAPIManagerLogicReconciler) getAccountUrls() ([]string, []string, error){
```
eguzki, line 253: There is a library where we have this implemented: https://github.com/3scale/3scale-porta-go-client

pkg/helper/podHelper.go
```diff
@@ -0,0 +1,131 @@
...
+	})
+	if err != nil {
+		return "", "", errors.Wrapf(err, "Failed executing command %s on %s/%s", command, ns, podName)
+	}
+
+	return buf.String(), errBuf.String(), nil
+}
+
+func (p PodExecutor) ExecuteRemoteContainerCommand(ns string, podName string, container string, command []string) (string, string, error) {
+	kubeClient, restConfig, err := getClient()
```
eguzki, line 74: about this client, is not available from the controller manager?

pkg/helper/podHelper.go
```diff
@@ -0,0 +1,131 @@
...
+	)
+	exec, err := remotecommand.NewSPDYExecutor(restConfig, "POST", req.URL())
+	if err != nil {
+		return "", "", errors.Wrapf(err, "Failed executing command %s on %s/%s", command, ns, podName)
+	}
+
+	buf := &bytes.Buffer{}
+	errBuf := &bytes.Buffer{}
+
+	err = exec.Stream(remotecommand.StreamOptions{
```
eguzki, line 62: This is blocking, right? Meaning that the controller's runtime will be waiting for the process in the pod to be finished.

eguzki (changes requested):
Overall, I have serious doubts whether this is the right approach or not. Not saying it is wrongly implemetned, just saying we are taking some decisions that are controversial. 

1. The operator is acquiring knowledge about the expected routes. This something that belongs to 3scale internals. Although it is true that the `status` reconciler reads the basic installation's expected routes, this implementation takes one step further.
2. The implementation is part of the 3scale reconciler, blocking 3scale deployment when doing the task. The operator installation pattern is about creating resources and reading state to take actions. It should never be blocked executing one process in a pod or waiting of some resource to be ready. 
3. The process can be run multiple times in parallel. Note there are multiple event sources that trigger apimanager controller reconciliation. For instance, when any of the deploymentconfigs change status, the reconcile loop is executed.
4. Have you considered a different approach? For instance, one shot CRD (like ProxyConfigPromote CRD) that will trigger one (and only one) process to recreate the routes. So, when the user misses some routes, or restores a deleted namespace, an instance can be created to run re-sync. I understand that it is much better experience if the operator detects automatically missing routes and self-heals installation. Just rising discussion as the implementation adds complexity

austincunningham:
closing this as won't do

PR #856, opened 2023-08-09
THREESCALE-7836 Exposed new env var to set porta client insecureSkipVerify flag

carlkyrillos:
# Issue link
[THREESCALE-7836](https://issues.redhat.com/browse/THREESCALE-7836)

# What
Refactors the PortaClient to default the TLS InsecureSkipVerify flag to false and exposes a new environment variable, `INSECURE_SKIP_VERIFY_CLIENT` that if `1` will set the InsecureSkipVerify to true.

# Verification steps
TODO

PR #858, opened 2023-08-16
THREESCALE-7836 Refactor TLS InsecureSkipVerify logic to use annotations

carlkyrillos:
# Issue link
[THREESCALE-7836](https://issues.redhat.com/browse/THREESCALE-7836)

# What
This refactors the insecure skip verify logic for the porta client to be enabled via annotations on individual CRs rather than globally as an environment variable.

This PR also add documentation around supported environment variables.

# Verification steps

### DIFF
--- a/controllers/capabilities/activedoc_controller.go
+++ b/controllers/capabilities/activedoc_controller.go
@@ -144,7 +144,8 @@ func (r *ActiveDocReconciler) reconcileSpec(activeDocCR *capabilitiesv1beta1.Act
 		return statusReconciler, err
 	}
 
-	threescaleAPIClient, err := controllerhelper.PortaClient(providerAccount)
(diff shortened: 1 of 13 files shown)


MStokluska (approved):

eguzki:
After https://github.com/3scale/3scale-operator/pull/856, the operator verifies TLS connections. For backward compatibility or because I am using self-signed certs, me, as a user, how do I enable `insecureSkipVerify` ?? Let's say I install the operator using OLM (usual way). How do I enable skip verification?

Have you considered reading enable/disable skip TLS verification from an annotation of the CR? I find it more flexible than global setting at the operator level.

MStokluska:
Looks good to me, good job!

/lgtm

PR #784, opened 2022-09-28
THREESCALE-8754 Enhance replica reconciliation

eguzki:
### what
Implements https://issues.redhat.com/browse/THREESCALE-8754

* Remove adding replica values to the CR
* Reconciliation (implemented by the mutator) only added when the `replicas` field is set

Components
- [x] backend listener
- [x] backend worker
- [x] backend cron
- [x] apicast staging
- [x] apicast production
- [x] system app
- [x] system sidekiq
- [x] zync
- [x] zync que

### Verification steps

run env

```
make install
make run
```
Deploy APIManager CR with no replica definition
```yaml
k apply -f -<<EOF
---
apiVersion: v1
kind: Secret
metadata:
  creationTimestamp: null
  name: aws-auth
stringData:
  AWS_ACCESS_KEY_ID: somekey
  AWS_SECRET_ACCESS_KEY: something
  AWS_BUCKET: example.com
  AWS_REGION: us-east-1
type: Opaque
---
apiVersion: apps.3scale.net/v1alpha1
kind: APIManager
metadata:
  name: apimanager1
spec:
  wildcardDomain: example.com
  resourceRequirementsEnabled: false
  system:
    fileStorage:
      simpleStorageService:
        configurationSecretRef:
          name: aws-auth
EOF
```

Check that the APIManager  CR does not have replicas in the CR

```yaml
k get apimanager.apps.3scale.net/apimanager1 -o yaml
apiVersion: apps.3scale.net/v1alpha1
kind: APIManager
metadata:
  annotations:
    apps.3scale.net/apimanager-threescale-version: "2.13"
    apps.3scale.net/threescale-operator-version: 0.10.0
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"apps.3scale.net/v1alpha1","kind":"APIManager","metadata":{"annotations":{},"name":"apimanager1","namespace":"3scale-operator-test"},"spec":{"resourceRequirementsEnabled":false,"system":{"fileStorage":{"simpleStorageService":{"configurationSecretRef":{"name":"aws-auth"}}}},"wildcardDomain":"3scale-operator-test.apps.dev-eng-ocp4-operator.dev.3sca.net"}}
  creationTimestamp: "2022-09-29T09:34:19Z"
  generation: 2
  name: apimanager1
  namespace: 3scale-operator-test
  resourceVersion: "9968928"
  uid: 14854404-5511-43fc-bfbb-181155e50aa3
spec:
  apicast:
    managementAPI: status
[...]

PR #860, opened 2023-08-17
THREESCALE-7864 set async by default on backend work

austincunningham:
# Jira
[THREESCALE-7864](https://issues.redhat.com//browse/THREESCALE-7864)

# What
set async by default on backend work

# Verification
- Create a project `oc new-project 3scale-test`
- `make install`
- `make run`
- Create a s3 secret
```bash
kubectl apply -f - <<EOF
---
kind: Secret
apiVersion: v1
metadata:
  name: s3-credentials
data:
  AWS_BUCKET: UkVQTEFDRV9NRQ==
  AWS_REGION: UkVQTEFDRV9NRQ==
type: Opaque
EOF
```
- Create an apimanager CR using that secret e.g.
```bash
kubectl apply -f - <<EOF
---
kind: APIManager
metadata:
  name: apimanager-sample
spec:
  system:
    fileStorage:
      simpleStorageService:
        configurationSecretRef:
          name: s3-credentials
  wildcardDomain: apps.aucunnin.wx3t.s1.devshift.org
EOF
```
- Check the backend work pod for the `CONFIG_REDIS_ASYNC` envar
```bash
$ oc get po | grep backend-worker
backend-worker-1-deploy       0/1     Completed   0             12m
backend-worker-1-jn29j        1/1     Running     0             12m
$ oc exec backend-worker-1-jn29j -- env | grep CONFIG_REDIS_ASYNC
Defaulted container "backend-worker" out of: backend-worker, backend-redis-svc (init)
CONFIG_REDIS_ASYNC=1
```
- remove the envar from the backend-worker deployment config `spec.containers.env` and see if its reconciled back e.g. remove this
```yaml
- name: CONFIG_REDIS_ASYNC
  value: '1'
```

### DIFF
--- a/pkg/reconcilers/deploymentconfig.go
+++ b/pkg/reconcilers/deploymentconfig.go
@@ -10,6 +10,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/3scale/3scale-operator/pkg/3scale/amp/component"
 	"github.com/3scale/3scale-operator/pkg/common"
 	"github.com/3scale/3scale-operator/pkg/helper"
 )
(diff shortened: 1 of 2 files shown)


carlkyrillos:
Reviewed code changes and completed verification steps
/lgtm

pkg/reconcilers/deploymentconfig.go
```diff
@@ -2,6 +2,7 @@ package reconcilers
 
 import (
 	"fmt"
+	"github.com/3scale/3scale-operator/pkg/3scale/amp/component"
```
MStokluska, line 5: Seems that the import order is not in line with what we do in other files.

pkg/reconcilers/deploymentconfig.go
```diff
@@ -54,9 +55,23 @@ func GenericBackendMutators() []DCMutateFn {
 		DeploymentConfigPriorityClassMutator,
 		DeploymentConfigTopologySpreadConstraintsMutator,
 		DeploymentConfigPodTemplateAnnotationsMutator,
+		DeploymentConfigEnvMutator,
```
MStokluska, line 58: Since we are not doing this for listener, personally, instead of adding it to both, listener, cron and worker and then checking if dealing with worker proceed otherwise dont, I would add the mutator here instead:
https://github.com/3scale/3scale-operator/blob/master/pkg/3scale/amp/operator/backend_reconciler.go#L66

```
workerConfigMutator = append(workerConfigMutator, reconcilers.DeploymentConfigReplicasMutator, reconcilers.DeploymentConfigEnvMutator)
```
and remove the IF statement.

austincunningham: Think I'll change it up, I added the if as I didn't want manually applied ENV getting reverted on other backend components. But this is not needed using the individual mutator.

austincunningham: Hmm doesn't reconcile back the ENV when removed so I will have to stick to the original method

MStokluska: Thanks, we figured out why ( related to https://github.com/3scale/3scale-operator/pull/784 )

austincunningham: Ok so that logic only gets hit if you have replicas set for backendworker in the apimanager CR. Will leave as is

MStokluska:
/lgtm

Going to lgtm, personally, I would still move the env mutator specific to worker to only be applied as part of worker mutators but leaving it up to Austin to decide. Thanks Austin!

PR #862, opened 2023-08-29
THREESCALE-9315 set backend-listener to async mode also

austincunningham:
# Jira
[THREESCALE-9315](https://issues.redhat.com//browse/THREESCALE-9315)

# What
set async by default on backend worker and listener for non logical database see https://github.com/3scale/apisonator/blob/master/docs/openshift_horizontal_scaling.md#enable-async for reference

# Verification
## 3scale configured with Logical databases ( default installation)
- Create a project `oc new-project 3scale-test`
- `make install`
- `make run`
- Create a s3 secret
```bash
kubectl apply -f - <<EOF
---
kind: Secret
apiVersion: v1
metadata:
  name: s3-credentials
data:
  AWS_BUCKET: UkVQTEFDRV9NRQ==
  AWS_REGION: UkVQTEFDRV9NRQ==
type: Opaque
EOF
```
- get the default application router
```bash
```
- Create an apimanager CR using that secret e.g.
```bash
kubectl apply -f - <<EOF
---
kind: APIManager
metadata:
  name: apimanager-sample
spec:
  system:
    fileStorage:
      simpleStorageService:
        configurationSecretRef:
          name: s3-credentials
  wildcardDomain: $DOMAIN
EOF
```
- Check the backend worker pod for the `CONFIG_REDIS_ASYNC` envar
```bash
$ oc get po | grep backend-worker
backend-worker-1-deploy       0/1     Completed   0             12m
backend-worker-1-jn29j        1/1     Running     0             12m
$ oc exec backend-worker-1-jn29j -- env | grep CONFIG_REDIS_ASYNC
# should return nothing
```
- Check the backend listener pod for the `CONFIG_REDIS_ASYNC` envar
```bash
$ oc get po | grep backend-listener
backend-listener-1-deploy       0/1     Completed   0             12m
backend-listener-1-jn29j        1/1     Running     0             12m
$ oc exec backend-listener-1-jn29j -- env | grep CONFIG_REDIS_ASYNC
Defaulted container "backend-listener" out of: backend-listener, backend-redis-svc (init)
CONFIG_REDIS_ASYNC=0
$ oc exec backend-listener-1-jn29j -- env | grep LISTENER_WORKERS
# should return nothing
```
- delete the apimanger cr
- stop the 3scale operator from running
## 3scale configured with non logical databases
- you can Run [CRO](https://github.com/integr8ly/cloud-resource-operator) locally to provision 2 on cluster redis 
```bash
make cluster/prepare
make run
REDIS_NAME=redis-storage make cluster/seed/redis
REDIS_NAME=redis-que make cluster/seed/redis
```
The redis will be installed in the cloud-resources-operator ns there should be 2 secrets  `redis-que-sec` and `redis-storage-sec` with the port and the uri for the redis
- once they are created you can stop the CRO operator running
- create a key/value secret in the 3scale-test ns called `backend-redis` and add 2 name values with the connection uri and port for the env vars REDIS_QUEUES_URL & REDIS_STORAGE_URL 
image

- start the 3scale-operator running locally
- Create an apimanager CR with external components set for the backend
```bash
kubectl apply -f - <<EOF
---
kind: APIManager
metadata:
  name: apimanager-sample
spec:
  externalComponents:
    backend:
      redis: true
  system:
    fileStorage:
      simpleStorageService:
        configurationSecretRef:
          name: s3-credentials
  wildcardDomain: $DOMAIN
EOF
```
- Check the backend worker pod for the `CONFIG_REDIS_ASYNC` envar
```bash
$ oc get po | grep backend-worker
backend-worker-1-deploy       0/1     Completed   0             12m
backend-worker-1-jn29j        1/1     Running     0             12m
$ oc exec backend-worker-1-jn29j -- env | grep CONFIG_REDIS_ASYNC
Defaulted container "backend-worker" out of: backend-worker, backend-redis-svc (init)
CONFIG_REDIS_ASYNC=1
```
- Check the backend listener pod for the `CONFIG_REDIS_ASYNC` envar and `LISTENER_WORKERS`
```bash
$ oc get po | grep backend-listener
backend-listener-1-deploy       0/1     Completed   0             12m
backend-listener-1-jn29j        1/1     Running     0             12m
$ oc exec backend-listener-1-jn29j -- env | grep CONFIG_REDIS_ASYNC
Defaulted container "backend-listener" out of: backend-listener, backend-redis-svc (init)
CONFIG_REDIS_ASYNC=1
$ oc exec backend-listener-1-jn29j -- env | grep LISTENER_WORKERS
Defaulted container "backend-listener" out of: backend-listener, backend-redis-svc (init)
LISTENER_WORKERS=1
```
- Check the backend listener for the `falcon` Arg
```bash
$ oc get dc backend-listener -oyaml | yq '.spec.template.spec.containers[0].args'
- bin/3scale_backend
- -s
- falcon
- start
- -e
- production
- -p
- "3000"
- -x
- /dev/stdout
```

- remove the envar from the backend-worker  and backend-listener deployment config `spec.containers.env` and see if its reconciled back e.g. remove this
```yaml
- name: CONFIG_REDIS_ASYNC
  value: '1'
```
- set the value in env var CONFIG_REDIS_ASYNC to '0' and it should revert to '1'

### DIFF
--- a/pkg/3scale/amp/operator/backend_reconciler.go
+++ b/pkg/3scale/amp/operator/backend_reconciler.go
@@ -51,8 +51,8 @@ func (r *BackendReconciler) Reconcile() (reconcile.Result, error) {
 
 	listenerConfigMutator := reconcilers.GenericBackendMutators()
 	if redisStorageUrl != redisQueuesUrl {
-		listenerConfigMutator = append(listenerConfigMutator, reconcilers.DeploymentConfigEnvMutator)
-		listenerConfigMutator = append(listenerConfigMutator, reconcilers.DeploymentConfigArgsMutator)
+		listenerConfigMutator = append(listenerConfigMutator, reconcilers.DeploymentConfigListenerEnvMutator)
+		listenerConfigMutator = append(listenerConfigMutator, reconcilers.DeploymentConfigListenerArgsMutator)
 	}
 	if r.apiManager.Spec.Backend.ListenerSpec.Replicas != nil {
 		listenerConfigMutator = append(listenerConfigMutator, reconcilers.DeploymentConfigReplicasMutator)
@@ -78,7 +78,7 @@ func (r *BackendReconciler) Reconcile() (reconcile.Result, error) {
 	// Worker DC
 	workerConfigMutator := reconcilers.GenericBackendMutators()
 	if redisStorageUrl != redisQueuesUrl {
-		workerConfigMutator = append(workerConfigMutator, reconcilers.DeploymentConfigEnvMutator)
+		workerConfigMutator = append(workerConfigMutator, reconcilers.DeploymentConfigWorkerEnvMutator)
 	}
 	if r.apiManager.Spec.Backend.WorkerSpec.Replicas != nil {
 		workerConfigMutator = append(workerConfigMutator, reconcilers.DeploymentConfigReplicasMutator)
--- a/pkg/reconcilers/deploymentconfig.go
+++ b/pkg/reconcilers/deploymentconfig.go
@@ -57,74 +57,81 @@ func GenericBackendMutators() []DCMutateFn {
 	}
 }
 
-func DeploymentConfigArgsMutator(_, existing *appsv1.DeploymentConfig) (bool, error) {
+func DeploymentConfigListenerArgsMutator(_, existing *appsv1.DeploymentConfig) (bool, error) {
 	update := true
-	existing.Spec.Template.Spec.Containers[0].Args = []string{"bin/3scale_backend", "-s", "falcon", "start", "-e", "production", "-p", "3000", "-x", "/dev/stdout"}
+	falconArgs := []string{"bin/3scale_backend", "-s", "falcon", "start", "-e", "production", "-p", "3000", "-x", "/dev/stdout"}
+	if !reflect.DeepEqual(existing.Spec.Template.Spec.Containers[0].Args, falconArgs) {
+		existing.Spec.Template.Spec.Containers[0].Args = falconArgs
+		return update, nil
+	}
+	update = false
 	return update, nil
 }
 
-func DeploymentConfigEnvMutator(desired, existing *appsv1.DeploymentConfig) (bool, error) {
+func DeploymentConfigListenerEnvMutator(desired, existing *appsv1.DeploymentConfig) (bool, error) {
 	update := false
-
-	// Always set env var CONFIG_REDIS_ASYNC to 1 this logic is only hit when you don't have logical redis db
+	updateListenerWorkers := true
+	updateConfigRedisAsync := true
+	// This may be redundant as operator crashes if LISTENER_WORKERS=0
+	// Update LISTENER_WORKERS and CONFIG_REDIS_ASYNC to 1 if found
 	for envId, envVar := range existing.Spec.Template.Spec.Containers[0].Env {
 		if envVar.Name == "LISTENER_WORKERS" {
+			updateListenerWorkers = false
 			if envVar.Value == "0" {
 				existing.Spec.Template.Spec.Containers[0].Env[envId].Value = "1"
 				update = true
-				return update, nil
 			}
 		}
+		if envVar.Name == "CONFIG_REDIS_ASYNC" {
+			updateConfigRedisAsync = false
+			if envVar.Value == "0" {
+				existing.Spec.Template.Spec.Containers[0].Env[envId].Value = "1"
+				update = true
+
+			}
+		}
+		if update {
+			return update, nil
+		}
+	}
+	// if either updateListenerWorkers or updateConfigRedisAsync is true then proceed to the append logic
+	// to add the env var LISTENER_WORKERS and CONFIG_REDIS_ASYNC
+	if updateListenerWorkers || updateConfigRedisAsync {
+		update = true
+	} else {
+		update = false
+	}
+	if updateConfigRedisAsync {
+		existing.Spec.Template.Spec.Containers[0].Env = append(existing.Spec.Template.Spec.Containers[0].Env,
+			helper.EnvVarFromValue("CONFIG_REDIS_ASYNC", "1"))
+	}
+	if updateListenerWorkers {
+		existing.Spec.Template.Spec.Containers[0].Env = append(existing.Spec.Template.Spec.Containers[0].Env,
+			helper.EnvVarFromValue("LISTENER_WORKERS", "1"))
+	}
+
+	return update, nil
+}
+
+func DeploymentConfigWorkerEnvMutator(desired, existing *appsv1.DeploymentConfig) (bool, error) {
+	update := true
+	// Always set env var CONFIG_REDIS_ASYNC to 1 this logic is only hit when you don't have logical redis db
+	for envId, envVar := range existing.Spec.Template.Spec.Containers[0].Env {
 		if envVar.Name == "CONFIG_REDIS_ASYNC" {
 			if envVar.Value == "0" {
 				existing.Spec.Template.Spec.Containers[0].Env[envId].Value = "1"
 				update = true
 				return update, nil
 			}
+			update = false
+
 		}
 	}
-	// Always set env var create CONFIG_REDIS_ASYNC if not present
-	hasListenerWorkers := true
-	hasConfigRedisAsync := true
-	for _, envVar := range existing.Spec.Template.Spec.Containers[0].Env {
-		if existing.Name == "backend-worker" {
-			if envVar.Name == "CONFIG_REDIS_ASYNC" {
-				update = false
-				return update, nil
-			}
-		}
-		if existing.Name == "backend-listener" {
-			if envVar.Name == "LISTENER_WORKERS" {
-				hasListenerWorkers = false
-			}
-			if envVar.Name == "CONFIG_REDIS_ASYNC" {
-				hasConfigRedisAsync = false
-			}
-		}
-	}
-	if hasListenerWorkers || hasConfigRedisAsync {
-		update = true
-	} else {
-		update = false
-		return update, nil
-	}
-	if existing.Name == "backend-listener" {
-		if hasConfigRedisAsync {
-			existing.Spec.Template.Spec.Containers[0].Env = append(existing.Spec.Template.Spec.Containers[0].Env,
-				helper.EnvVarFromValue("CONFIG_REDIS_ASYNC", "1"))
-		}
-		if hasListenerWorkers {
-			existing.Spec.Template.Spec.Containers[0].Env = append(existing.Spec.Template.Spec.Containers[0].Env,
-				helper.EnvVarFromValue("LISTENER_WORKERS", "1"))
-		}
-		update = true
-	}
-	if existing.Name == "backend-worker" {
+	// Adds the env CONFIG_REDIS_ASYNC if not present
+	if update {
 		existing.Spec.Template.Spec.Containers[0].Env = append(existing.Spec.Template.Spec.Containers[0].Env,
 			helper.EnvVarFromValue("CONFIG_REDIS_ASYNC", "1"))
-		update = true
 	}
-
 	return update, nil
 }
 


pkg/reconcilers/deploymentconfig.go
```diff
@@ -55,18 +54,24 @@ func GenericBackendMutators() []DCMutateFn {
...
-			diff := cmp.Diff(existing.Spec.Template.Spec.Containers[0].Env, desired.Spec.Template.Spec.Containers[0].Env)
-			log.Info(fmt.Sprintf("%s spec.template.spec.containers[0].Env has changed: %s", common.ObjectInfo(desired), diff))
-			existing.Spec.Template.Spec.Containers[0].Env = desired.Spec.Template.Spec.Containers[0].Env
-			update = true
+	if !reflect.DeepEqual(existing.Spec.Template.Spec.Containers[0].Env, desired.Spec.Template.Spec.Containers[0].Env) {
+		diff := cmp.Diff(existing.Spec.Template.Spec.Containers[0].Env, desired.Spec.Template.Spec.Containers[0].Env)
+		log.Info(fmt.Sprintf("%s spec.template.spec.containers[0].Env has changed: %s", common.ObjectInfo(desired), diff))
+		for _, envVar := range existing.Spec.Template.Spec.Containers[0].Env {
+			if envVar.Name == "CONFIG_REDIS_ASYNC" {
+				return update, nil
```
MStokluska, line 67: If we are returning false update here and the CONFIG_REDIS_ASYNC has been set manually to 0, we are not updating it back to 1. Since we want the async mode on at all times I think this needs further refactor.

austincunningham: Good point ,  but are we then overriding the users desire to disable async ? 
Default async should be enabled at all time , but you can override it by manually changing the env var.

Not sure on this one, 
- I kind of side with, sensible defaults and the ability to override them.  
- On the other hand if user has enabled hpa then disabling async may cause difficulties. 

Maybe documentation that the env var is set by default to on and can be unset but should not be if hpa is enabled

Or logic to reconcile the change out . 

Not to hard to implement either way . :thinking:

austincunningham: Think I am leaning the other way now 
The operator will set this back to "1" for your own good. Docs changes might not avert a potential issue with pre-existing env vars set to "0"

MStokluska: I think that given there's no real benefit of having it in sync mode, enforcing async mode is good to have

austincunningham: Added the extra logic , and updated the verification steps.

MStokluska:
:eyes:

eguzki (dismissed):
Async mode on does not support logical databases https://github.com/3scale/apisonator/issues/135

So, default deployment (with backend redis having two logical databases for stats and queue) will be broken

We are currently using aysnc-redis v0.7.0. I have checked and latest is v0.8.0 and it does not seem to be addressed this issue.

```
* f11a366 - (3 months ago) Bump minor version. - Samuel Williams (HEAD -> main, tag: v0.8.0, origin/main, origin/HEAD)
* 8920797 - (3 months ago) Modernize gem. - Samuel Williams
* 100b4e0 - (3 months ago) More test fixes. - Samuel Williams
* 5a03fb9 - (3 months ago) Update tests. (#47) - Samuel Williams
* f67fd18 - (3 months ago) Relax dependency on protocol-redis. - Samuel Williams
* 36f3786 - (3 months ago) Add pubsub example. - Samuel Williams
* 71155be - (3 months ago) Modernize gem. (#46) - Samuel Williams
| * 03ecd4d - (3 months ago) Modernize gem. - Samuel Williams (origin/modernize-gem)
|/  
* 491f098 - (3 months ago) Disable performance spec. - Samuel Williams
* d9cdb85 - (3 months ago) Add bake-test to `gems.rb`. - Samuel Williams
* ab4f8ef - (9 months ago) A bunch of random examples. - Samuel Williams
* c3321ad - (1 year, 1 month ago) Bump minor version. - Samuel Williams (tag: v0.7.0)
```

Their issue asking for this was closed https://github.com/socketry/async-redis/issues/15
I would not expect this to be implemented soon unless we go for it ourselves

eguzki:
Looks like the backend listener needs more deployment changes. Check out the doc about HPA in backend.  https://github.com/3scale/apisonator/blob/master/docs/openshift_horizontal_scaling.md

Basically, instead of "pumas" (backend HTTP framework), async modes requires "falcons" (backend async HTTP framework).

austincunningham:
/hold 
hitting an issue with async where authentication fails for all product endpoints e.g. the welcome wizard 

image

austincunningham:
> /hold hitting an issue with async where authentication fails for all product endpoints e.g. the welcome wizard
> 
> image

fixed with  https://github.com/3scale/apisonator/pull/351

valerymo:
@austincunningham , seems need rebase

valerymo:
Hi @austincunningham , looks like CONFIG_REDIS_ASYNC and  LISTENER_WORKERS env vars are missing initial setting. DeploymentConfigEnvMutator() is handling initial setting case of env vars, but this method could be never entered.
To check DeploymentConfigEnvMutator() I did temporary changes in Reconciler  - https://github.com/austincunningham/3scale-operator/blob/THREESCALE-9315-backend-listener/pkg/3scale/amp/operator/backend_reconciler.go#L53 , + L79. 
So, in current logic - env vars are set only if `if redisStorageUrl != redisQueuesUrl`, otherwise - they missing.
If we need have initial setting of these env vars - need implement add/get initial values to/from secret, ...places like system.go, redis_options_provider, etc.
Thanks

pkg/3scale/amp/operator/backend_reconciler.go
```diff
@@ -38,7 +41,19 @@ func (r *BackendReconciler) Reconcile() (reconcile.Result, error) {
 	}
 
 	// Listener DC
+	backendRedisSecret := &v1.Secret{}
+	r.Client().Get(r.Context(), types.NamespacedName{
+		Name:      "backend-redis",
+		Namespace: r.apiManager.Namespace,
+	}, backendRedisSecret)
+	redisQueuesUrl := strings.TrimSuffix(string(backendRedisSecret.Data["REDIS_QUEUES_URL"]), "1")
```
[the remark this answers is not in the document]
austincunningham: @valerymo It is expected not to apply the env var on the default case as the default case has logical databases. The redis connection strings on a default case will always be as these are set by the operator
```
redis://backend-redis:6379/1
redis://backend-redis:6379/0
``` 
Sorry there may have been out of date verification steps for the default case if you are expecting the env var to be set. Will review the steps to confirm.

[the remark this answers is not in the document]
austincunningham: Yep out of date steps in the default case will reword them. I was originally setting both env vars to `0` but found that the default install failed if the `LISTENER_WORKER` set. So rather than set them to `0` I removed them.

pkg/reconcilers/deploymentconfig.go
```diff
@@ -57,6 +57,77 @@ func GenericBackendMutators() []DCMutateFn {
...
+		if envVar.Name == "CONFIG_REDIS_ASYNC" {
+			if envVar.Value == "0" {
+				existing.Spec.Template.Spec.Containers[0].Env[envId].Value = "1"
+				update = true
+				return update, nil
+			}
+		}
+	}
+	// Always set env var create CONFIG_REDIS_ASYNC if not present
+	hasListenerWorkers := true
```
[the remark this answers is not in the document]
austincunningham: addressed here https://github.com/3scale/3scale-operator/pull/862#discussion_r1390397727

pkg/3scale/amp/operator/backend_reconciler.go
```diff
@@ -62,6 +77,9 @@ func (r *BackendReconciler) Reconcile() (reconcile.Result, error) {
 
 	// Worker DC
 	workerConfigMutator := reconcilers.GenericBackendMutators()
+	if redisStorageUrl != redisQueuesUrl {
+		workerConfigMutator = append(workerConfigMutator, reconcilers.DeploymentConfigEnvMutator)
```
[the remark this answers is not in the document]
austincunningham: addressed here https://github.com/3scale/3scale-operator/pull/862#discussion_r1390397727

austincunningham:
> @austincunningham , I checked both cases. These are few notes.
> 
> 1. First case (default installation ) - not working as expected.  Env vars CONFIG_REDIS_ASYNC and LISTENER_WORKERS are missing in backend-listener / worker pods.
>    It happens as noted in code comments because initial urls REDIS_QUEUES_URL & REDIS_STORAGE_URL are differ just with suffix 1/0, and suffix is deleted here https://github.com/austincunningham/3scale-operator/blob/THREESCALE-9315-backend-listener/pkg/3scale/amp/operator/backend_reconciler.go#L49-L50, so Mutation method where initial values should be set - is not entered. It's only in 1st test case.
> 2. Second test case is working as expected
> 3. Seems to me that initial values for CONFIG_REDIS_ASYNC and LISTENER_WORKERS should be set NOT in Mutation method (that seems intended for changes),  Maybe initial values could be set similar to BACKEND_REDIS_SENTINEL_HOSTS (?), from secret (or simple hardcoded initial values =  0 ? : )
>    Thanks

See comment https://github.com/3scale/3scale-operator/pull/862#discussion_r1390397727 for 1 the default case.

pkg/3scale/amp/component/backend.go
```diff
@@ -2,16 +2,14 @@ package component
 
 import (
 	"fmt"
-	"strconv"
```
MStokluska, line 5: Small but I would fix the imports order

pkg/3scale/amp/operator/backend_reconciler.go
```diff
@@ -5,6 +5,9 @@ import (
 	"github.com/3scale/3scale-operator/pkg/3scale/amp/component"
 	"github.com/3scale/3scale-operator/pkg/helper"
 	"github.com/3scale/3scale-operator/pkg/reconcilers"
+	v1 "k8s.io/api/core/v1"
```
MStokluska, line 8: same as above comment

pkg/reconcilers/deploymentconfig.go
```diff
@@ -57,6 +57,77 @@ func GenericBackendMutators() []DCMutateFn {
 	}
 }
 
+func DeploymentConfigArgsMutator(_, existing *appsv1.DeploymentConfig) (bool, error) {
+	update := true
+	existing.Spec.Template.Spec.Containers[0].Args = []string{"bin/3scale_backend", "-s", "falcon", "start", "-e", "production", "-p", "3000", "-x", "/dev/stdout"}
+	return update, nil
```
MStokluska, line 63: is this always updating the dc on cluster? can we somehow check if the args already are in the expected state?

pkg/reconcilers/deploymentconfig.go
```diff
@@ -57,6 +57,77 @@ func GenericBackendMutators() []DCMutateFn {
...
+func DeploymentConfigEnvMutator(desired, existing *appsv1.DeploymentConfig) (bool, error) {
+	update := false
+
+	// Always set env var CONFIG_REDIS_ASYNC to 1 this logic is only hit when you don't have logical redis db
+	for envId, envVar := range existing.Spec.Template.Spec.Containers[0].Env {
+		if envVar.Name == "LISTENER_WORKERS" {
+			if envVar.Value == "0" {
+				existing.Spec.Template.Spec.Containers[0].Env[envId].Value = "1"
+				update = true
+				return update, nil
```
MStokluska, line 75: can we do a single update? 
In general I think it would be better to have specific mutator for worker and specific for listener instead of mixing them both.

pkg/reconcilers/deploymentconfig.go
```diff
@@ -57,6 +57,77 @@ func GenericBackendMutators() []DCMutateFn {
...
+	for _, envVar := range existing.Spec.Template.Spec.Containers[0].Env {
+		if existing.Name == "backend-worker" {
+			if envVar.Name == "CONFIG_REDIS_ASYNC" {
+				update = false
+				return update, nil
+			}
+		}
+		if existing.Name == "backend-listener" {
+			if envVar.Name == "LISTENER_WORKERS" {
+				hasListenerWorkers = false
```
MStokluska, line 98: To me this is a bit confusing to read, if env exists we are assigning false to "hasListenerWorkers" - something small but I would update the naming here or the values so that it is more readable. 

Also as noted above, I would change the flow to:

-    first check if envs are present
-    if not, add them with expected values
-    if yes, ensure that they are in expected state

MStokluska:
Thanks @austincunningham, to me, it could use some improvements around the env and args reconciliation functions.

pkg/reconcilers/deploymentconfig.go
austincunningham: Agree this would be better

pkg/3scale/amp/component/backend.go
austincunningham: sure can do

pkg/3scale/amp/operator/backend_reconciler.go
austincunningham: sure can do

pkg/reconcilers/deploymentconfig.go
austincunningham: Not apposed to the idea but its a bit `DRY` in a coding sense.  Although I always find DRY code to be more readable .  Yea will pull it out.

pkg/reconcilers/deploymentconfig.go
```diff
@@ -57,6 +57,77 @@ func GenericBackendMutators() []DCMutateFn {
...
+	existing.Spec.Template.Spec.Containers[0].Args = []string{"bin/3scale_backend", "-s", "falcon", "start", "-e", "production", "-p", "3000", "-x", "/dev/stdout"}
+	return update, nil
+}
+
+func DeploymentConfigEnvMutator(desired, existing *appsv1.DeploymentConfig) (bool, error) {
+	update := false
+
+	// Always set env var CONFIG_REDIS_ASYNC to 1 this logic is only hit when you don't have logical redis db
+	for envId, envVar := range existing.Spec.Template.Spec.Containers[0].Env {
+		if envVar.Name == "LISTENER_WORKERS" {
```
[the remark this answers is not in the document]
austincunningham: I think the present order is fine
- first check if envs are present
- if yes update the values to the expected state
- if not, add them with the expected values.

pkg/reconcilers/deploymentconfig.go
austincunningham: sure can  update the env name. 
As noted above  I believe the order is fine.

pkg/reconcilers/deploymentconfig.go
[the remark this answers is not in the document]
austincunningham: I think pulling these out into separate mutators will clean up the logic a bit

pkg/reconcilers/deploymentconfig.go
austincunningham: probably won't need the boolean when separated into different mutators for listener and worker .

pkg/reconcilers/deploymentconfig.go
austincunningham: As for single update that means adding more update variables and an extra if  to check and only update when both are true, not sure the juice is worth the squeeze :thinking:

pkg/3scale/amp/operator/backend_reconciler.go
austincunningham: Goland disagrees with you on this one, keeps reverting it :facepalm:

pkg/3scale/amp/component/backend.go
austincunningham: IDE keeps reverting these import orders , not a hill I am willing to die on.

pkg/reconcilers/deploymentconfig.go
austincunningham:
```go
func DeploymentConfigListenerArgsMutator(_, existing *appsv1.DeploymentConfig) (bool, error) {
	update := true
	falconArgs := []string{"bin/3scale_backend", "-s", "falcon", "start", "-e", "production", "-p", "3000", "-x", "/dev/stdout"}
	if reflect.DeepEqual(existing.Spec.Template.Spec.Containers[0].Args, falconArgs) {
		existing.Spec.Template.Spec.Containers[0].Args = falconArgs
		return update, nil
	}
	update = false
	return update, nil
}
```

austincunningham: Will see if this works

austincunningham: nope it doesn't

austincunningham: should have used  `!reflect.DeepEqual(existing.Spec.Template.Spec.Containers[0].Args, falconArgs)` :facepalm:

pkg/reconcilers/deploymentconfig.go
austincunningham: na still need them.

pkg/reconcilers/deploymentconfig.go
austincunningham: rejigged ti for a single update

pkg/reconcilers/deploymentconfig.go
austincunningham: renamed the env vars

PR #874, opened 2023-09-28
THREESCALE-10090 add tenant status and refactor the controller

MStokluska:
# What

Refactor tenant reconciler and add status block with detailed error messages

Jira: https://issues.redhat.com/browse/THREESCALE-10090

# Verification

Verification consists of few tests.

## Prepare for verification

- Create namespace

```
oc new-project threescale
```

- Install CRDs

```
make install
```

- Run the operator

```
make run
```

- Create new secret to be consumed by APIM

```
oc apply -f - <<EOF   
---
apiVersion: v1
kind: Secret
metadata:
  creationTimestamp: null
  name: s3-credentials
stringData:
  AWS_ACCESS_KEY_ID: something
  AWS_BUCKET: something
  AWS_REGION: us-east-1
type: Opaque
EOF
```

- Fetch cluster domain

```
```

- Create APIM

```
kubectl apply -f - <<EOF                                                        
---
kind: APIManager
metadata:
  name: apimanager-sample
  namespace: threescale
spec:
  system:
    fileStorage:
      simpleStorageService:
        configurationSecretRef:
          name: s3-credentials
  wildcardDomain: $DOMAIN
EOF
```

- Confirm 3scale is fully installed

```
oc wait --for=condition=available apimanager/apimanager-sample --timeout=-1s
```

## Tenant successfully created

- Fetch master route

```
MASTER_ROUTE=https://$(oc get route -n threescale | grep master |awk '{print $2}')
```

- Create tenant with admin secret

```
oc apply -f - <<EOF
---
apiVersion: v1
kind: Secret
metadata:
  name: example-admin-secret
type: Opaque
stringData:
  admin_password: "password"
---
kind: Tenant
metadata:
  name: example-tenant
spec:
  username: admin
  systemMasterUrl: $MASTER_ROUTE
  email: admin@example.com
  organizationName: example
  masterCredentialsRef:
    name: system-seed
  passwordCredentialsRef:
    name: example-admin-secret
  tenantSecretRef:
    name: example-tenant-secret
EOF
```
- Expected outcome is:

    - Tenant is created successfully - this can be double checked by navigating to the 3scale portal of the admin
    - Tenant CR has the following: tenant ID annotation, tenant finalizer, .status tenantID, adminID and 2 conditions, ready - true, failed - false

## Tenant successfully deleted

- Delete the tenant CR
```
oc delete tenant example-tenant
```
- Expected outcome is:
    - Tenant is successfully deleted and marked for deletion in 3scale

## Failed creation logging when missing tenant admin secret

- Delete the previously created secret (created when testing if tenant creation works as expected)
```
oc delete secret example-admin-secret
```
- Create tenant pointing to non-existing secret
```
oc apply -f - <<EOF
---
kind: Tenant
metadata:
  name: example-tenant
spec:
  username: admin
  systemMasterUrl: $MASTER_ROUTE
  email: admin@example.com
  organizationName: example
  masterCredentialsRef:
    name: system-seed
  passwordCredentialsRef:
    name: example-admin-secret
  tenantSecretRef:
    name: example-tenant-secret
EOF
```
- Expected outcome is:
    - failed condition - true, with message: `Secret "example-admin-secret" not found`

- Create the missing secret but with typo in admin password:
```
oc apply -f - <<EOF
---
apiVersion: v1
kind: Secret
metadata:
  name: example-admin-secret
type: Opaque
stringData:
  adminpassword: "password"
EOF
```
- Expected outcome is:
    - failed condition - true, with message: `not found admin password secret (threescale/example-admin-secret) -
        missing required attribute: admin_password`

- Patch the secret with proper name of the attribute:
```
oc apply -f - <<EOF
---
apiVersion: v1
kind: Secret
metadata:
  name: example-admin-secret
type: Opaque
stringData:
  admin_password: "password"
EOF
```
- Expected outcome is:
    - tenant successfully created and conditions status updated accordingly

- Delete the tenant:
```
oc delete tenant example-tenant
```

## Backup and restore tenant

- Create a tenant secret and tenant:
```
oc apply -f - <<EOF
---
apiVersion: v1
kind: Secret
metadata:
  name: backuprestore-admin-secret
type: Opaque
stringData:
  admin_password: "password"
---
kind: Tenant
metadata:
  name: backuprestore-tenant
spec:
  username: admin
  systemMasterUrl: $MASTER_ROUTE
  email: admin@example.com
  organizationName: backuprestore
  masterCredentialsRef:
    name: system-seed
  passwordCredentialsRef:
    name: backuprestore-admin-secret
  tenantSecretRef:
    name: example-tenant-secret
EOF
```

- Grep routes for backuprestore tenant
```
oc get route -n threescale | grep backuprestore |awk '{print $2}'
```

- There should be 4 routes created:
```
api-backuprestore-apicast-staging.<DOMAIN>
api-backuprestore-apicast-production.<DOMAIN>
backuprestore-admin.<DOMAIN>
backuprestore.<DOMAIN>
```
- Save the tenantCR for backup
```
oc get tenant backuprestore-tenant -o yaml >> tenant.yaml
```
- Stop the operator from running

- Remove the tenant CR (ensure to remove the finalizer as well)

- Run the operator again

- Apply the tenant CR (from the saved copy you got from previous steps)

```
oc apply -f tenant.yaml -n threescale
```
- Expected outcome is:
    - Tenant CR is populated with status again
    - Tenant is NOT re-created, to confirm run the routes command again:
    ```
    oc get route -n threescale | grep backuprestore |awk '{print $2}'
    ```
    - Amount of routes from command above should match 4.
    - Status fields are the same between what is in the status section of the CR on cluster and with your local copy


controllers/capabilities/tenant_controller.go
```diff
@@ -95,119 +107,116 @@ func (r *TenantReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 
-	insecureSkipVerify := controllerhelper.GetInsecureSkipVerifyAnnotation(tenantR.GetAnnotations())
-	portaClient, err := controllerhelper.PortaClientFromURLString(tenantR.Spec.SystemMasterUrl, masterAccessToken, insecureSkipVerify)
-	if err != nil {
-		reqLogger.Error(err, "Error creating porta client object")
-		// Error reading the object - requeue the request.
-		return ctrl.Result{}, err
+	// Requeue if either object or status were updated
+	if res.Requeue {
+		return reconcile.Result{}, nil
```
MStokluska, line 122: This looks off a bit so suggestions are very much welcomed.
The reasoning behind sending empty result with nil (which equals to no requeue) is that the only time we can land here is when status or spec got updated, which under the hood, automatically triggers a reconcile. 

For any error's we handle the requeue or not (based on error types).

controllers/capabilities/tenant_controller.go
```diff
@@ -95,119 +107,116 @@ func (r *TenantReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+			reqLogger.Info("ERROR", "wait error", reconcileErr)
+			return ctrl.Result{Requeue: true}, nil
 		}
 
-		controllerutil.RemoveFinalizer(tenantR, tenantFinalizer)
-		err = r.UpdateResource(tenantR)
-		if err != nil {
-			return ctrl.Result{}, err
+		// On orphan error, do not retry
+		if helper.IsOrphanSpecError(reconcileErr) {
```
MStokluska, line 140: The only time orphaned is used when tenant CR is marked for deletion but the tenant is already marked for deletion in 3scale as well. IMO this is not a validation error but is also an error that should not re-trigger the reconciler. 
We used to delete the CR however I've removed this logic since to me it seemed that the tenant/3scale must have been some weird state for this to happen. Now, I think we might also have the issue when tenant delete is successful but reconciler fails somewhere else (after deleting tenant from 3scale) - in that case we should remove CR.. any suggestions?

controllers/capabilities/tenant_internal_reconciler.go
```diff
@@ -104,54 +100,56 @@ func (r *TenantInternalReconciler) reconcileTenant() (ctrl.Result, error) {
...
+		return err
 	}
 
 	if adminUser == nil {
 		adminUser, err = controllerhelper.CreateAdminUser(r.portaClient,
 			tenantID, r.tenantR.Spec.Email, r.tenantR.Spec.Username)
 		if err != nil {
-			return ctrl.Result{}, err
+			return &helper.WaitError{
+				Err: fmt.Errorf("3scale client failed creating the admin user: %v", err),
```
MStokluska, line 139: Could be left as it was - however I found that some of the error responses from 3scale api aren't suitable to be displayed on the CR

controllers/capabilities/tenant_status_reconciler.go
```diff
@@ -0,0 +1,100 @@
...
+		BaseReconciler:   b,
+		resourceUpdated:  resourceUpdated,
+		resourceOriginal: resourceOriginal,
+		reconcileError:   reconcileError,
+		logger:           b.Logger().WithValues("Status Reconciler", resourceUpdated.Name),
+	}
+}
+
+// Compare and update the status of the tenant CR (both, object and object.status)
+func (s *TenantStatusReconciler) Reconcile() (reconcile.Result, error) {
```
MStokluska, line 34: Idea here is to have an update to the CR done in one place and one place only to avoid:
1) multiple reconcile loops triggering
2) avoid k8s errors based on resource out of sync

controllers/capabilities/tenant_status_reconciler.go
```diff
@@ -0,0 +1,100 @@
...
+			return reconcile.Result{}, err
+		}
+
+		return reconcile.Result{Requeue: true}, nil
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (s *TenantStatusReconciler) calculateStatus() capabilitiesv1alpha1.TenantStatus {
```
MStokluska, line 65: alternatively - we could only have single condition - ready or not (not = error/in progress)

controllers/capabilities/tenant_controller.go
```diff
@@ -87,6 +93,12 @@ func (r *TenantReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return ctrl.Result{}, err
 	}
 
+	// Ignore deleted resources, this can happen when foregroundDeletion is enabled
+	// https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#foreground-cascading-deletion
+	if tenantR.GetDeletionTimestamp() != nil && !controllerutil.ContainsFinalizer(tenantR, tenantFinalizer) {
+		return ctrl.Result{}, nil
+	}
+
 	if reqLogger.V(1).Enabled() {
```
eguzki, line 102: log the object read always (only when debug level). But also when deletion timestamp is added

controllers/capabilities/tenant_controller.go
```diff
@@ -95,119 +107,116 @@ func (r *TenantReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 
-	if !tenantFinalizerFound || !tenantIdAnnotationFound {
-		err = r.UpdateResource(tenantR)
-		if err != nil {
-			return ctrl.Result{}, err
-		}
-		return ctrl.Result{}, nil
-	}
+	// Adding annotations and finalizers
+	changedMetadata := r.reconcileMetadata(tenantCR)
```
eguzki, line 174: this shouldn't be in the `specReconciler`. The `specReconciler` should be called when the object is completed to be used for reconciliation.

controllers/capabilities/tenant_controller.go
```diff
@@ -87,6 +93,12 @@ func (r *TenantReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return ctrl.Result{}, err
 	}
 
+	// Ignore deleted resources, this can happen when foregroundDeletion is enabled
+	// https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#foreground-cascading-deletion
+	if tenantR.GetDeletionTimestamp() != nil && !controllerutil.ContainsFinalizer(tenantR, tenantFinalizer) {
```
eguzki, line 98: Keep the pattern:

```go
if tenantR.GetDeletionTimestamp() != nil && controllerutil.ContainsFinalizer(tenantR, tenantFinalizer) { 
   // handle deletion
}
if tenantR.GetDeletionTimestamp() != nil {         
    return ctrl.Result{}, nil                      
}                  
 if !controllerutil.ContainsFinalizer(tenantR, tenantFinalizer) {    
     controllerutil.AddFinalizer(tenantR, tenantFinalizer)           
     err = r.UpdateResource(tenantR)                                  
     if err != nil {                                                  
         return ctrl.Result{}, err                                    
     }                                                                
     return ctrl.Result{}, nil                                        
 }                                                                    
```

controllers/capabilities/tenant_controller.go
eguzki: Not sure I follow. 

If the 3scale tenant is either deleted (not found) or marked for deletion, it is safe to remove the finalizer and let the Tenant CR be removed.

eguzki, line 140: orphan state is conceptually between custom resources. Some custom resources may be children of other custom resources (like Account and application). 

For the tenant,  I do not thing this applies.

controllers/capabilities/tenant_controller.go
```diff
@@ -95,119 +107,116 @@ func (r *TenantReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		statusReconciler := NewTenantStatusReconciler(r.BaseReconciler, tenantCR, originalTenantCR, err)
+		return statusReconciler, err
 	}
 
-	tenantFinalizerFound := true
-	if !controllerutil.ContainsFinalizer(tenantR, tenantFinalizer) {
-		controllerutil.AddFinalizer(tenantR, tenantFinalizer)
-		tenantFinalizerFound = false
+	// Process tenant deletion
+	tenantInDeletion, err := r.reconcileDeletion(tenantCR, portaClient, logger)
```
eguzki, line 167: reconcile deletion in the `SpecReconciler` ??? I think this should be done elsewhere.

controllers/capabilities/tenant_controller.go
```diff
@@ -95,119 +107,116 @@ func (r *TenantReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
-	if changed {
-		err = r.UpdateResource(tenantR)
-		if err != nil {
-			return ctrl.Result{}, err
-		}
-		reqLogger.Info("Tenant resource updated with defaults")
-		// Expect for re-trigger
-		return ctrl.Result{}, nil
+	// Bring tenant secret ref and namespace to lower if setup incorrectly
+	changedDefaults := tenantCR.SetDefaults()
```
eguzki, line 177: if we change the tenantCR, we do not update it?

controllers/capabilities/tenant_controller.go
```diff
@@ -95,119 +107,116 @@ func (r *TenantReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+
+	if changedMetadata || changedDefaults {
+		statusReconciler := NewTenantStatusReconciler(r.BaseReconciler, tenantCR, originalTenantCR, err)
+		return statusReconciler, err
 	}
 
-	internalReconciler := NewTenantInternalReconciler(r.BaseReconciler, tenantR, portaClient, reqLogger)
-	res, err := internalReconciler.Run()
+	// Main logic for reconciling the tenant CR against 3scale
+	internalReconciler := NewTenantInternalReconciler(r.BaseReconciler, tenantCR, portaClient, logger)
```
eguzki, line 185: I would rename it to `TenantThreescaleReconciler`

no need for a "internal reconciler"

controllers/capabilities/tenant_controller.go
```diff
@@ -217,10 +226,70 @@ func (r *TenantReconciler) fetchMasterCredentials(tenantR *capabilitiesv1alpha1.
 
 	masterAccessTokenByteArray, ok := masterCredentialsSecret.Data[component.SystemSecretSystemSeedMasterAccessTokenFieldName]
 	if !ok {
-		return "", fmt.Errorf("Key not found in master secret (%s) key: %s",
-			tenantR.MasterSecretKey(),
-			component.SystemSecretSystemSeedMasterAccessTokenFieldName)
+		return "", &helper.WaitError{
```
eguzki, line 229: nice!

controllers/capabilities/tenant_controller.go
```diff
@@ -95,119 +107,116 @@ func (r *TenantReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		return statusReconciler, err
 	}
 
-	tenantFinalizerFound := true
-	if !controllerutil.ContainsFinalizer(tenantR, tenantFinalizer) {
-		controllerutil.AddFinalizer(tenantR, tenantFinalizer)
-		tenantFinalizerFound = false
+	// Process tenant deletion
+	tenantInDeletion, err := r.reconcileDeletion(tenantCR, portaClient, logger)
+	if err != nil || tenantInDeletion {
```
eguzki, line 168: I wonder if a tenant can be in deletion state for days, weeks or even months. If that is the case, it does not make sense to have the operator trying again and again.

eguzki (changes requested):

controllers/capabilities/tenant_status_reconciler.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+}
+
+// Compare and update the status of the tenant CR (both, object and object.status)
+func (s *TenantStatusReconciler) Reconcile() (reconcile.Result, error) {
+	// Check for changes to spec or metadata
+	equalSpec := s.resourceOriginal.SpecEqual(s.resourceUpdated, s.logger)
+	if !equalSpec {
+		s.logger.Info("updating tenant object")
+		err := s.Client().Update(context.TODO(), s.resourceUpdated)
```
eguzki, line 39: The tenant status reconciler can update the object (the spec)... I would not do this.

controllers/capabilities/tenant_status_reconciler.go
eguzki: My take is that we have a single condition. The tenant can be ready right after being created. I would have "Ready" or "Available" only. Their state can be `false` and still you have `message` and `reason` to specify further what is going on. For instance if the user input is required (like when the secret lacks some field)

controllers/capabilities/tenant_controller.go
MStokluska: We do, so the general flow of this PR is that the updates to the CR happen only in the statusReconciler. 
If the meta or defaults are changed the https://github.com/3scale/3scale-operator/pull/874/files#diff-70cc47073a2159ac4f5b5dd5b01d060d1dc3b898c37c0e322e739a223b09665cR36 picks it up and updates the spec.

controllers/capabilities/tenant_controller.go
MStokluska: Ok, I will remove the CR if tenant is already marked for deletion.

controllers/capabilities/tenant_controller.go
MStokluska: I agree. But currently I'm doing it in mind with adding unit tests. Ideally the delete should be one of the first things to do in the main reconciler however, since the tenant controller requires porta for deletion, we would need to first progress the epic around mocking porta on top level reconciler (since we already got a way of doing it within functions called from the reconciler I moved it here to be able to mock it with what we have).

So my only alternative is to move it to the main reconciler (as it was) without unit testing it until we get the tests setup in place.

controllers/capabilities/tenant_controller.go
MStokluska: Yes it can, currently, if tenant is marked for deletion already - we do not retrigger the reconciler on the CR because the current error type is isOrphaned.
As noted on following comments the improvements we can make are to remove the isOrphaned error and simply, remove the CR.

MStokluska:
@eguzki to summarise (might be more readable than addressing comments above)

Your suggestions seems to me to be as follows:
- Do not reconcile the spec within the specReconciler - do it before the specReconciler is called. This would be the main reconciler. 
- If so, the deletion logic would require porta config at the reconciler level - which is fine but currenty we won't be unit testing it until we resolve how to mock 3scale porta at main reconciler level (part of different issue)
- StatusReconciler will be reconciling only the status
- Remove usage of orphanedError and instead, remove the CR immediately if the tenant is already marked for deletion

Does that summarize the comments?

controllers/capabilities/tenant_controller.go
MStokluska: Updated, thank you

controllers/capabilities/tenant_status_reconciler.go
MStokluska: Updated

controllers/capabilities/tenant_status_reconciler.go
MStokluska: Updated, thank you

controllers/capabilities/tenant_controller.go
MStokluska: Updated, thank you

controllers/capabilities/tenant_controller.go
MStokluska: Updated, thank you

controllers/capabilities/tenant_controller.go
MStokluska: Updated, thank you

controllers/capabilities/tenant_controller.go
MStokluska: Updated, thank you

controllers/capabilities/tenant_status_reconciler.go
MStokluska: Updated, thank you

apis/capabilities/v1alpha1/tenant_types.go
```diff
@@ -18,16 +18,30 @@ package v1alpha1
...
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+const (
+	// TenantReadyConditionType indicates the tenant has been successfully created.
+	// Steady state
+	TenantReadyConditionType common.ConditionType = "Ready"
+
+	// TenantFailedConditionType indicates that an error occurred during creation.
+	// The operator will retry.
+	TenantFailedConditionType common.ConditionType = "Failed"
```
MStokluska, line 42: To be removed

eguzki (changes requested):
looking good. It goes in the right direction IMO. I left some comments, tho

controllers/capabilities/tenant_controller.go
MStokluska: Not sure I follow this comment

pkg/helper/reconcile_error_handler.go
```diff
@@ -0,0 +1,30 @@
+package helper
+
+import (
+	"github.com/go-logr/logr"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+// ReconcileErrorHandler reads the error type and based on that, it responsds with either a requeue true or false
```
carlkyrillos, line 8:
```suggestion
// ReconcileErrorHandler reads the error type and based on that, it responds with either a requeue true or false
```
Minor typo

carlkyrillos (commented):
Completed verification steps and reviewed the changed files. The `Tenant successfully created` section in the verification steps references 2 conditions on the Tenant CR but I only saw 1. I'm guessing the steps just weren't updated after removing the `Failed` type condition.

/lgtm

controllers/capabilities/tenant_controller.go
MStokluska: Hey @eguzki 

We have the intial lgtm on it, I wonder if you could explain to me what exactly is the ask here to move this PR forward since it looks like the last comment left.

eguzki: You fixed already. Before the fix, the object was not logged when deletion timestamp was added to the object.

eguzki (approved):

PR #877, opened 2023-10-06
OCP version managemenet

eguzki:
### What

[Controlling Operator compatibility with OpenShift Container Platform versions](https://docs.openshift.com/container-platform/4.13/operators/operator_sdk/osdk-working-bundle-images.html#osdk-control-compat_osdk-working-bundle-images) at the upstream level.

### DIFF
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -10,8 +10,7 @@ LABEL operators.operatorframework.io.metrics.mediatype.v1=metrics+v1
 LABEL operators.operatorframework.io.metrics.project_layout=go.kubebuilder.io/v2
 LABEL operators.operatorframework.io.test.config.v1=tests/scorecard/
 LABEL operators.operatorframework.io.test.mediatype.v1=scorecard+v1
-LABEL com.redhat.openshift.versions: v4.11-v4.14
-
+LABEL com.redhat.openshift.versions=v4.11-v4.14
 COPY bundle/manifests /manifests/
 COPY bundle/metadata /metadata/
 COPY bundle/tests/scorecard /tests/scorecard/
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -10,6 +10,8 @@ LABEL operators.operatorframework.io.metrics.mediatype.v1=metrics+v1
 LABEL operators.operatorframework.io.metrics.project_layout=go.kubebuilder.io/v2
 LABEL operators.operatorframework.io.test.config.v1=tests/scorecard/
 LABEL operators.operatorframework.io.test.mediatype.v1=scorecard+v1
+LABEL com.redhat.openshift.versions: v4.11-v4.14
+
 COPY bundle/manifests /manifests/
 COPY bundle/metadata /metadata/
 COPY bundle/tests/scorecard /tests/scorecard/
--- a/bundle/manifests/3scale-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/3scale-operator.clusterserviceversion.yaml
@@ -202,9 +202,11 @@ metadata:
     capabilities: Deep Insights
     categories: Integration & Delivery
     certified: "false"
+    com.redhat.openshift.versions: v4.11-v4.14
     containerImage: quay.io/3scale/3scale-operator:master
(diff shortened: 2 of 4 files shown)


eguzki:
Having an issue with the `operator-sdk` tool. It does not allow me to add bundle annotations. They are being removed when running `operator-sdk generate bundle` command. 

```diff
diff --git a/bundle.Dockerfile b/bundle.Dockerfile
index 6c97c553..8c812a23 100644
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -10,7 +10,6 @@ LABEL operators.operatorframework.io.metrics.mediatype.v1=metrics+v1
 LABEL operators.operatorframework.io.metrics.project_layout=go.kubebuilder.io/v2
 LABEL operators.operatorframework.io.test.config.v1=tests/scorecard/
 LABEL operators.operatorframework.io.test.mediatype.v1=scorecard+v1
-LABEL com.redhat.openshift.versions=v4.11-v4.14
 COPY bundle/manifests /manifests/
 COPY bundle/metadata /metadata/
 COPY bundle/tests/scorecard /tests/scorecard/
diff --git a/bundle/metadata/annotations.yaml b/bundle/metadata/annotations.yaml
index df5c05bf..7d7cbc95 100644
--- a/bundle/metadata/annotations.yaml
+++ b/bundle/metadata/annotations.yaml
@@ -9,4 +9,3 @@ annotations:
   operators.operatorframework.io.metrics.project_layout: go.kubebuilder.io/v2
   operators.operatorframework.io.test.config.v1: tests/scorecard/
   operators.operatorframework.io.test.mediatype.v1: scorecard+v1
-  com.redhat.openshift.versions: v4.11-v4.14
```

Maybe upgrade needed?

MStokluska:
Any reason why you want that annotation in dockerfile? 
IMO it doesn't need to be there, as long as the CSV contains the annotation we are fine?

eguzki:
> Any reason why you want that annotation in dockerfile? IMO it doesn't need to be there, as long as the CSV contains the annotation we are fine?

In the CSV you can specify `olm.properties: '[{"type": "olm.maxOpenShiftVersion", "value": "4.14"}]'`, for preventing cluster upgrades if the operator does not support it. But the catalog target (`com.redhat.openshift.versions=v4.11-v4.14`) needs to be as label in the bundle image as well as in the `bundle/metadata/annotations.yaml` file. Let me know if I am wrong.

MStokluska:
> > Any reason why you want that annotation in dockerfile? IMO it doesn't need to be there, as long as the CSV contains the annotation we are fine?
> 
> In the CSV you can specify `olm.properties: '[{"type": "olm.maxOpenShiftVersion", "value": "4.14"}]'`, for preventing cluster upgrades if the operator does not support it. But the catalog target (`com.redhat.openshift.versions=v4.11-v4.14`) needs to be as label in the bundle image as well as in the `bundle/metadata/annotations.yaml` file. Let me know if I am wrong.

 


> > Any reason why you want that annotation in dockerfile? IMO it doesn't need to be there, as long as the CSV contains the annotation we are fine?
> 
> In the CSV you can specify `olm.properties: '[{"type": "olm.maxOpenShiftVersion", "value": "4.14"}]'`, for preventing cluster upgrades if the operator does not support it. But the catalog target (`com.redhat.openshift.versions=v4.11-v4.14`) needs to be as label in the bundle image as well as in the `bundle/metadata/annotations.yaml` file. Let me know if I am wrong.

Following docs https://docs.openshift.com/dedicated/operators/operator_sdk/osdk-working-bundle-images.html it seems that the values in dockerfile and annotations should be added manually.

There's also this PR but it doesn't look like it's merged - https://github.com/operator-framework/operator-sdk/pull/5560 

I think we could investigate further to see if the issue of syncing annotations info to dockerfile was resolved but IMO, we need sdk bump since it's definitely not working as expected on current version we are using.

eguzki:
> I think we could investigate further to see if the issue of syncing annotations info to dockerfile was resolved but IMO, we need sdk bump since it's definitely not working as expected on current version we are using.

It seems to me that even with the latest operator-sdk, the issue is still there. We should test that first.

PR #882, opened 2023-10-17
THREESCALE-10229 bump 2.12 go version

MStokluska:
To verify:

- prow checks/circle CI checks should be efficient
- alternatively provision 4.11 OCP/OSD and install the operator following dev guide

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,15 +62,15 @@ commands:
     steps:
       - restore_cache:
           keys:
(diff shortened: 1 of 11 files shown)


MStokluska:
Deploying template job seems to be failing also for no reason in this PR https://github.com/3scale/3scale-operator/pull/852 - I think we are fine.

eguzki (approved):

PR #884, opened 2023-10-20
THREESCALE-10098 Remove Prometheus Rule 'ThreescaleApicastRequestTime'

ConorOM1:
# Issue Link

Jira: https://issues.redhat.com/browse/THREESCALE-10098

# What

The rule 'ThreescaleApicastRequestTime' will be removed in the next 3scale release.

# Verification Steps
- `checkout this branch`
- provision a cluster

when cluster is ready:
- set up 3scale operator locally, https://github.com/3scale/3scale-operator/blob/master/doc/development.md#run-3scale-operator-locally but don't  `make run` the operator yet.

- create a new `secret` resource

-  add an API Manager CR. In the spec, ensure to enable monitoring like so:
```
monitoring:
  enablePrometheusRules: true
  enabled: true
```

- now `make run` to start the operator

- open the `PrometheusRule` CRs & in the `apicast` YAML verify that the 'ThreescaleApicastRequestTime' rule does not exist.

- at this stage `stop` the operator

- go back into `apicast` and create your own rule, name it anything you like. Also, rename one of the existing rules to 'ThreescaleApicastRequestTime'.

- `make run` the operator again

- `apicast` YAML should now request a reload 

- after reload, verify that your own created rule has persisted & the 'ThreescaleApicastRequestTime' rule has been successfully removed.

### DIFF
--- a/pkg/3scale/amp/operator/apicast_reconciler.go
+++ b/pkg/3scale/amp/operator/apicast_reconciler.go
@@ -181,7 +181,7 @@ func (r *ApicastReconciler) Reconcile() (reconcile.Result, error) {
 		return reconcile.Result{}, err
 	}
 
-	err = r.ReconcilePrometheusRules(apicast.ApicastPrometheusRules(), reconcilers.UpdatePrometheusRulesMutator)
+	err = r.ReconcilePrometheusRules(apicast.ApicastPrometheusRules(), reconcilers.RemovePrometheusRulesMutator)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
--- a/pkg/3scale/amp/operator/apicast_reconciler.go
+++ b/pkg/3scale/amp/operator/apicast_reconciler.go
@@ -181,7 +181,12 @@ func (r *ApicastReconciler) Reconcile() (reconcile.Result, error) {
 		return reconcile.Result{}, err
 	}
 
-	err = r.ReconcilePrometheusRules(apicast.ApicastPrometheusRules(), reconcilers.UpdatePrometheusRulesMutator)
+	err = r.ReconcilePrometheusRules(apicast.ApicastPrometheusRules(), reconcilers.CreateOnlyMutator)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	err = r.ReconcilePrometheusRules(apicast.ApicastPrometheusRules(), reconcilers.RemovePrometheusRulesMutator)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
--- a/pkg/3scale/amp/operator/apicast_reconciler.go
+++ b/pkg/3scale/amp/operator/apicast_reconciler.go
@@ -181,7 +181,7 @@ func (r *ApicastReconciler) Reconcile() (reconcile.Result, error) {
 		return reconcile.Result{}, err
 	}
 
-	err = r.ReconcilePrometheusRules(apicast.ApicastPrometheusRules(), reconcilers.CreateOnlyMutator)
+	err = r.ReconcilePrometheusRules(apicast.ApicastPrometheusRules(), reconcilers.UpdatePrometheusRulesMutator)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
--- a/pkg/3scale/amp/operator/apicast_reconciler.go
+++ b/pkg/3scale/amp/operator/apicast_reconciler.go
@@ -181,12 +181,7 @@ func (r *ApicastReconciler) Reconcile() (reconcile.Result, error) {
 		return reconcile.Result{}, err
 	}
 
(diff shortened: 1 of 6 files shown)


austincunningham (dismissed):
/lgtm 
Verified the code yesterday

austincunningham (changes requested):
issue when you add an extra rule

pkg/3scale/amp/operator/apicast_reconciler.go
```diff
@@ -186,6 +186,11 @@ func (r *ApicastReconciler) Reconcile() (reconcile.Result, error) {
 		return reconcile.Result{}, err
 	}
 
+	err = r.ReconcilePrometheusRules(apicast.ApicastPrometheusRules(), reconcilers.RemovePrometheusRulesMutator)
```
eguzki, line 189: calling twice `ReconcilePrometheusRules` with the same rules and different mutator? that does not seem to be correct.

eguzki (changes requested):

pkg/3scale/amp/operator/apicast_reconciler.go
ConorOM1: amended. single call with new mutator

austincunningham:
# Verification 
- [x] installed this branch locally 
- [x] confirmed that  `ThreescaleApicastRequestTime` rule is not present
```bash
oc get prometheusrule apicast -oyaml | grep ThreescaleApicastRequestTime
```
- [x] Stopped the operator from running
- [x] added two rules to the apicast prometheusRule array one of which was  `ThreescaleApicastRequestTime` and the other was `ThreescaleApicastHappyTime` 
- [x] confirmed that only  `ThreescaleApicastRequestTime` rule was removed
```bash
~ oc get prometheusrule apicast -oyaml | grep ThreescaleApicastRequestTime
~ oc get prometheusrule apicast -oyaml | grep ThreescaleApicastHappyTime 
    - alert: ThreescaleApicastHappyTime
```
- [x] Verified

pkg/reconcilers/base_reconciler.go
```diff
@@ -31,6 +31,44 @@ func CreateOnlyMutator(existing, desired common.KubernetesObject) (bool, error)
...
+
+		log.Info("Alert 'ThreescaleApicastRequestTime' removed from PrometheusRules")
+		return true, nil
+	}
+	log.Info("Alert 'ThreescaleApicastRequestTime' not found, no update required.")
+	return false, nil
+}
+
+func UpdatePrometheusRulesMutator(existing, desired common.KubernetesObject) (bool, error) {
+	created, err := CreateOnlyMutator(existing, desired)
```
eguzki, line 60: `CreateOnlyMutator` is special mutator that will never update the existing resource, that is why it is called "CreateOnlyMutator". You can safely remove it from this function.

pkg/reconcilers/base_reconciler.go
```diff
@@ -31,6 +31,44 @@ func CreateOnlyMutator(existing, desired common.KubernetesObject) (bool, error)
 	return false, nil
 }
 
+// Remove 'ThreeScaleApicastRequestTime' alert
+func RemovePrometheusRulesMutator(existing, desired common.KubernetesObject) (bool, error) {
```
eguzki, line 35: I would move this method from `base_reconciler` to `pkg/reconcilers/prometheus_rules.go`

pkg/reconcilers/base_reconciler.go
```diff
@@ -31,6 +31,44 @@ func CreateOnlyMutator(existing, desired common.KubernetesObject) (bool, error)
...
+	existingPrometheusRule := existing.(*monitoringv1.PrometheusRule)
```
eguzki, line 36: The implementation is correct. HAve you considered using https://pkg.go.dev/golang.org/x/exp/slices ?

pkg/reconcilers/base_reconciler.go
eguzki: And add tests to it :)

pkg/3scale/amp/operator/apicast_reconciler.go
```diff
@@ -181,7 +181,7 @@ func (r *ApicastReconciler) Reconcile() (reconcile.Result, error) {
 		return reconcile.Result{}, err
 	}
 
-	err = r.ReconcilePrometheusRules(apicast.ApicastPrometheusRules(), reconcilers.CreateOnlyMutator)
+	err = r.ReconcilePrometheusRules(apicast.ApicastPrometheusRules(), reconcilers.UpdatePrometheusRulesMutator)
```
eguzki, line 184: Call `RemovePrometheusRulesMutator` mutator instead directly.

eguzki (changes requested):

eguzki:
Looking good. Implementation is correct. However I left some comments to make it "better" IMO :)

eguzki (approved):
Verification steps passed :ok:

austincunningham (approved):

PR #891, opened 2023-11-14
[WIP] THREESCALE-10110 - Investigation - Unit Tests - test Reconciler()

valerymo:
##  PR is NOT for Merge
_**For discussion only**_

Jira: https://issues.redhat.com/browse/THREESCALE-10110
Improve unit tests around the proxyConfigPromote controller

This PR was attempt to see how to test Reconciler()  method, main flow.
- The problem found - looks like need mock **inner functions** in Reconciler(), but this require changes in main code and not Unit tests (UT) only.
- The approach used in this PR - create alias vars for inner function(s).  Another approache, like "Interfaces", also require changes in main code to mock inner functions.
- If this is the only way to test main flow in some controllers,  - for existing project will be more reliable to check methods of controllers separately and not impact on main flow, that can be validated in **E2E tests**.
- In some cases, like for example `ProxyConfigPromoteStatusReconciler` - the changes in main code is not required (  already was implemented in ProxyConfigPromoteStatusReconciler Unit tests).

### DIFF
--- a/controllers/capabilities/proxyconfigpromote_controller.go
+++ b/controllers/capabilities/proxyconfigpromote_controller.go
@@ -38,6 +38,8 @@ type ProxyConfigPromoteReconciler struct {
 	*reconcilers.BaseReconciler
 }
 
+var proxyConfigPromoteReconcilerNew = (*ProxyConfigPromoteReconciler).proxyConfigPromoteReconciler
+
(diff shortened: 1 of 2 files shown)


MStokluska:
@valerymo I think we can close this and re-open whenever we are ready to proceed with this epic

MStokluska:
Closing for now, to be re-opened once we are ready to take on this task.

PR #893, opened 2023-11-15
THREESCALE-10387 add 4.14 as maxOpenShiftVersion to threescale 2.14

carlkyrillos:
JIRA: [THREESCALE-10387](https://issues.redhat.com/browse/THREESCALE-10387)

Adding 4.14 as maxOpenShiftVersion to 3scale-operator 2.14 stable.

### DIFF
--- a/bundle/manifests/3scale-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/3scale-operator.clusterserviceversion.yaml
@@ -205,6 +205,7 @@ metadata:
     containerImage: quay.io/3scale/3scale-operator:master
     createdAt: "2019-05-30T22:40:00Z"
     description: 3scale Operator to provision 3scale and publish/manage API
+    olm.properties: '[{"type": "olm.maxOpenShiftVersion", "value": "4.14"}]'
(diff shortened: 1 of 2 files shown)


eguzki:
Just heads up that this is missing `com.redhat.openshift.versions: v4.11-v4.14`. Tried here without luck https://github.com/3scale/3scale-operator/pull/877

With this, we still would need to add that label in midstream. The annotation would already be covered.

MStokluska:
> Just heads up that this is missing the label `com.redhat.openshift.versions: v4.11-v4.14`. Tried here without luck #877
> 
> With this, we still would need to add that label in midstream. The annotation would already be covered.

Thanks Eguzki, yes, that flag is already in midstream.

eguzki:
> > Just heads up that this is missing the label `com.redhat.openshift.versions: v4.11-v4.14`. Tried here without luck #877
> > With this, we still would need to add that label in midstream. The annotation would already be covered.
> 
> Thanks Eguzki, yes, that flag is already in midstream.

I know :) My point was we cannot remove that label from midstream, yet.

PR #897, opened 2023-11-23
deployment migration: handle not found of the new deployment

eguzki:


carlkyrillos:
@eguzki The changes look good to me thanks for catching this
/lgtm

PR #901, opened 2023-12-01
THREESCALE-10500 enable hpa for backend listener, worker and apicast production

austincunningham:
## What
THREESCALE-10500 enable hpa for backend listener, worker and apicast production

## Verification 
TBD
catchphrase here

image


pkg/reconcilers/deploymentconfig_test.go
```diff
@@ -1078,15 +1078,15 @@ func TestDeploymentConfigProbesMutator(t *testing.T) {
 		wantErr bool
 	}{
 		{
-            name: "Liveness Probe Updated",
-            args: args{
```
austincunningham, line 1082: IDE did this but it looks like the code was in the wrong place anyway. So will leave it

controllers/apps/apimanager_status_reconciler.go
```diff
@@ -42,9 +42,14 @@ func (s *APIManagerStatusReconciler) Reconcile() (reconcile.Result, error) {
 	if err != nil {
 		return reconcile.Result{}, fmt.Errorf("failed to calculate status: %w", err)
 	}
-
-	equalStatus := s.apimanagerResource.Status.Equals(newStatus, s.logger)
-	s.logger.V(1).Info("Status", "status is different", !equalStatus)
+	var equalStatus bool
```
MStokluska, line 45: although I can see a bit of room for improvement around this logic I'm struggling to understand why are we doing all these checks below? 

Why are these additional 2 checks required?

doc/operator-user-guide.md
```diff
@@ -735,6 +736,105 @@ spec:
         listenerSpec:
             priorityClassName: openshift-user-critical
 ```
+#### Setting Horizontal Pod Autoscaling 
+Horizontal Pod Autoscaling(HPA) is available for Apicast-production, Backend-listener and Backend-worker. The backend
+components require Redis running [async mode](https://github.com/3scale/apisonator/blob/master/docs/openshift_horizontal_scaling.md#async). 
+Async is enabled by default by the operator provide you aren't [logical Redis databases](https://github.com/3scale/apisonator/blob/master/docs/openshift_horizontal_scaling.md#redis-databases).
```
MStokluska, line 742:
```suggestion
Async is enabled by default by the operator provided you aren't using [logical Redis databases](https://github.com/3scale/apisonator/blob/master/docs/openshift_horizontal_scaling.md#redis-databases).
```

doc/operator-user-guide.md
```diff
@@ -735,6 +736,105 @@ spec:
...
+If you are not running in Async mode they you won't be able to enable HPA.
```
MStokluska, line 743:
```suggestion
If you are not running in Async mode you won't be able to enable HPA.
```

doc/operator-user-guide.md
```diff
@@ -735,6 +736,105 @@ spec:
...
+            enabled: true
+        workerSpec:
+          hpa:
+            enabled: true
+```
+Removing hpa field or setting enabled to false will remove the HPA for the component. 
+Once `enabled: true` you can manually edit the HPA instances that are created to optimize your [configuration](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/). 
+
+Alternatively you can set the HPA configuration in the APIManager CR. this will override the defaults. With this change 
+you are enabling sync with the values in the APIManager CR and it will be the owner of the HPA created. That is if you
```
MStokluska, line 802:
```suggestion
you are enabling syncing of the values between the APIManager CR and HPA CRs. APIManager will be the source of truth for the HPA created. That is if you 
```

pkg/3scale/amp/component/backend_hpa.go
```diff
@@ -0,0 +1,119 @@
...
+	// needs to change from dc to deployment
+	return &hpa.HorizontalPodAutoscaler{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      BackendWorkerName,
+			Namespace: namespace,
+		},
+		Spec: hpa.HorizontalPodAutoscalerSpec{
+			ScaleTargetRef: hpa.CrossVersionObjectReference{
+				Kind:       "DeploymentConfig",
+				Name:       BackendWorkerName,
```
MStokluska, line 85: IMO it would be better to have a single function for creating HPAs with addition of one extra field - Name. 
No need to be replicating the function 3 times just to change the name IMO.

pkg/3scale/amp/component/backend_hpa.go
```diff
@@ -0,0 +1,119 @@
+package component
+
+import (
+	appsv1alpha1 "github.com/3scale/3scale-operator/apis/apps/v1alpha1"
+	"github.com/3scale/3scale-operator/pkg/helper"
+	hpa "k8s.io/api/autoscaling/v2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (backend *Backend) BackendListenerHpa(namespace string, minPods *int32, maxPods int32, cpuPercent *int32, memoryPercent *int32) *hpa.HorizontalPodAutoscaler {
```
MStokluska, line 10: IMO it would be better to have a single function for creating HPAs with addition of one extra field - Name.
No need to be replicating the function 3 times just to change the name IMO.

pkg/3scale/amp/component/apicast_hpa.go
```diff
@@ -0,0 +1,60 @@
+package component
+
+import (
+	"github.com/3scale/3scale-operator/pkg/helper"
+	hpa "k8s.io/api/autoscaling/v2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (apicast *Apicast) ApicastProductionHpa(namespace string, minPods *int32, maxPods int32, cpuPercent *int32, memoryPercent *int32) *hpa.HorizontalPodAutoscaler {
```
MStokluska, line 9: IMO it would be better to have a single function for creating HPAs with addition of one extra field - Name.
No need to be replicating the function 3 times just to change the name IMO.

pkg/3scale/amp/component/backend_hpa.go
```diff
@@ -0,0 +1,119 @@
...
+							AverageUtilization: cpuPercent,
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func (backend *Backend) ApimanagerMutate(message string, cr *appsv1alpha1.APIManager) *appsv1alpha1.APIManager {
```
MStokluska, line 116: Can be removed since it doesn't seem to be used?

pkg/3scale/amp/operator/backend_reconciler.go
```diff
@@ -154,6 +154,60 @@ func (r *BackendReconciler) Reconcile() (reconcile.Result, error) {
...
+				if r.apiManager.Status.Conditions.IsTrueFor("Available") {
+					if r.apiManager.Status.Conditions[0].Message == "" {
+						r.apiManager.Status.Conditions[0].Message = message
+						r.UpdateResourceStatus(r.apiManager)
+					}
+				}
+			}
+
+		}
+	} else {
```
MStokluska, line 199: I think we should improve this a bit. 
Either now or in the future, we might want to log a message to APIM unrelated to HPA entirely. Are we at risk then that the message will be cleared every time by the code below if not running in HPA mode for backends?

pkg/3scale/amp/operator/backend_reconciler.go
```diff
@@ -154,6 +154,60 @@ func (r *BackendReconciler) Reconcile() (reconcile.Result, error) {
 	if err != nil {
 		return reconcile.Result{}, err
 	}
+	if r.apiManager.Spec.Backend.ListenerSpec.Hpa.Enabled || r.apiManager.Spec.Backend.WorkerSpec.Hpa.Enabled {
```
MStokluska, line 157: so basically this means that we can have apicast in non-hpa while backends in hpa but we can't have for example a worker in hpa but listener in non-hpa mode right?

MStokluska: also, if I want to delete my hpas - I set it to false, the deletion is triggered in the r.ReconcileHPA, correct me if this is incorrect but this means in the event where I want to have my apicast hpa on, but turn off the backend hpa - it won't work from this reconciler but the deletion of backend hpas would be more of a side effect of apicast reconciler going into the ReconcileHPA func.

controllers/apps/apimanager_status_reconciler.go
austincunningham: Orignally flow was
```go
equalStatus = s.apimanagerResource.Status.Equals(newStatus, s.logger)
s.logger.V(1).Info("Status", "status is different", !equalStatus)
```
was reverting the status update on every reconcile 

The first check is to check weather the current state of the apimanger is avaiable if so don't update again. 

The second check is to check if the current state of the status is available i.e. there is an edge case where if you delete a deployment  and it is restored you will have a case where newStatus is available but s.apimangerResouces is not.

pkg/3scale/amp/component/apicast_hpa.go
austincunningham: Think I went down this road and was hitting some issues.  I can give it another crack to see what the issue was .

pkg/3scale/amp/operator/backend_reconciler.go
austincunningham: agree had some taughts on it here https://issues.redhat.com/browse/THREESCALE-10563?focusedId=23610601&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-23610601

pkg/3scale/amp/component/backend_hpa.go
austincunningham: Dup will address comment in the first one

pkg/3scale/amp/component/backend_hpa.go
austincunningham: Dup will address comment in the first one

pkg/3scale/amp/operator/backend_reconciler.go
austincunningham: first point makes sense, I would need to check again before calling r.ReconcileHpa if I want the backend components to be individually enabled.

I think your second point also makes sense,  if both backend components are false then we never hit r.ReconcileHpa, and never hit the delete logic.  

Need to have a think about it.

pkg/3scale/amp/component/backend_hpa.go
austincunningham: I think so yes

controllers/apps/apimanager_status_reconciler.go
MStokluska: hmm, I think we might have a different bug somewhere then. I've reverted your changes locally and it works fine. Can you tell me how to exactly reproduce the issue with old configuration?

austincunningham: Enable hpa on backend with out logical db to get the status update you will notice that the message will get removed every reconcile. 

if you enable the first check on newStatus the status message doesn't get removed.

delete apicast production you will see that the status never corrects its self

set the second check for s.apimanagerResources, and you will see it correct it's self

MStokluska: I think we need to find a better way of adding an info status to the CRs in general or at least agree on that whether we want to have separate calls to update status of CR within the code. IMO any status updates should be done as part of the status reconcilers. One of the examples on how could we achieve this is in this test PR: https://github.com/3scale/3scale-operator/pull/905

Not a huge fan of it due to the fact that we are adding a lot of extra code just to cover some use cases. There's probably a better solution...

Just looking at how the backends reconciler is adding the status seems wrong to me since we are always trying to replace the condition[0]. Maybe we can sync after xmas break and discuss how to achieve info message status reporting.

pkg/3scale/amp/component/apicast_hpa.go
austincunningham: had a go at this and the backend delete wasn't functioning correctly. Probally related to the points raised here https://github.com/3scale/3scale-operator/pull/901#discussion_r1423794896, I think over all a single hpa function makes sense and I will see if I can resolve the deletion issue.

pkg/3scale/amp/operator/backend_reconciler.go
austincunningham: the check is not needed as we check if enabled in r.ReconcileHPA,   its only needed to set the status to nil so a seperate check on the status is needed.

pkg/3scale/amp/operator/backend_reconciler.go
austincunningham: removing status updates from the pr replace with log messages until we deside a way forward.

controllers/apps/apimanager_status_reconciler.go
austincunningham: removed this code

pkg/3scale/amp/component/backend_hpa.go
austincunningham: done

pkg/3scale/amp/component/apicast_hpa.go
austincunningham: done

pkg/3scale/amp/operator/backend_reconciler.go
austincunningham: done

pkg/3scale/amp/operator/backend_reconciler.go
austincunningham: done

doc/operator-user-guide.md
austincunningham: done

pkg/3scale/amp/operator/backend_reconciler.go
```diff
@@ -154,6 +154,30 @@ func (r *BackendReconciler) Reconcile() (reconcile.Result, error) {
...
+		}
+	} else {
+		// set status message if logical redis db are detected in the backend
+		if r.apiManager.Spec.Backend.ListenerSpec.Hpa || r.apiManager.Spec.Backend.WorkerSpec.Hpa {
+			message := "logical redis instances found in the backend, which is blocking redis async mode, horizontal pod autoscaling for backend cannot be enabled without async mode"
+			r.logger.Info(message)
+			// TODO update the status once the status update framework is in place.
+		}
+	}
+	if redisStorageUrl == redisQueuesUrl {
```
MStokluska, line 174: WDYT if we move the inside if to above "else" statement? Also, not sure if we even need to log this tbh

MStokluska:
/lgtm
Going to leave the decision re moving the "if block" to you.
All looks good with a couple of follow ups:
- resources change in another PR
- status message would be handy when hpa on backends is not possible

pkg/3scale/amp/operator/backend_reconciler.go
austincunningham: will leave it for now and drop it in 10499 as we handle the status update there

PR #903, opened 2023-12-07
THREESCALE-10238 - duplicated mapping rules initial handling

valerymo:


doc/operator-application-capabilities.md
```diff
@@ -558,6 +558,12 @@ spec:
 
 * **NOTE 1**: `httpMethod`, `pattern`, `increment` and `metricMethodRef` fields are required.
 * **NOTE 2**: `metricMethodRef` holds a reference to the existing metric or method map key name `system_name`. In the example, `hits`.
+* **NOTE 3**: The combination of `pattern + httpMethod + metricMethodRef` must be unique among all mapping rules.
+  - If Product CR contains mapping rules where `pattern and httpMethod` are same, even if `metricMethodRef` fields differ -
```
MStokluska, line 562: This should not be the case. If patter and httpMethod are same, but metric is different, the metric should not be a duplicated metric following 3scale.

MStokluska: From what it looks here:
```
		if desired.Pattern == existingRule.Pattern &&
			desired.HTTPMethod == existingRule.HTTPMethod &&
			desiredMetricID == existingRule.MetricID 
```
The above sentence isn't true?

controllers/capabilities/mapping_rules.go
```diff
@@ -182,15 +183,13 @@ func (t *ProductThreescaleReconciler) reconcileMappingRuleWithPosition(desired c
...
+	isValidRule, err := t.validateMappingRulesDuplication(desired, metricID)
 	if err != nil {
-		return fmt.Errorf("Error creating product [%s] mappingrule: %w", t.resource.Spec.SystemName, err)
+		return err
 	}
-
-	if metricID < 0 {
-		// Should not happen as metric and method references have been validated and should exists
-		return errors.New("product metric method ref for mapping rule not found")
+	if !isValidRule {
```
MStokluska, line 191: I'm trying to understand under what condition we would end up in this block.
The validateMappingRulesDuplication returns:
1. false and error - if there's an error in getting mapping rules - means we are not going to be here since we are going to exit above in the if err != nil
2. false and error if pattern is same, method is same and metric id is same - means we are not going to be here since we are going to exit above in the if err != nil
3. true and nil - we are not going to be here is isValidRule is true.

This checks from what it looks to me doesn't do anything?

doc/operator-application-capabilities.md
MStokluska, line 562: this is hard to follow, although we have no rules for what to call a duplicate what not (as of yet) we are saing that pattern method and metric ref must be unique, but then we are saying that when pattern and method are same this is duplication. 
I would much rather have something like, pattern, method and metric must be unique. Having same values across all 3 is a duplicated mapping rule. 
But maybe let's discuss this further when we have duplication pattern established.

valerymo: Thanks very for comment. It's updated, and hope more clear now. If required will be updated after duplication pattern established finally.
Sorry that forgot to create separate commit for review.

controllers/capabilities/mapping_rules.go
valerymo: Thank you for comment. 
- if error happens in `validateMappingRulesDuplication()` , it can happen [here](https://github.com/valerymo/3scale-operator/blob/THREESCALE-10238/controllers/capabilities/mapping_rules.go#L219-L222)  . In this case we can't say nothing about Duplication as we didn't check it. 
- from my pov - we need have these 2 checks to provide correct log. But I improved error message,  and hope it's better now. Done in separate commit. Thanks for comment.

valerymo:
@MStokluska , The current PR has been reworked and an updated solution is now provided. For me it works as expected. Thank you for review notes and discussion.

MStokluska:
After verification of the pr:
- the rules that are the same as existing ones do not get created for product, but do for backend
- the proxy is being updated after operator restart
- the position of rules matches the position in the CR

Is this what's expected?

valerymo:
> After verification of the pr:
> 
> * the rules that are the same as existing ones do not get created for product, but do for backend
> * the proxy is being updated after operator restart
> * the position of rules matches the position in the CR
> 
> Is this what's expected?

- **Test3** added  - it includes identical rules in CR. 
      - All rules do created for Product. 
      - No Integration configuration changes after operator restart.
      - Positions saved for current case.

valerymo:
@MStokluska , thank you for review, and sorry for inconvenience. I created integration branch for mapping rules issue, as we have several tasks opened for mapping rules. I'm closing current PR,  and will continue on https://github.com/valerymo/3scale-operator/pull/4 . Thanks

MStokluska:
@valerymo 
`- All rules do created for Product.` - yes, initially, but if you have existing product and add a new rule with same config, it won't get created
`- No Integration configuration changes after operator restart.` - on my end, it looks like every restart refreshes the config

PR #905, opened 2023-12-15
Test mappingrules reporting [WIP]

MStokluska:
Jira https://issues.redhat.com/browse/THREESCALE-10563

### DIFF
--- a/controllers/capabilities/backend_threescale_reconciler.go
+++ b/controllers/capabilities/backend_threescale_reconciler.go
@@ -41,7 +41,7 @@ func NewThreescaleReconciler(b *reconcilers.BaseReconciler,
 	}
 }
 
-func (t *BackendThreescaleReconciler) Reconcile() (*controllerhelper.BackendAPIEntity, error) {
+func (t *BackendThreescaleReconciler) Reconcile() (*controllerhelper.BackendAPIEntity, []string, error) {
 	taskRunner := helper.NewTaskRunner(nil, t.logger)
 	taskRunner.AddTask("SyncBackend", t.syncBackend)
 	// First methods and metrics, then mapping rules.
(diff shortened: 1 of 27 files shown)


controllers/capabilities/backend_threescale_reconciler.go
```diff
@@ -447,18 +450,32 @@ func (t *BackendThreescaleReconciler) reconcileMatchedMetrics(matchedMap map[str
...
 }
 
-func (t *BackendThreescaleReconciler) syncMappingRules(_ interface{}) error {
+func (t *BackendThreescaleReconciler) syncMappingRules(_ interface{}) (error, []string) {
+	var warnings []string
 	desiredKeys := make([]string, 0, len(t.backendResource.Spec.MappingRules))
 	desiredMap := map[string]capabilitiesv1beta1.MappingRuleSpec{}
 	for _, spec := range t.backendResource.Spec.MappingRules {
-		key := fmt.Sprintf("%s:%s", spec.HTTPMethod, spec.Pattern)
+		key := fmt.Sprintf("%s:%s:%s", spec.HTTPMethod, spec.Pattern, spec.MetricMethodRef)
```
valerymo, line 458: the difference with my approach (validation rules in Product) that you adding just first rule (from duplicates), and me - last. But main issue not this, but UI, and need decide how to handle it. We can remove duplication from Existing rules (that including alsso adding from UI), but seems need add some env var to allow user to decide about cleanup. It will allow control UI without changes in Porta (UI) and porta-go-clint. I think at first step - need just address the BUG, and next - extend the process of validation and control.

valerymo (commented):
comment added regaring approach of validation. As noted - main issue not this but how handle changes in UI. Suggested to do this work by steps. First one - address just BUG notes. We already did it in https://github.com/3scale/3scale-operator/pull/903 while added third parameter to key, and this fixed main issue of wrong Method name (used last method for ALL rules with same pattern). This issue fixed in [903](https://github.com/3scale/3scale-operator/pull/903) . This should be first PR/fix. Next steps - more improvements - handle UI, next - warnings.

PR #908, opened 2023-12-20
THREESCALE-10518 - Operator crash - OpenAPI CR doesn't create correct backend...

valerymo:
**Closing current PR, as commits looks not clear after squash / rebase process (probably mistake was done)**
**Please see new PR created with changes of this PR https://github.com/3scale/3scale-operator/pull/915**
**Could you please approve new PR** 

## WHAT
Jira: https://issues.redhat.com/browse/THREESCALE-10518
Operator crash - OpenAPI CR doesn't create correct backend usage section in Product CR

# Validation - Preparation
<details>
Prepare cluster (could be OSD)

## Install RHSSO 
Create project rhsso-test
1. Install RHSSO from **OperatorHub** into project rhsso-test
2. Create Keycloak Instance of RHSSO
3. Open RHSSO portal
4. Create a realm for the ***petstore*** product 
In RH User SSO web console:
  - Click **Add realm**. Add realm page will be opened
  - Set name, as `petstore`
  - Click Create. Petstore realm page will be opened. No changes required.

5. Create a client for 3scale  

- In RH User SSO web console:
  - Choose **Petstore** realm (in top-left corner)
  - Select Configure -> Client
  - Click Create
  - Set `Client ID`: **3scale-zync**
  - Click Save. Clients->3scale-zync page will be opened
  
- Client Settings:

3scale-zync Client setting will be as in the table below 

| Parameter          | Value              |
|-----------------------|--------------------|
|Name|   3scale-zync   |
|Client Protocol |    openid-connect   |
|Access Type | confidential |
|Standard Flow Enabled |Off |
|Direct Access Grants Enabled |Off |
|Service Accounts Enabled |On |

- Click Save
- Provide Realm-Management - **manage-clients**:
  - Enter `Service Account Roles` tab -> **Client Roles**
    - Choose `realm-management` -> `manage-clients`
    - Click Add selected

## Install 3scale 

1. Install and run 3scale operator

```
cd 3scale-operator
make install
oc new-project 3scale-test
make download
make run
```

2. Apply s3 secret
- Secret example (s3-creds-secret.yaml):
```
kind: Secret
apiVersion: v1
metadata: 
  name: s3-credentials
  namespace: 3scale-test
data: 
  AWS_ACCESS_KEY_ID: QU12345
  AWS_SECRET_ACCESS_KEY: aU12345=
  AWS_BUCKET: dm12345=
  AWS_REGION: ZX1234==
type: Opaque
```
```
oc apply -f s3-creds-secret.yaml
```

[...]

PR #915, opened 2024-01-14
THREESCALE-10518 - Operator crash - OpenAPI CR doesn't create correct backend…

valerymo:


MStokluska:
/lgmt based on https://github.com/3scale/3scale-operator/pull/908

PR #927, opened 2024-02-07
THREESCALE-10722 Refactored system-app Jobs to be recreated for each image change

carlkyrillos:
# Issue Link
JIRA: [THREESCALE-10722](https://issues.redhat.com/browse/THREESCALE-10722)

# What
This PR improves upon the existing PreHook and PostHook Jobs so that they are deleted and recreated whenever there is an image change.

# Verification Steps

1. Provision a cluster
2. Checkout this PR
3. Prepare to run the operator locally:
```bash
make install
make download
```
4. Create the Namespace, `s3-credentials` Secret, and the APIManager CR:
```bash
export NAMESPACE=3scale-test
oc new-project $NAMESPACE

cat << EOF | oc create -f -
kind: Secret
apiVersion: v1
metadata:
  name: s3-credentials
  namespace: $NAMESPACE
data:
  AWS_ACCESS_KEY_ID: c29tZXRoaW5nCg==
  AWS_BUCKET: c29tZXRoaW5nCg==
  AWS_REGION: dXMtd2VzdC0xCg==
  AWS_SECRET_ACCESS_KEY: c29tZXRoaW5nCg==
type: Opaque
EOF

cat << EOF | oc create -f -
kind: APIManager
apiVersion: apps.3scale.net/v1alpha1
metadata:
  name: 3scale
  namespace: $NAMESPACE
spec:
  wildcardDomain: $REPLACE_ME_WITH_WILDCARD_DOMAIN
  system:
    fileStorage:
      simpleStorageService:
        configurationSecretRef:
          name: s3-credentials
EOF
```
5. Run the operator locally:
```bash
make run
```
6. Wait for the installation to complete
```bash
oc get apimanager 3scale -n 3scale-test -ojson | jq '.status'
```
7. Update the system-app Deployment to trigger the Job deletion/re-creation:
```bash
oc patch deployment system-app -n 3scale-test --type=json -p='[{"op": "replace", "path": "/spec/template/spec/terminationGracePeriodSeconds", "value": 29}]'
```
8. Verify that the following occurs in order:
a. The old PreHook Job and the old PostHook Job are deleted
b. A new PreHook Job is created which runs to completion
c. Once the new `system-app` Pod is healthy (Ready = 3/3), the old `system-app` Pod is removed
d. A new PostHook Job is created which runs to completion

PR #923, opened 2024-01-26
THREESCALE-10698 Remove remaining DeploymentConfig references

carlkyrillos:
# Issue Link
JIRA: [THREESCALE-10698](https://issues.redhat.com/browse/THREESCALE-10698)

# What
This PR removes or updates several references to DeploymentConfigs in the operator code, docs, and CSV/CRDs. Note that not all references to DeploymentConfigs can be removed at this time because the operator needs to be able to `get` and `delete` them as part of the migration for the upgrade path. These last few references to DeploymentConfigs will need to be removed in a future release (likely 2.16).

# Verification Steps
The main verification for this PR is confirming that the GrafanaDashboards still work on an upgrade path from the master branch as this will also verify the overall functionality of 3scale-operator.

1. Provision a cluster
2. Checkout the master branch
3. Prepare to run the operator locally:
```bash
make install
make download
```
4. Create the Namespace, `s3-credentials` Secret, and the APIManager CR (with .spec.monitoring.enabled = true):
```bash
export NAMESPACE=3scale-test
oc new-project $NAMESPACE

cat << EOF | oc create -f -
kind: Secret
apiVersion: v1
metadata:
  name: s3-credentials
  namespace: $NAMESPACE
data:
  AWS_ACCESS_KEY_ID: c29tZXRoaW5nCg==
  AWS_BUCKET: c29tZXRoaW5nCg==
  AWS_REGION: dXMtd2VzdC0xCg==
  AWS_SECRET_ACCESS_KEY: c29tZXRoaW5nCg==
type: Opaque
EOF

cat << EOF | oc create -f -
kind: APIManager
apiVersion: apps.3scale.net/v1alpha1
metadata:
  name: 3scale
  namespace: $NAMESPACE
spec:
  monitoring:
    enabled: true
  wildcardDomain: $REPLACE_ME_WITH_WILDCARD_DOMAIN
  system:
    fileStorage:
      simpleStorageService:
        configurationSecretRef:
          name: s3-credentials
EOF
```
5. Install the prometheus-operator (Red Hat version) in the `3scale-test` Namespace using the default Update channel (`beta`) and Version(`4.10.0`)
6. Wait for the prometheus-operator to finish installing and then create a Prometheus CR and the required RBAC:
```yaml
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: 3scale-prometheus
  namespace: 3scale-test
---
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
[...]

PR #894, opened 2023-11-16
THREESCALE-10388 Migrate apicast DeploymentConfigs to Deployments

carlkyrillos:
# Issue link
JIRA: [THREESCALE-10388](https://issues.redhat.com/browse/THREESCALE-10388)

# What
This is the first PR against the new feature branch `threescale-10163` that will be used while implementing the changes needed to migrate each of 3scale-operator's DeploymentConfigs to Deployments - see [epic](https://issues.redhat.com/browse/THREESCALE-10163).

This PR replaces the `apicast-staging` and `apicast-production` DeploymentConfigs with Deployments. The PR also provides the upgrade scaffolding that will be used by the remaining DeploymentConfigs.

# Verification steps

1. Provision a cluster
2. Install 3scale locally using master
3. Checkout this PR
4. Re-run 3scale locally
5. Verify that the new `apicast-staging` and `apicast-production` Deployments are created
6. Verify that the `apicast-staging` and `apicast-production` DeploymentConfigs are deleted once the Deployments report healthy
7. Login to admin portal and create dummy product/backend.
8. Promote the APIcast configuration to Staging APIcast and curl the staging endpoint with a valid user_key to confirm it's working
9. Promote the APIcast configuration to Production APIcast and curl the production endpoint with a valid user_key to confirm it's working

PR #929, opened 2024-02-15
THREESCALE-10163 Migrate from DeploymentConfigs to Deployments

carlkyrillos:
# Issue Link
JIRA: [THREESCALE-10163](https://issues.redhat.com/browse/THREESCALE-10163)

# What
This PR merges the feature branch into master and includes all of the changes needed to run the 3scale-operator with Deployment objects instead of DeploymentConfigs as they are now deprecated. These changes accommodate for both fresh install and upgrade scenarios.

# Verification Steps
## Prerequisites
*  OSD v4.14 cluster
* `oc` CLI
* `ocm` CLI
* `jq` CLI

## Fresh Install Scenario
1. Checkout this PR
2. Prepare to run the operator locally:
```bash
make install
make download
```
3. Create the Namespace and `s3-credentials` Secret:
```bash
export NAMESPACE=3scale-test
oc new-project $NAMESPACE

cat << EOF | oc create -f -
kind: Secret
apiVersion: v1
metadata:
  name: s3-credentials
  namespace: $NAMESPACE
data:
  AWS_BUCKET: c29tZXRoaW5nCg==
  AWS_REGION: dXMtd2VzdC0xCg==
type: Opaque
EOF
```
4. Create the APIManager CR (make sure to replace `<REPLACE_WITH_CLUSTER_NAME>` with the name of your cluster):
```bash
export CLUSTER_NAME=<REPLACE_WITH_CLUSTER_NAME>
```
```bash
export WILDCARD_DOMAIN=$(ocm get /api/clusters_mgmt/v1/clusters --parameter search="name is '$CLUSTER_NAME'" | jq -r '.items[].console.url' | cut -d'.' -f2-)

cat << EOF | oc create -f -
kind: APIManager
metadata:
  name: 3scale
  namespace: $NAMESPACE
spec:
  wildcardDomain: $WILDCARD_DOMAIN
  monitoring:
    enabled: true
  system:
    fileStorage:
      simpleStorageService:
        configurationSecretRef:
          name: s3-credentials
EOF
```
5. Run the operator locally:
```bash
make run
```
6. Wait for the installation to complete
```bash
oc get apimanager 3scale -n 3scale-test -ojson | jq '.status'
```
7. Complete each step in the `Things to Confirm` section below
8. When finished, delete the test namespace to prepare for the upgrade scenario:
```bash
oc delete project 3scale-test
```

## Upgrade Scenario

1. Checkout master branch 
2. Prepare to run the operator locally:
```bash
make install
make download
```
3. Create the Namespace and `s3-credentials` Secret:
```bash
export NAMESPACE=3scale-test
oc new-project $NAMESPACE

cat << EOF | oc create -f -
kind: Secret
apiVersion: v1
metadata:
  name: s3-credentials
  namespace: $NAMESPACE
data:
  AWS_BUCKET: c29tZXRoaW5nCg==
  AWS_REGION: dXMtd2VzdC0xCg==
type: Opaque
EOF
```
4. Create the APIManager CR (make sure to replace `<REPLACE_WITH_CLUSTER_NAME>` with the name of your cluster):
```bash
export CLUSTER_NAME=<REPLACE_WITH_CLUSTER_NAME>
```
```bash
export WILDCARD_DOMAIN=$(ocm get /api/clusters_mgmt/v1/clusters --parameter search="name is '$CLUSTER_NAME'" | jq -r '.items[].console.url' | cut -d'.' -f2-)

cat << EOF | oc create -f -
kind: APIManager
metadata:
  name: 3scale
  namespace: $NAMESPACE
spec:
  wildcardDomain: $WILDCARD_DOMAIN
  monitoring:
    enabled: true
  system:
    fileStorage:
      simpleStorageService:
        configurationSecretRef:
          name: s3-credentials
EOF
```
5. Run the operator locally:
```bash
make run
```
6. Wait for the installation to complete
```bash
oc get apimanager 3scale -n 3scale-test -ojson | jq '.status'
```
7. Checkout this PR
8. Re-run the operator locally:
```bash
make run
```
9. Wait for the upgrade to complete
```bash
oc get apimanager 3scale -n 3scale-test -ojson | jq '.status'
```
10. Complete each step in the `Things to Confirm` section below

## Things to Confirm

* Verify there are no DeploymentConfig objects:
```bash
oc get deploymentconfigs -n 3scale-test
```
* Complete steps 7-9 from #894 to verify the apicast Deployments are working
* Complete steps 5-9 from #923 to verify the Dashboards are working
    - **NOTE:** For step 9 you can also just expose the routes with `oc expose service prometheus-operated --hostname prometheus.3scale-test.$WILDCARD_DOMAIN` and `oc expose service grafana-service --hostname grafana.3scale-test.$WILDCARD_DOMAIN` to access the UIs.
    - **NOTE:** Not all of the panels on each dashboard will be working but they should match what is working on the master branch.
* Complete steps 7-8 from #927 to verify the system-app PreHook and PostHook Jobs are working as expected


pkg/3scale/amp/operator/redis_reconciler.go
```diff
@@ -60,25 +60,39 @@ func (r *RedisReconciler) Reconcile() (reconcile.Result, error) {
...
-		reconcilers.DeploymentConfigImageChangeTriggerMutator,
-		reconcilers.DeploymentConfigContainerResourcesMutator,
-		reconcilers.DeploymentConfigAffinityMutator,
-		reconcilers.DeploymentConfigTolerationsMutator,
-		reconcilers.DeploymentConfigPodTemplateLabelsMutator,
-		reconcilers.DeploymentConfigPriorityClassMutator,
-		reconcilers.DeploymentConfigTopologySpreadConstraintsMutator,
-		reconcilers.DeploymentConfigPodTemplateAnnotationsMutator,
-		// 3scale 2.13 -> 2.14
-		upgrade.Redis6CommandArgsEnv,
```
MStokluska, line 73: Q:
What's the impact on 2.13 > 2.15 if this is removed? Not that we support multi-minor upgrades but just curious.

pkg/3scale/amp/operator/redis_reconciler.go
```diff
@@ -60,25 +60,39 @@ func (r *RedisReconciler) Reconcile() (reconcile.Result, error) {
...
-	err = r.ReconcileDeploymentConfig(r.DeploymentConfig(redis), dcMutator)
+	redisDeployment := r.Deployment(redis)
+	err = r.ReconcileDeployment(redisDeployment, deploymentMutator)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
 
-	// redis Service
-	err = r.ReconcileService(r.Service(redis), reconcilers.CreateOnlyMutator)
+	// 3scale 2.14 -> 2.15
```
MStokluska, line 79: Q
Out of curiosity, if I bump from 2.13 to 2.15, it should still migrate correctly right?

pkg/3scale/amp/operator/zync_reconciler.go
```diff
@@ -44,69 +50,102 @@ func (r *ZyncReconciler) Reconcile() (reconcile.Result, error) {
...
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		if !isMigrated {
+			return reconcile.Result{Requeue: true}, nil
+		}
+
 		// Zync DB Service
-		err = r.ReconcileService(zync.DatabaseService(), reconcilers.CreateOnlyMutator)
+		err = r.ReconcileService(zync.DatabaseService(), reconcilers.ServiceMutator(serviceMutators...))
```
MStokluska, line 148: Q
Similar to one of the above questions, create only vs expected/desired reconciliation. Why can/must we do this?

pkg/reconcilers/service.go
```diff
@@ -8,6 +8,32 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
+func ServiceMutator(opts ...MutateFn) MutateFn {
+	return func(existingObj, desiredObj common.KubernetesObject) (bool, error) {
```
MStokluska, line 12: Q
Why the need for these? Is this something related to move to deployments?

pkg/3scale/amp/operator/system_reconciler.go
```diff
@@ -182,6 +143,134 @@ func (r *SystemReconciler) Reconcile() (reconcile.Result, error) {
...
+
+	// Used to synchronize the system-app Deployment with the PreHook/PostHook Jobs
+	currentAppDeploymentGeneration, err := getSystemAppDeploymentGeneration(r.apiManager.GetNamespace(), r.Client())
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// SystemApp PreHook Job
+	preHookJob := system.AppPreHookJob(ampImages.Options.SystemImage, currentAppDeploymentGeneration)
+	err = r.ReconcileJob(preHookJob, reconcilers.CreateOnlyMutator)
```
MStokluska, line 179: So during 2.14 > 2.15, we will run the prehook twice right? 
First, on initial when we hit this line of code (since no jobs are present)
Second, after updating the deployment via deployment mutators?

pkg/3scale/amp/operator/system_reconciler.go
```diff
@@ -182,6 +143,134 @@ func (r *SystemReconciler) Reconcile() (reconcile.Result, error) {
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// Block reconciling system-app Deployment until PreHook Job has completed
+	if !helper.HasJobCompleted(preHookJob.Name, preHookJob.Namespace, r.Client()) {
+		systemComponentNotReady = true
```
MStokluska, line 186: Isn't it better to just cancel here instead of continuing with check before applying mutators? 
So if job has no completed yet, leave and come back later.

pkg/3scale/amp/operator/system_reconciler.go
```diff
@@ -182,6 +143,134 @@ func (r *SystemReconciler) Reconcile() (reconcile.Result, error) {
...
+		reconcilers.DeploymentTopologySpreadConstraintsMutator,
+		reconcilers.DeploymentPodTemplateAnnotationsMutator,
+		r.systemAppDeploymentResourceMutator,
+		reconcilers.DeploymentRemoveDuplicateEnvVarMutator,
+		reconcilers.DeploymentPodContainerImageMutator,
+	}
+	if r.apiManager.Spec.System.AppSpec.Replicas != nil {
+		systemAppDeploymentMutators = append(systemAppDeploymentMutators, reconcilers.DeploymentReplicasMutator)
+	}
+	if !systemComponentNotReady {
```
MStokluska, line 204: I think if we drop the reconciler above if the job hasn't been completed yet, this is not required. But might be missing something

pkg/3scale/amp/operator/system_reconciler.go
```diff
@@ -182,6 +143,134 @@ func (r *SystemReconciler) Reconcile() (reconcile.Result, error) {
...
+	}, deployment)
+	if err != nil && !k8serr.IsNotFound(err) {
+		return reconcile.Result{}, err
+	}
+	if k8serr.IsNotFound(err) || !helper.IsDeploymentAvailable(deployment) || helper.IsDeploymentProgressing(deployment) || !helper.HasJobCompleted(preHookJob.Name, preHookJob.Namespace, r.Client()) {
+		systemComponentNotReady = true
+	}
+
+	// SystemApp PostHook Job
+	if !systemComponentNotReady {
```
MStokluska, line 225: WDYT about systemComponentReady instead? 
the !systemComponentNotReady is hard to read, but no strong opinion, might be just me...

pkg/3scale/amp/operator/system_reconciler.go
```diff
@@ -182,6 +143,134 @@ func (r *SystemReconciler) Reconcile() (reconcile.Result, error) {
...
+		err = r.ReconcileJob(system.AppPostHookJob(ampImages.Options.SystemImage, currentAppDeploymentGeneration), reconcilers.CreateOnlyMutator)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+
+	// 3scale 2.14 -> 2.15
+	isMigrated, err := upgrade.MigrateDeploymentConfigToDeployment(component.SystemAppDeploymentName, r.apiManager.GetNamespace(), false, r.Client(), nil)
```
MStokluska, line 233: Q
Can you describe the flow of 2.14 > 2.15 upgrade please. 
In my head:
- create job and run it
- mutators = we should get deployment at this stage
- migrate = remove the dc
- posthook?

MStokluska:
Just a couple of questions left for clarification, overall looks good :+1:  Good Job!

pkg/3scale/amp/operator/redis_reconciler.go
carlkyrillos: I'll be honest I'm not sure what the impact would be I haven't tested it but I can. My understanding was that we don't support multi-minor upgrades like you said and so it could be safely removed after 2.14

pkg/3scale/amp/operator/redis_reconciler.go
carlkyrillos: Again I'm not sure what the behavior would be, I thought that we only needed to support 2.13 -> 2.14 and then 2.14 -> 2.15. But if that's not the case then I think this change (as well as some others) will need to be reverted.

That being said, the actual migration logic for this PR (DeploymentConfigs to Deployments) should still work and migrate correctly. But I'm not sure about the Redis6 stuff that was required for 2.13 -> 2.14.

pkg/3scale/amp/operator/zync_reconciler.go
```diff
@@ -21,13 +22,18 @@ func NewZyncReconciler(baseAPIManagerLogicReconciler *BaseAPIManagerLogicReconci
...
+	}
+
 	zync, err := Zync(r.apiManager, r.Client())
 	if err != nil {
 		return reconcile.Result{}, err
 	}
 
 	// Zync Que Role
-	err = r.ReconcileRole(zync.QueRole(), reconcilers.CreateOnlyMutator)
+	err = r.ReconcileRole(zync.QueRole(), reconcilers.RoleRuleMutator)
```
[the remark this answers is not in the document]
carlkyrillos: You're correct that we used to only create the Role and nothing more. However that Role's rule used to be for [DeploymentConfigs](https://github.com/3scale/3scale-operator/blame/de98237e57d7a68a4f44c5a27edabeebd8184136/pkg/3scale/amp/component/zync.go#L112) but it now needs to apply for [Deployments](https://github.com/3scale/3scale-operator/blob/845cddc4d9cc1405c6f10b65f50c727063e438d5/pkg/3scale/amp/component/zync.go#L113).

I thought it would be better to update the Role rather than delete it and create a new one which is why we need the new mutator to update the RBAC.

pkg/3scale/amp/operator/zync_reconciler.go
carlkyrillos: Same as above, the Service selector used to reference [DeploymentConfigs](https://github.com/3scale/3scale-operator/blame/de98237e57d7a68a4f44c5a27edabeebd8184136/pkg/3scale/amp/component/zync.go#L537) but it now needs to reference [Deployments](https://github.com/3scale/3scale-operator/blob/845cddc4d9cc1405c6f10b65f50c727063e438d5/pkg/3scale/amp/component/zync.go#L493).

pkg/reconcilers/service.go
carlkyrillos: Yep this is related to move to Deployments. See this [comment](https://github.com/3scale/3scale-operator/pull/929#discussion_r1502910357) for more context.

pkg/3scale/amp/operator/system_reconciler.go
carlkyrillos: /No the PreHook will only run once (on initial run since Job doesn't exist). Subsequent reconciles won't recreate/run the Job since it already exists. The exception is if the `system-app` Deployment's generation changes which will cause the PreHook Job to be deleted and then recreated.

pkg/3scale/amp/operator/system_reconciler.go
carlkyrillos: We could do this but that would unnecessarily delay the rollout of the `system-sidekiq` Deployment which will in turn make the install/reconcile time slightly longer.

pkg/3scale/amp/operator/system_reconciler.go
carlkyrillos: 1. Create PreHook Job and run it
2. Once the PreHook Job has completed, create the `system-app` Deployment
3. Once the `system-app` Deployment is ready, create the PostHook Job and run it
4. Migrate the DeploymentConfig to Deployment (i.e. delete the old DeploymentConfig now that the Deployment is healthy)

pkg/3scale/amp/operator/system_reconciler.go
carlkyrillos: I'm not sure I follow. Correct me if I'm wrong but I think you're getting at moving the `system-app` Deployment mutators to the `if` block so we aren't creating them unnecessarily if the PreHook Job isn't ready yet.

carlkyrillos: There's no harm in moving the mutators to inside the `if` block so I included that in the latest push. Let me know if I misunderstood the original question though

pkg/3scale/amp/operator/system_reconciler.go
carlkyrillos: I agree its confusing, I just pushed changes to use `systemComponentsReady` instead.

pkg/3scale/amp/operator/redis_reconciler.go
MStokluska: I would keep this in then if it's not an issue

pkg/3scale/amp/operator/redis_reconciler.go
MStokluska: same as above if we can

pkg/3scale/amp/operator/zync_reconciler.go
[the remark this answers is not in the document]
MStokluska: thanks

pkg/3scale/amp/operator/system_reconciler.go
MStokluska: ah yes, I see, so we are returning "1" generation in AppPreHookJob on initial reconcile in upgrade scenario but then create the deployment - with generation 1. So the second job doesn't trigger because the generation is the same. Thanks

pkg/3scale/amp/operator/redis_reconciler.go
MStokluska: actually, after a call with Carl, we are dropping this.

pkg/3scale/amp/operator/redis_reconciler.go
MStokluska: update same as above

controllers/apps/apimanager_status_reconciler.go
```diff
@@ -55,6 +56,10 @@ func (s *APIManagerStatusReconciler) Reconcile() (reconcile.Result, error) {
 	if equalStatus && apiManagerAvailable {
 		// Steady state
 		s.logger.V(1).Info("Status was not updated")
+
+		// 3scale 2.14 -> 2.15
+		err = upgrade.DeleteImageStreams(s.apimanagerResource.Namespace, s.Client())
```
MStokluska, line 61: hey @carlkyrillos why did you placed it here in status reconciler?

WDYT if we move it https://github.com/3scale/3scale-operator/pull/929/files#diff-6ebc088c3986da1f5f483589ce7d214c2047c71ce02c8a7c697bc216563398e7R22

Or to the apimanager reconciler (this can be the last step in apim reconciliation however, I'm in favor of moving the migration logic to migration function)

controllers/capabilities/proxyconfigpromote_status_reconciler_test.go
```diff
@@ -39,11 +39,10 @@ func getApiManger() (apimanager *appsv1alpha1.APIManager) {
 		},
 		Spec: appsv1alpha1.APIManagerSpec{
 			APIManagerCommonSpec: appsv1alpha1.APIManagerCommonSpec{
-				AppLabel:                     &appLabel,
-				ImageStreamTagImportInsecure: &trueValue,
```
MStokluska, line 43: I am wondering if there's anything in deployment that we could use to replace this?
Any thoughts?

controllers/apps/apimanager_status_reconciler.go
carlkyrillos: Good call, I moved it to the APIManager controller

controllers/capabilities/proxyconfigpromote_status_reconciler_test.go
carlkyrillos: We decided to remove the `ImageStreamTagImportInsecure` field from the APIManager .spec because Deployments don't expose a way to bypass TLS cert verification out of the box. There are workarounds to skip TLS server cert verification at both the [OpenShift level](https://docs.openshift.com/container-platform/4.14/openshift_images/image-configuration.html#images-configuration-insecure_image-configuration) and the [kubernetes level](https://stackoverflow.com/questions/51893120/kubernetes-pull-from-insecure-docker-registry). We will include these workarounds in docs/release notes.

carlkyrillos:
@MStokluska FYI the e2e tests are failing due to an issue with prow but `make test-e2e` is passing locally.

MStokluska:
/lgtm
/approve

Happy to for e2e to be bypassed. Unrelated to the PR

MStokluska (approved):

PR #931, opened 2024-02-22
THREESCALE-10792 The operator doesn't delete the objects in the correct order

valerymo:
Jira: https://issues.redhat.com/browse/THREESCALE-10792

### DIFF
--- a/controllers/capabilities/methods.go
+++ b/controllers/capabilities/methods.go
@@ -23,7 +23,7 @@ func (t *ProductThreescaleReconciler) syncMethods(_ interface{}) error {
 	existingMap := map[string]threescaleapi.MethodItem{}
 	existingList, err := t.productEntity.Methods()
 	if err != nil {
-		return fmt.Errorf("Error sync product methods [%s]: %w", t.resource.Spec.SystemName, err)
+		return fmt.Errorf("Error sync product methods (methods) [%s]: %w", t.resource.Spec.SystemName, err)
 	}
 
 	existingKeys := make([]string, 0, len(existingList.Methods))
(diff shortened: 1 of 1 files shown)


controllers/capabilities/methods.go
```diff
@@ -105,10 +105,18 @@ func (t *ProductThreescaleReconciler) createNewMethods(desiredNewMap map[string]
 
 func (t *ProductThreescaleReconciler) processNotDesiredMethods(notDesiredMap map[string]threescaleapi.MethodItem) error {
 	for _, notDesiredMethod := range notDesiredMap {
-		err := t.productEntity.DeleteMethod(notDesiredMethod.ID)
+		methodNotInUse, err := t.isMethodNotInUse(notDesiredMethod.ID)
 		if err != nil {
 			return err
 		}
+		if methodNotInUse {
```
MStokluska, line 112: I don't think we need this Valery. 
WDYT if all we would do is to capture this specific error and make it more understandable?

valerymo:
after more checking, it looks like these changes are not needed
will close PR.

controllers/capabilities/methods.go
valerymo: @MStokluska , I closed PR, as checked more, and agree with your notes in Task.
Regarding warning/error  to be more clear - also looks like it's clear enough what happens

```
$ oc describe product product2
..........
......
Status:
  Conditions:
    Last Transition Time:  2024-02-26T12:40:48Z
    Message:               Task failed SyncMethods: Error sync product methods, process not DesiredMethods [operatedproduct2]: product [operatedproduct2] delete method: error calling 3scale system - reason: {"errors":{"base":["Method is used by the latest gateway configuration and cannot be deleted"]}} - code: 403

```

PR #957, opened 2024-03-28
THREESCALE-10209 Remove support for Redis namespace

valerymo:
Jira: https://issues.redhat.com/browse/THREESCALE-10209

**Remove support for Redis namespace in 3scale-operator**

In version 2.15, support for Redis namespaces in the system (porta) is planned to be discontinued, and the operator must reflect this change.

**Validation. Initial installation of PR/THREESCALE-10209 branch**

<details>

- Image:  quay.io/vmogilev_rhmi/3scale-index:0.0.11
- Installation succedded.   `NAMESPACE` data field is not present in `system-redis` secret and Redis is working.
- Please see logs below

```
$ oc describe catalogsource threescale-operators --namespace=openshift-marketplace  |grep -i image
  Image:        quay.io/vmogilev_rhmi/3scale-index:0.0.11


$ oc get secret system-redis -oyaml
apiVersion: v1
data:
  SENTINEL_HOSTS: ""
  SENTINEL_ROLE: ""
  URL: xxxxxxx
kind: Secret
metadata:
  creationTimestamp: "2024-04-01T08:47:20Z"
  labels:
    app: 3scale-api-management
    threescale_component: system
  name: system-redis
  namespace: 3scale-test
...
type: Opaque


$ oc get deploy
NAME                                        READY   UP-TO-DATE   AVAILABLE   AGE
apicast-production                          1/1     1            1           8m47s
apicast-staging                             1/1     1            1           8m47s
backend-cron                                1/1     1            1           9m26s
backend-listener                            1/1     1            1           9m26s
backend-redis                               1/1     1            1           9m26s
backend-worker                              1/1     1            1           9m26s
system-app                                  1/1     1            1           8m22s
system-memcache                             1/1     1            1           9m25s
system-mysql                                1/1     1            1           9m26s
system-redis                                1/1     1            1           9m26s
system-searchd                              1/1     1            1           9m25s
system-sidekiq                              1/1     1            1           8m47s
threescale-operator-controller-manager-v2   1/1     1            1           18m
zync                                        1/1     1            1           8m47s
zync-database                               1/1     1            1           8m47s
zync-que                                    1/1     1            1           8m47s

$ oc get pod |grep redis
backend-redis-85b866f7c7-52gv8                              1/1     Running     0               10m
system-redis-6b765d75c9-f4q4s                               1/1     Running     0               10m

```

</details>


pkg/3scale/amp/operator/redis_reconciler.go
```diff
@@ -72,6 +76,25 @@ func (r *RedisReconciler) Reconcile() (reconcile.Result, error) {
 		return reconcile.Result{}, err
 	}
 
+	// 3scale 2.14 -> 2.15 Upgrade
+	// delete NAMESPACE attribute from secret system-redis
+	// and delete REDIS_NAMESPACE env var from deployments system-app and system-sidekiq
+	res, err := r.deleteSystemRedisSecretNamespaceAttribute()
```
MStokluska, line 82: @valerymo do you know if this happens before or after migrating to deployments?

MStokluska: happens before the migration, but when deployment is already up. Which means that the deployment (line 90) won't have the env anymore anyway due to the other changes in this PR.
Meaning that the deleteDeployment envar is not necessary but I think best would be to coordinate with @carlkyrillos on this

MStokluska: Alternative is, if @carlkyrillos won't filter out the env, is that you could change your removal of env to target deploymentConfiguration instead, since at this point. The dcs are up, the deployment is up, the migration has not started. 
If you remove env from the dc and Carl migrates all envs, it should be ok. But IMO, it's an unnecessary step

valerymo: @MStokluska , we talked with you, when you helped to find solution how to implement it, and seems we wanted to place Before. But following your question --  seems to me - to update Secret (data)- it maybe doesn't matter (?), but for deployments env vars update - or after Migration , when we already have ddeployments (maybe need wait ...) or in time of Migration.  Thanks for comment

pkg/3scale/amp/operator/redis_reconciler.go
```diff
@@ -124,3 +147,100 @@ func Redis(apimanager *appsv1alpha1.APIManager, client client.Client) (*componen
 	}
 	return component.NewRedis(opts), nil
 }
+
+func (r *RedisReconciler) deleteSystemRedisSecretNamespaceAttribute() (reconcile.Result, error) {
+	redis, err := Redis(r.apiManager, r.Client())
```
MStokluska, line 152: No harm in this but I don't think we need it. We just want the access to secret.

pkg/3scale/amp/operator/redis_reconciler.go
```diff
@@ -124,3 +147,100 @@ func Redis(apimanager *appsv1alpha1.APIManager, client client.Client) (*componen
 	}
 	return component.NewRedis(opts), nil
 }
+
+func (r *RedisReconciler) deleteSystemRedisSecretNamespaceAttribute() (reconcile.Result, error) {
```
MStokluska, line 151: No point in returning `reconcile.Result` if the only possible result is {}

pkg/3scale/amp/operator/redis_reconciler.go
```diff
@@ -124,3 +147,100 @@ func Redis(apimanager *appsv1alpha1.APIManager, client client.Client) (*componen
...
+	redis, err := Redis(r.apiManager, r.Client())
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	desired := redis.SystemRedisSecret()
```
MStokluska, line 157: In general, I think it's better to fetch secret directly. We know the secret name (system-redis) and namespace (from apim)

pkg/3scale/amp/operator/redis_reconciler.go
```diff
@@ -124,3 +147,100 @@ func Redis(apimanager *appsv1alpha1.APIManager, client client.Client) (*componen
...
+	existing := &corev1.Secret{}
+	err = r.Client().Get(context.TODO(), types.NamespacedName{Name: desired.Name, Namespace: r.apiManager.Namespace}, existing)
+	if k8serr.IsNotFound(err) {
```
MStokluska, line 160: I don't think we can ever land here, is it fair to say that at this point, we are pretty sure we are not going to get "isNotFound" error since it would have failed earlier? Can you confirm this is really required?

pkg/3scale/amp/operator/redis_reconciler.go
```diff
@@ -124,3 +147,100 @@ func Redis(apimanager *appsv1alpha1.APIManager, client client.Client) (*componen
...
+		return reconcile.Result{}, err
+	}
+
+	update := false
+	namespaceAttr := "NAMESPACE"
+	if _, ok := existing.Data[namespaceAttr]; ok {
+		update = true
+		delete(existing.Data, namespaceAttr)
+	}
+	if update {
```
MStokluska, line 173: I think we don't need `update` var here, if it's existing > delete > update.

pkg/3scale/amp/operator/redis_reconciler.go
```diff
@@ -124,3 +147,100 @@ func Redis(apimanager *appsv1alpha1.APIManager, client client.Client) (*componen
...
+	err = r.Client().Get(context.TODO(), types.NamespacedName{Name: desired.Name, Namespace: r.apiManager.Namespace}, existing)
+	if k8serr.IsNotFound(err) {
+		return reconcile.Result{}, nil
+	}
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	update := false
+	namespaceAttr := "NAMESPACE"
```
MStokluska, line 168: small thing but this is a key, not an attribute - attributes are characteristics or properties of an object

pkg/3scale/amp/operator/redis_reconciler.go
MStokluska: As agreed, please move the secret key removal to the post migration step

pkg/3scale/amp/operator/redis_reconciler.go
```diff
@@ -72,6 +76,25 @@ func (r *RedisReconciler) Reconcile() (reconcile.Result, error) {
...
 	}
 
+	// 3scale 2.14 -> 2.15 Upgrade
+	// delete NAMESPACE attribute from secret system-redis
+	// and delete REDIS_NAMESPACE env var from deployments system-app and system-sidekiq
+	res, err := r.deleteSystemRedisSecretNamespaceAttribute()
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	if res.Requeue {
```
MStokluska, line 86: Requeue is always nil, no need for handling it since we will never get here.

pkg/3scale/amp/operator/redis_reconciler.go
MStokluska: actually, that Redis is already initialized on line 58

pkg/3scale/amp/operator/redis_reconciler.go
valerymo: Done. Thank you

pkg/3scale/amp/operator/redis_reconciler.go
valerymo: I had the problem, so I placed it. I can retest and confirm if required.

pkg/3scale/amp/operator/redis_reconciler.go
valerymo: removed. Thank you

pkg/3scale/amp/operator/redis_reconciler.go
valerymo: Done. Thank you

pkg/3scale/amp/operator/redis_reconciler.go
valerymo: ReChecked. Not removing without this. Thank you

pkg/3scale/amp/operator/redis_reconciler.go
valerymo: sure, sorry, yes, renamed all related things. Thank you

pkg/3scale/amp/operator/redis_reconciler.go
valerymo: Done. Thank you

pkg/3scale/amp/operator/redis_reconciler.go
MStokluska: hmm..please re-check if you can. It seems a bit late to be checking if the secret is present because we already created deployments that reference the secret (in envs)

pkg/3scale/amp/operator/redis_reconciler.go
MStokluska: I think I have maybe worded it incorrectly. 
Doing this kind of style update is good when you need to remove multiple things but you don't know if these things are present. For example, if you remove "NAMESPACE" "NAMESPACE1" "NAMESPACE2" - you use for loop to go through them and find them, if found, you remove and set update to true. 
Then, you don't have to do multiple updates but one based on the "update" env. 
In this case, we are only removing one key - "NAMESPACE" - therefore the extra logic of having additional var to indicate that update is required is not needed. The code could look something like this:
```
	if _, ok := secret.Data["NAMESPACE"]; ok {
		delete(secret.Data, namespaceKey)
		err = r.UpdateResource(secret)
		if err != nil {
			return err
		}
	}
```

pkg/3scale/amp/operator/redis_reconciler.go
MStokluska: Thanks, just checked it too, so we do create secret afterwards. Thanks

pkg/3scale/amp/operator/redis_reconciler.go
```diff
@@ -124,3 +133,29 @@ func Redis(apimanager *appsv1alpha1.APIManager, client client.Client) (*componen
 	}
 	return component.NewRedis(opts), nil
 }
+
+func (r *RedisReconciler) deleteSystemRedisSecretNamespaceKey() error {
+	secret := &corev1.Secret{}
+	err := r.Client().Get(context.TODO(), client.ObjectKey{Namespace: r.apiManager.Namespace, Name: "system-redis"}, secret)
+	if k8serr.IsNotFound(err) {
```
MStokluska, line 140: Not a must, just another way of going through error handling in case when we want to handle isNotFound.
```
if err != nil && !k8serr.IsNotFound(err) {
       return err 
} else { 
      return nil 
}
```

pkg/3scale/amp/operator/redis_reconciler.go
valerymo: oh, yes, sure. Done. Thanks

pkg/3scale/amp/operator/redis_reconciler.go
valerymo: code after this will be unaccessable, as return nil. But yes, it could be written different way, sure. Thank you

MStokluska: Good point, long day. 
```
if err != nil {
	if !k8serr.IsNotFound(err) {
		return err
	} else {
		return nil
	}
}
```

MStokluska (approved):

austincunningham (approved):

PR #961, opened 2024-04-02
THREESCALE-10937 update upstream db images

MStokluska:
Jira: https://issues.redhat.com/browse/THREESCALE-10937

Move the centos images to sclorg images since centos7 is no longer maintained.

Verification:
For MySQL and Redis e2e passing is enough
For PostgreSQL, pull the image locally to confirm it's available.

### DIFF
--- a/pkg/helper/podHelper.go
+++ b/pkg/helper/podHelper.go
@@ -21,6 +21,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	redisDefaultImage      = "quay.io/sclorg/redis-6-c8s"
+	mySqlDefaultImage      = "quay.io/sclorg/mysql-80-c8s"
+	postgreSqlDefaultImage = "quay.io/sclorg/postgresql-10-c8s"
+)
+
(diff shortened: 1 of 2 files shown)


pkg/helper/podHelper.go
```diff
@@ -21,6 +21,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	redisDefaultImage      = "quay.io/sclorg/redis-6-c8s"
+	mySqlDefaultImage      = "quay.io/sclorg/mysql-80-c8s"
+	postgreSqlDefaultImage = "quay.io/sclorg/postgresql-10-c8s"
```
austincunningham, line 27: As the images are only used once I don't see the need to have them defined as consts. Just use the image strings

MStokluska: I moved these to consts because it makes it easier to update (read the file when updating the images) and you don't have to be going through the file to find and update images.

austincunningham: Prefer reading it the other way as I don't have to jump around a file to find out which image is used, but thats me.  Personal preference. I will leave it up to you

/lgtm

PR #986, opened 2024-06-11
THREESCALE-11105 bump maxOpenshiftVersion to 4.16

MStokluska:


austincunningham:
think you need to update the bases https://github.com/3scale/3scale-operator/pull/877 as well

PR #989, opened 2024-06-18
doc: product policy configuration from secrets

eguzki:


MStokluska:
:eyes:

MStokluska:
hey @eguzki this doesn't seem to work as expected with error:
```
error(*encoding/json.SyntaxError) *{msg: "invalid character 'r' looking for beginning of value", Offset: 1}
```
It fails here: https://github.com/3scale/3scale-operator/blob/master/controllers/capabilities/product_policies.go#L100

I think this is because format of the string appears to be yaml and not json.
We can extend the code within product policies reconciler to support json and yaml.

MStokluska (changes requested):
Please use json instead of yaml:
```
{
  "rules": [
    {
      "header": "X-Cache-Status",
      "condition": {
        "operations": [
          {
            "left_type": "plain",
            "right_type": "plain",
            "op": "==",
            "right": "GET",
            "left": "ein"
          },
          {
            "left_type": "plain",
            "right_type": "plain",
            "op": "!=",
            "right": "RIGHTein",
            "left": "elftein"
          }
        ],
        "combine_op": "and"
      },
      "cache": true
    }
  ]
}
```

eguzki:
Good catch @MStokluska 

I fixed the doc and tested it as well. Now it seems correct. 

I also added a note to make clear it should be `json` encoded value. 

I also simplified the secret creation step.

MStokluska (approved):

PR #1004, opened 2024-08-07
THREESCALE-11100 Support updates to OAS-generated Products

carlkyrillos:
# Issue Link
JIRA: [THREESCALE-11100](https://issues.redhat.com/browse/THREESCALE-11100)

# What
This PR allows users to make update to 3scale Products and Backends created via an OpenAPI CR. At a high level this PR does three things:
1. Any changes to Product CRs or Backend CRs that are owned by an OpenAPI CR will be reverted by the operator. This is needed to ensure that the OpenAPI Spec (OAS) and OpenAPI CR are the source of truth.
2. When an OpenAPI CR gets its OAS source from a Secret, that Secret will be watched by the operator and any changes made to the Secret will be passed on to the Product and Backend.
3. When an OpenAPI CR gets its OAS source from a URL, the operator will periodically scrape that URL for changes. If changes are detected they will be passed on to the Product and Backend. The interval in which the operator will check the URL for changes is defined by an annotation on the OpenAPI CR. This annotation is set to `poll_url_interval: "5m"` (5 minutes) by default. The user can change this interval to any value with the following units: `"ns"`, `"us"` (or `"µs"`), `"ms"`, `"s"`, `"m"`, `"h"`. If the user sets the interval to a invalid value that can't be parsed, the operator will revert the annotation to the default value of `"5m"`.

# Verification Steps
### Setup
1. Checkout this PR

2. Run the operator locally or via OLM using prebuilt image
```bash
BUNDLE_IMG=quay.io/ckyrillo/3scale-operator-bundles:threescale-11100 make bundle-run
```

3. Create a Namespace, dummy s3 Secret, and APIManager CR:
```bash
export NAMESPACE=3scale-test
oc new-project $NAMESPACE
cat << EOF | oc create -f -
kind: Secret
apiVersion: v1
metadata:
  name: s3-credentials
  namespace: $NAMESPACE
data:
  AWS_BUCKET: c29tZXRoaW5nCg==
  AWS_REGION: dXMtd2VzdC0xCg==
type: Opaque
EOF

cat << EOF | oc create -f -
kind: APIManager
metadata:
  name: 3scale
  namespace: $NAMESPACE
spec:
  wildcardDomain: $DOMAIN
  system:
    fileStorage:
      simpleStorageService:
        configurationSecretRef:
          name: s3-credentials
EOF
```

```bash
```

### Secret Ref Verification
1. Create the OAS Secret:
<details>

```bash
touch extended-oas-secret-data.yaml

cat << EOF > extended-oas-secret-data.yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Extended Petstore
  license:
    name: MIT
servers:
  - url: http://petstore.swagger.io/v1
paths:
  /pets:
    get:
      summary: List all pets
      operationId: listPets
      x-3scale-operation:
        mappingRule:
          metricMethodRef: "metric01"
          increment: 2
          last: true
      responses:
        '200':
          description: A paged array of pets
x-3scale-product:
  metrics:
    metric01:
      friendlyName: "My Metric 01"
      unit: "hits"
      description: "This is a custom metric"
EOF

oc create secret generic extended-oas-secret --from-file extended-oas-secret-data.yaml
```
</details>

2. Create the OpenAPI CR:
```bash
cat << EOF | oc create -f -
kind: OpenAPI
metadata:
  name: extended-oas
spec:
  openapiRef:
    secretRef:
      name: extended-oas-secret
EOF
```

3. Verify that the OpenAPI CR is ready:
```bash
oc get openapi extended-oas -oyaml |  yq e '.status.conditions[] | select(.type == "Ready")'
```

4. Verify that corresponding Product CR was created:
```bash
oc get product $(oc get products | grep -w extendedpetstore | awk '{print $1}') -oyaml
```

5. Login to the 3scale Admin UI and confirm that the Product exists

6. Make a change to the OAS Secret, for example change the `description` field for the `metric01` metric.

7. Verify that the change was passed on the Product CR

8. Verify that the change was passed on to the Product in the 3scale Admin UI

### URL Ref Verification
This is tricky to test because you will need to setup a dummy git repository similar to [this](https://github.com/carlkyrillos/3scale-oas-examples/tree/main) and provide the [raw url](https://raw.githubusercontent.com/carlkyrillos/3scale-oas-examples/main/extended-oas-example.yaml) to the OpenAPI CR. It's probably easier to verify this offline.


MStokluska:
@carlkyrillos with the label addition is this sentence correct?
```
the OpenAPI Spec (OAS) and OpenAPI CR are the source of truth
```
it seems to me that in case of secret the oas secret is the only source of truth.
in case of URL, the OpenAPI CR is source of truth in between reconciliation loops but then the URL oas?

carlkyrillos:
> @carlkyrillos with the label addition is this sentence correct?
> 
> ```
> the OpenAPI Spec (OAS) and OpenAPI CR are the source of truth
> ```
> 
> it seems to me that in case of secret the oas secret is the only source of truth. in case of URL, the OpenAPI CR is source of truth in between reconciliation loops but then the URL oas?

Yes it's still true that the OAS and OpenAPI CR are _both_ sources of truth. Regardless if using urlRef or secretRef, the OpenAPI CR is the source of truth for auth related settings via the `.spec.deployment` block and the OAS is the source of truth for everything else.

controllers/capabilities/openapi_controller.go
```diff
@@ -220,6 +250,15 @@ func (r *OpenAPIReconciler) reconcileSpec(openapiCR *capabilitiesv1beta1.OpenAPI
 	}
 
 	statusReconciler := NewOpenAPIStatusReconciler(r.BaseReconciler, openapiCR, providerAccount.AdminURLStr, err, productSynced)
+
+	// If the product is successfully synced AND the OpenAPI CR is using URL ref, then requeue after set amount of time
+	// The duration is parsed from an annotation on the OpenAPI CR
+	// We have to requeue like this in case there were updates to the URL source because we can't watch the URL directly
+	if productSynced && openapiCR.Spec.OpenAPIRef.SecretRef == nil {
+		duration := getPollURLInterval(openapiCR.GetAnnotations())
```
MStokluska, line 258: something small but would it be better to move that func to openapi_types?

controllers/capabilities/openapi_controller.go
```diff
@@ -261,13 +300,108 @@ func (r *OpenAPIReconciler) checkProductSynced(resource *capabilitiesv1beta1.Ope
...
+	}
+	openAPICR.ObjectMeta.Labels[openAPISecretRefLabelKey] = string(oasSourceSecret.GetUID())
+	if err := r.Client().Update(r.Context(), openAPICR); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (r *OpenAPIReconciler) annotateOpenAPICR(openAPICR *capabilitiesv1beta1.OpenAPI) error {
```
MStokluska, line 371: something small but would it be better to move that func to openapi_types?

controllers/capabilities/openapi_controller.go
```diff
@@ -261,13 +300,108 @@ func (r *OpenAPIReconciler) checkProductSynced(resource *capabilitiesv1beta1.Ope
...
+		if err := r.Client().Update(r.Context(), openAPICR); err != nil {
+			return err
+		}
+	} else {
+		// If the user provided interval is invalid, reset it to the default value
+		// This is to ensure a valid interval exists and to let the user know that their interval couldn't be parsed
+		_, err := time.ParseDuration(intervalStr)
+		if err != nil {
+			openAPICR.Annotations[pollOasURLIntervalAnnotation] = defaultPollOasURLInterval
+			if err2 := r.Client().Update(r.Context(), openAPICR); err2 != nil {
```
MStokluska, line 396: isn't that if we are not doing anything with err but using it just to update the annotation we don't need it and therefore the above can be again - err? 
It looks like we could really use admission webhook here I think, it might be too much though...
Just wonder if there's anything we could do here that would communicate to users about the error. 
Also I'm not sure I like the idea of operator reverting it to default when wrong :/

controllers/capabilities/openapi_controller.go
carlkyrillos: I removed this function altogether because we decided to remove functionality to let the user specify the OAS URL scrape interval.

controllers/capabilities/openapi_controller.go
carlkyrillos: I removed this function altogether because we decided to remove functionality to let the user specify the OAS URL scrape interval.

controllers/capabilities/openapi_controller.go
carlkyrillos: I removed this function altogether because we decided to remove functionality to let the user specify the OAS URL scrape interval.

MStokluska:
hey @carlkyrillos 
```
      responses:
        '200':
          description: A paged array of pets
```
value in the oas secret, not related to this PR but where is this presented in 3scale?

MStokluska:
hey @carlkyrillos 
Possibly not related to this PR but I guess worth addressing, another OpenAPI CR can be created for the same secret, which results in updating 3scale products based on updated CR. Might be worth to look into validating if OpenAPI CR is unique

MStokluska:
hey @carlkyrillos I've noticed when deleting 3scale namespace that oas CR and backend CRs are not deleted, makes sense since they have finalizers. Product CRs were deleted, can we double check if they have finalizers?

carlkyrillos:
> hey @carlkyrillos
> 
> ```
>       responses:
>         '200':
>           description: A paged array of pets
> ```
> 
> value in the oas secret, not related to this PR but where is this presented in 3scale?

This isn't presented in 3scale. It's left over unrelated OAS that I got from the generic petstore example.

carlkyrillos:
> hey @carlkyrillos Possibly not related to this PR but I guess worth addressing, another OpenAPI CR can be created for the same secret, which results in updating 3scale products based on updated CR. Might be worth to look into validating if OpenAPI CR is unique

Good catch I will make a note of it to address it in [THREESCALE-10564](https://issues.redhat.com/browse/THREESCALE-10564) which also deals with OpenAPI validation

carlkyrillos:
> hey @carlkyrillos I've noticed when deleting 3scale namespace that oas CR and backend CRs are not deleted, makes sense since they have finalizers. Product CRs were deleted, can we double check if they have finalizers?

Just confirmed that the Product CR does have a finalizer `product.capabilities.3scale.net/finalizer`

MStokluska (approved):
Thanks @carlkyrillos it looks good to me. We can address some questions in another Jira. Verified on cluster and looks fine.

PR #1044, opened 2024-12-13
THREESCALE-10577 Make SystemName field in Backend CR immutable

carlkyrillos:
# Issue Link
JIRA: [THREESCALE-10577](https://issues.redhat.com/browse/THREESCALE-10577)

# What
This PR makes the `SystemName` field in the Backend CR immutable. This prevents an issue where previously if a user changed the `SystemName` on an already created Backend CR, a duplicate Backend would be created in the database.

Note that in the 3scale UI, the `SystemName` field is already immutable so this PR just makes the capabilities behave the same way as the UI.

# Verification Steps
1. Checkout this branch

2. Prepare for a local install:
```bash
make install
make download

export NAMESPACE=3scale-test
oc new-project $NAMESPACE

cat << EOF | oc create -f -
kind: Secret
apiVersion: v1
metadata:
  name: s3-credentials
  namespace: $NAMESPACE
data:
  AWS_BUCKET: c29tZXRoaW5nCg==
  AWS_REGION: dXMtd2VzdC0xCg==
type: Opaque
EOF

cat << EOF | oc create -f -
kind: APIManager
metadata:
  name: 3scale
  namespace: $NAMESPACE
spec:
  wildcardDomain: $DOMAIN
  system:
    fileStorage:
      simpleStorageService:
        configurationSecretRef:
          name: s3-credentials
EOF
```

3. Run the operator locally:
```bash
make run 
```

```bash
```

5. Create a Backend:
```bash
cat << EOF | oc create -f -
---
kind: Backend
metadata:
  name: backend1-cr
spec:
  name: testbackend
  privateBaseURL: 'http://testbackend.svc:8080'
  systemName: backend1
EOF
```

6. Verify that the Backend was created successfully in the admin portal.

7. Attempt to change the `SystemName` field in the Backend CR and confirm that you get an error message saying `SystemName is immutable`.

8. Verify in the admin portal that the SystemName wasn't changed and that the Backend wasn't duplicated.

### DIFF
--- a/config/crd/bases/capabilities.3scale.net_backends.yaml
+++ b/config/crd/bases/capabilities.3scale.net_backends.yaml
@@ -125,6 +125,9 @@ spec:
                 description: SystemName identifies uniquely the backend within the
                   account provider Default value will be sanitized Name
                 type: string
+                x-kubernetes-validations:
+                - message: SystemName is immutable
+                  rule: self == oldSelf
             required:
             - name
             - privateBaseURL
--- a/apis/capabilities/v1beta1/backend_types.go
+++ b/apis/capabilities/v1beta1/backend_types.go
@@ -71,6 +71,7 @@ type BackendSpec struct {
 	// SystemName identifies uniquely the backend within the account provider
 	// Default value will be sanitized Name
 	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="SystemName is immutable"
 	SystemName string `json:"systemName,omitempty"`
 
 	// PrivateBaseURL Private Base URL of the API


config/crd/bases/capabilities.3scale.net_backends.yaml
```diff
@@ -125,6 +125,9 @@ spec:
                 description: SystemName identifies uniquely the backend within the
                   account provider Default value will be sanitized Name
                 type: string
+                x-kubernetes-validations:
```
valerymo, line 128: Hi @carlkyrillos , this approach to make field immutable looks ok, but I'd like to ask if you compared it with K8S "native" way , as following 
```
                systemName:
                  type: string
                  x-kubernetes-immutable: true
```
I asked ChatGPT to compare both approaches, and this is table below

### **Comparison of Both Approaches**

| **Feature**                  | **`x-kubernetes-immutable: true` (in CRD)**                     | **`+kubebuilder:validation:XValidation` (in Go)**              |
|------------------------------|------------------------------------------------------------------|-----------------------------------------------------------------|
| **Where is the validation enforced?** | Kubernetes API server (global enforcement)                   | Operator (custom code in controller-runtime)                     |
| **Scope of enforcement**      | Validates immutability across all Kubernetes clients and tools (kubectl, API, etc.) | Validates only when the operator is active and running          |
| **Ease of implementation**    | Simple, no code needed outside the CRD definition               | Requires adding validation logic to the Go code (operator)      |
| **Customization**             | No customization beyond immutability                           | Can customize the validation rule and error message             |
| **Error handling**            | Automatically handled by the Kubernetes API server             | You need to handle errors manually in the operator code         |
| **Flexibility**               | Only supports basic immutability                               | Supports complex validation logic (e.g., conditional immutability) |
| **Operational dependencies**  | No dependency on the operator or controller                    | Depends on the operator running to enforce validation          |

valerymo: I checked solution. It's working.

valerymo: Regarding second way I noted before. Checked again, and meanwhile it's not working for me, option looks like not accepted. I also more checked google, Not ChatGpt , and I see your approach. So it's good. Thank you

MStokluska (approved):

PR #1047, opened 2024-12-18
THREESCALE-11502 Add flag to disable zync

carlkyrillos:
# Issue Link
JIRA: [THREESCALE-11502](https://issues.redhat.com/browse/THREESCALE-11502)

# What
This PR adds a flag that allows the user to disable the creation of zync components. This flag, `.spec.zync.enabled`, is set to `true` by default but when it is set to `false`, the operator will skip creating the following components:
* zync/zync-que/zync-database Deployments
* zync-que RBAC(Role, RoleBinding, and ServiceAccount)
* zync Service
* zync Secret
* zync/zync-que PodDisruptionBudgets
* zync/zync-que PodMonitors
* zync GrafanaDashboards
* zync/zync-que PrometheusRules

# Verification Steps

### Scenario 1: `.spec.zync.enabled` initially not specified, then set to `false`

1. Checkout this branch

2. Prepare for a local install:
```bash
make install
make download

export NAMESPACE=3scale-test
oc new-project $NAMESPACE

cat << EOF | oc create -f -
kind: Secret
apiVersion: v1
metadata:
  name: s3-credentials
  namespace: $NAMESPACE
data:
  AWS_BUCKET: c29tZXRoaW5nCg==
  AWS_REGION: dXMtd2VzdC0xCg==
type: Opaque
EOF

cat << EOF | oc create -f -
kind: APIManager
metadata:
  name: 3scale
  namespace: $NAMESPACE
spec:
  wildcardDomain: $DOMAIN
  system:
    fileStorage:
      simpleStorageService:
        configurationSecretRef:
          name: s3-credentials
EOF
```

3. Run the operator locally:
```bash
make run 
```

```bash
```

5. Note that `.spec.zync.enabled` was set to `true` by default:
```bash
oc get apimanager 3scale -n 3scale-test -oyaml | yq .spec.zync
```

6. Change the APIManager's `.spec.zync.enabled` to `false`:
```bash
oc patch apimanager 3scale -n 3scale-test --type='json' -p '[{"op": "replace", "path": "/spec/zync/enabled", "value": false}]'
```

7. Wait for the install to reconcile:
```bash
```

8. Verify that the following zync components have been removed (note: there shouldn't be any output for all of the below commands):
```bash
oc get rolebindings -A | grep zync-que-rolebinding
oc get roles -A | grep zync-que-role
oc get serviceaccounts -A | grep zync-que-sa
oc get deployments -A | grep zync
oc get services -A | grep zync
oc get secrets -A | grep zync
```

9. Clean up the cluster to prepare for Scenario 2

### Scenario 2: `.spec.zync.enabled` initially set to `false`, then set to `true`

1. Prepare for a local install with `.spec.zync.enabled` set to `false`:
```bash
make install
make download

export NAMESPACE=3scale-test
oc new-project $NAMESPACE

cat << EOF | oc create -f -
kind: Secret
apiVersion: v1
metadata:
  name: s3-credentials
  namespace: $NAMESPACE
data:
  AWS_BUCKET: c29tZXRoaW5nCg==
  AWS_REGION: dXMtd2VzdC0xCg==
type: Opaque
EOF

cat << EOF | oc create -f -
kind: APIManager
metadata:
  name: 3scale
  namespace: $NAMESPACE
spec:
  wildcardDomain: $DOMAIN
  zync:
    enabled: false
  system:
    fileStorage:
      simpleStorageService:
        configurationSecretRef:
          name: s3-credentials
EOF
```

2. Run the operator locally:
```bash
make run 
```

3. Wait for the install to complete:
```bash
```

4. Verify that the following zync components were not provisioned:
```bash
oc get rolebindings -A | grep zync-que-rolebinding
oc get roles -A | grep zync-que-role
oc get serviceaccounts -A | grep zync-que-sa
oc get deployments -A | grep zync
oc get services -A | grep zync
oc get secrets -A | grep zync
```

5. Change the APIManager's `.spec.zync.enabled` to `true`:
```bash
oc patch apimanager 3scale -n 3scale-test --type='json' -p '[{"op": "replace", "path": "/spec/zync/enabled", "value": true}]'
```

6. Wait for the install to reconcile:
```bash
```

7. Verify that the following zync components were created:
```bash
oc get rolebindings -n 3scale-test | grep zync-que-rolebinding
oc get roles -n 3scale-test | grep zync-que-role
oc get serviceaccounts -n 3scale-test | grep zync-que-sa
oc get deployments -n 3scale-test | grep zync
oc get services -n 3scale-test | grep zync
oc get secrets -n 3scale-test | grep zync
```

8. Verify that all 6 standard routes were created:
```bash
oc get routes -n 3scale-test
```


pkg/3scale/amp/operator/zync_reconciler.go
```diff
@@ -206,6 +215,152 @@ func (r *ZyncReconciler) Reconcile() (reconcile.Result, error) {
 	return reconcile.Result{}, nil
 }
 
+// deleteZyncComponents handles the removal of all zync components
+// This should only happen when the APIManager's .spec.zync.enabled was initially set to true but then changed to false
```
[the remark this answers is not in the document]
carlkyrillos: @valerymo `deleteZyncComponents()` should be called on each reconciliation cycle as long as zync has been marked as `enabled: false`. I updated the comment above the function to make it more clear.

valerymo:
Hi @carlkyrillos , I tested validation scenarious. Could you please look at comments below
1. Scenario 1 - is fine, working as expected, as described in validation notes + routes removed correctly (backend route remains). 
2. Scenario 2 - Routes were not recreated when Zync was enabled. Just one route exists - backedn. Retested twice - same issue.
Thank you

tkan145:
Sorry for dump question, but why do we need a new flag when user can just set the zync/zync-que replicate to 0?

zync/zync-que also sync changes to RH-SSO so disable it completely would cause trouble. From the original ticket [THREESCALE-8102](https://issues.redhat.com/browse/THREESCALE-8102), the way to disable route creation is to set `DISABLE_K8S_ROUTES_CREATION = 1` env var.

carlkyrillos:
> Sorry for dump question, but why do we need a new flag when user can just set the zync/zync-que replicate to 0?
> 
> zync/zync-que also sync changes to RH-SSO so disable it completely would cause trouble. From the original ticket [THREESCALE-8102](https://issues.redhat.com/browse/THREESCALE-8102), the way to disable route creation is to set `DISABLE_K8S_ROUTES_CREATION = 1` env var.

Just to summarize our conversation from Slack: we agreed to keep the current implementation since it's almost complete but we need to tweak the logic to set the zync endpoint to an empty string to avoid spamming system-sidekiq with ZyncWorker jobs.

valerymo:
@carlkyrillos , @MStokluska , @tkan145   I retested Snenario 2, and it looks fine for me. Code also is good to me. Thank you

controllers/apps/apimanager_controller.go
```diff
@@ -470,7 +470,7 @@ func (r *APIManagerReconciler) reconcileAPIManagerLogic(cr *appsv1alpha1.APIMana
 		return result, err
 	}
 
-	zyncReconciler := operator.NewZyncReconciler(baseAPIManagerLogicReconciler)
+	zyncReconciler := operator.NewZyncReconciler(baseAPIManagerLogicReconciler, cr.IsZyncEnabled())
```
MStokluska, line 473: Instead of doing that, would it not be better to use the GetZyncOptions func? Not a big deal

doc/apimanager-reference.md
```diff
@@ -517,6 +517,7 @@ Note: Deploying databases internally with this section is meant for evaluation p
 
 | **Field** | **json/yaml field**| **Type** | **Required** | **Default value** | **Description** |
 | --- | --- | --- | --- | --- | --- |
+| Enabled | `enabled` | bool | No | `true` | Zync components are enabled by default. Set to false to disable all Zync components. All other zync related settings will be ignored. |
```
MStokluska, line 520:
```suggestion
| Enabled | `enabled` | bool | No | `true` | Zync components are enabled by default. Set to false to disable all Zync components. If disabled, all other zync related settings will be ignored. |
```

pkg/3scale/amp/operator/system_reconciler.go
```diff
@@ -90,7 +92,7 @@ func (r *SystemReconciler) Reconcile() (reconcile.Result, error) {
 	}
 
 	// System CM
-	err = r.ReconcileConfigMap(system.SystemConfigMap(), reconcilers.CreateOnlyMutator)
+	err = r.ReconcileConfigMap(system.SystemConfigMap(), systemConfigMapMutator)
```
MStokluska, line 95: so this was a create only mutator. Meaning we are allowing changing other values post create.
Is that logic now changes and we will overwrite the operator expected values with potentially updated values?

pkg/3scale/amp/operator/system_reconciler.go
```diff
@@ -322,6 +326,34 @@ func (r *SystemReconciler) Reconcile() (reconcile.Result, error) {
 		return reconcile.Result{RequeueAfter: 5 * time.Second}, nil
 	}
 
+	// Check if enough routes exist
+	routesExist, err := r.routesExist()
```
MStokluska, line 330: why are we doing this?

MStokluska: resyncRoutesJob nvm

MStokluska: I don't know if this is actually good enough...
So let's say I have 100s of routes created.
I disable zync - do all my routes get deleted?

pkg/3scale/amp/operator/system_reconciler.go
```diff
@@ -322,6 +326,34 @@ func (r *SystemReconciler) Reconcile() (reconcile.Result, error) {
...
+	}
+
+	// Check if zync deployments are ready
+	zyncReady, err := r.isZyncReady()
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// If zync is enabled and zync is ready but the routes are missing then create the ResyncRoutes Job
+	if system.Options.ZyncEnabled && zyncReady && !routesExist {
```
MStokluska, line 342: I know we are making the call against system..but this feels to me like a zync reconciler task rather than system task..

pkg/3scale/amp/operator/system_reconciler.go
```diff
@@ -373,6 +405,89 @@ func (r *SystemReconciler) systemAppDeploymentResourceMutator(desired, existing
...
+	}
+
+	if len(routes.Items) < expectedRoutes {
+		r.Logger().Info(fmt.Sprintf("Missing routes - expected at least %d routes but only found %d", expectedRoutes, len(routes.Items)))
+		return false, nil
+	}
+	return true, nil
+}
+
+func (r *SystemReconciler) isZyncReady() (bool, error) {
```
MStokluska, line 442: this feels to me like a function that could be used by other reconcilers in future as well? Possibly a shared func? Not a big deal if it stays

pkg/3scale/amp/operator/system_reconciler.go
MStokluska: Then, I guess we need to check if system is ready right? Also, does the system-reconciler completes when system is up and running OR when system deployment reports completed? 
I guess what I'm asking is - is there a chance that we are going to run the re-sync job against non-working system?

pkg/3scale/amp/operator/system_reconciler.go
```diff
@@ -373,6 +405,89 @@ func (r *SystemReconciler) systemAppDeploymentResourceMutator(desired, existing
...
+			desired.Spec.Template.Spec.Containers[idx].Env,
+			&existing.Spec.Template.Spec.Containers[idx].Env,
+			"ZYNC_AUTHENTICATION_TOKEN")
+		update = update || tmpChanged
+	}
+
+	return update, nil
+}
+
+func (r *SystemReconciler) routesExist() (bool, error) {
```
MStokluska, line 423: this seems like something apim status reconciler could use as well no?

pkg/3scale/amp/operator/system_reconciler.go
```diff
@@ -373,6 +405,89 @@ func (r *SystemReconciler) systemAppDeploymentResourceMutator(desired, existing
...
+		deployment := &k8sappsv1.Deployment{}
+		err := r.Client().Get(r.Context(), k8sclient.ObjectKey{
+			Namespace: r.apiManager.GetNamespace(),
+			Name:      deploymentName,
+		}, deployment)
+
+		// Return error if can't get deployment
+		if err != nil && !k8serr.IsNotFound(err) {
+			return false, fmt.Errorf("error getting deployment %s: %w", deployment.Name, err)
+		}
```
MStokluska, line 455: for all this..can we not just rely on apim deployment status where zync is either ready/starting or stopped?

pkg/3scale/amp/operator/zync_reconciler.go
```diff
@@ -206,6 +215,153 @@ func (r *ZyncReconciler) Reconcile() (reconcile.Result, error) {
...
 }
 
+// deleteZyncComponents handles the removal of all zync components
+// The operator only enters this function when the APIManager .spec.zync.enabled was initially set to true but then changed to false
+// When this happens, the operator will call this function on each reconcile loop to ensure the resources are cleaned up
+func (r *ZyncReconciler) deleteZyncComponents(zync *component.Zync, ampImages *component.AmpImages) error {
+	// ZyncQue PrometheusRules
+	zyncQuePrometheusRules := zync.ZyncPrometheusRules()
+	common.TagObjectToDelete(zyncQuePrometheusRules)
+	err := r.ReconcilePrometheusRules(zyncQuePrometheusRules, reconcilers.CreateOnlyMutator)
```
MStokluska, line 225: A bit confused about this...
So I understand what we want to do here, but does it really work if it's a CreateOnlyMutator ?!
I mean..so if we have prometheusRules, and then we tag it but run it through CreateOnlyMutator will it really delete the rule?

pkg/3scale/amp/operator/system_reconciler.go
carlkyrillos: Yes all routes (except the backend route which is created by the operator) will get deleted but I thought this was what we wanted. 

@briangallagher Was this what you were talking about the other day wrt MT RHOAM on the DeveloperSandbox?

pkg/3scale/amp/operator/system_reconciler.go
carlkyrillos: No there's no risk of running the re-sync Job against a non-working system because [this block](https://github.com/3scale/3scale-operator/blob/master/pkg/3scale/amp/operator/system_reconciler.go#L324-L328) requeues the reconciler until `system-app` is fully ready. I initially had the Job in the zync reconciler but I thought it was better suited for the system reconciler because the command is run in a porta (`system-sidekiq`) image.

pkg/3scale/amp/operator/zync_reconciler.go
carlkyrillos: The deletion does work as intended even when using the CreateOnlyMutator but I agree it's counter intuitive. I'll create a DeleteMutator to make it more clear.

pkg/3scale/amp/operator/system_reconciler.go
carlkyrillos: We can but I figured fetching the Deployments directly was more accurate in case the APIManager .status hadn't been updated yet in the reconciliation flow

pkg/3scale/amp/operator/system_reconciler.go
carlkyrillos: Good call, I'll move the function

pkg/3scale/amp/operator/system_reconciler.go
carlkyrillos: I'll move this function as well

controllers/apps/apimanager_controller.go
carlkyrillos: I think the current approach `cr.IsZyncEnabled()` is less computationally expensive because to use `GetZyncOptions()` you first need to create a NewZyncOptionsProvider which we then do again when creating the reconciler right after. So I think I'll leave as is if it's all the same.

pkg/3scale/amp/operator/system_reconciler.go
carlkyrillos: Yes good catch, any user-defined values in the `zync.yml` field would be overwritten. I'll tweak the mutator to account for that

pkg/3scale/amp/operator/system_reconciler.go
MStokluska: thanks, IMO this should be a part of "Enable Zync" function etc, but, no strong opinion, I leave it up to you.

pkg/3scale/amp/operator/zync_reconciler.go
MStokluska: I think that is fine if it works, I would leave it.

Just thinking out loud here...would not be easier to tag all zync objects with zync deployment as the owner so we leverage garbage collector?

However, although it would make cleanup significantly easier, is this creating a risk of non-working zync if someone decides just to nuke the deployment and have operator re-creating it?

WDYT?

MStokluska: just to be clear, not a big deal to leave it as it is as well..

pkg/3scale/amp/operator/system_reconciler.go
carlkyrillos: Yes this can give us a false positive. I'll leverage the existing `defaultRoutesReady()` in  `apimanager_status_reconciler.go` instead.

doc/apimanager-reference.md
carlkyrillos: Fixed in latest commit

pkg/3scale/amp/operator/system_reconciler.go
carlkyrillos: Fixed in latest commit

pkg/3scale/amp/operator/system_reconciler.go
carlkyrillos: Fixed in latest commit

pkg/3scale/amp/operator/system_reconciler.go
carlkyrillos: Fixed in latest commit

pkg/3scale/amp/operator/zync_reconciler.go
carlkyrillos: Fixed in latest commit

MStokluska (approved):

PR #1094, opened 2025-06-18
THREESCALE-11805 refactor system-app migration jobs

tkan145:
## What

Fix https://issues.redhat.com/browse/THREESCALE-11805

Change the hook jobs to fire when the deployment revision changes instead of generation, meaning the jobs will only run when a new deployment is deployed/rolled out (similar to DeploymentConfig hooks)

IMO, this is still not the best approach as it relies heavily on the fact that users won't delete completed jobs. The alternative to using the operator version is unnecessarily complex so I'm happy with this solution for now until I figure out a better one.

## Verification steps
* Checkout this branch
* Prepare cluster
```
make cluster/prepare/local
```
* Create APIManager
``` bash
export NAMESPACE=3scale-test

cat << EOF | oc create -f -
kind: Secret
apiVersion: v1
metadata:
  name: s3-credentials
  namespace: $NAMESPACE
data:
  AWS_BUCKET: c29tZXRoaW5nCg==
  AWS_REGION: dXMtd2VzdC0xCg==
type: Opaque
EOF

cat << EOF | oc create -f -
kind: APIManager
metadata:
  name: 3scale
  namespace: $NAMESPACE
spec:
  wildcardDomain: $DOMAIN
  system:
    fileStorage:
      simpleStorageService:
        configurationSecretRef:
          name: s3-credentials
  externalComponents:
    backend:
      redis: true
    system:
      database: true
      redis: true
EOF
```
* Start the operator
```
make run
```
* Wait for the deployment to complete
* The system-app pre/post hooks should run once
* Change the system-app replias to `0`
* system-app pre/post hooks should not trigger
* Change the system-app replias to `1`. Again this should not trigger the job
* Rollout
```
oc rollout restart deployment/system-app
```
* The above will trigger new hook jobs

### DIFF
--- a/controllers/apps/apimanager_controller.go
+++ b/controllers/apps/apimanager_controller.go
@@ -19,10 +19,7 @@ package controllers
 import (
 	"context"
 	"fmt"
-	"github.com/3scale/3scale-operator/pkg/3scale/amp/component"
 	"time"
-
-	"github.com/3scale/3scale-operator/pkg/upgrade"
 
 	"github.com/go-logr/logr"
 	routev1 "github.com/openshift/api/route/v1"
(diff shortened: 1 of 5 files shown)


MStokluska (approved):
Looks good to me An, thanks! 
Verified on cluster, code looks good as well!

PR #1156, opened 2026-04-07
Include controllers/apps in unit test coverage

urbanikb:
### What

Since the `make test-unit` target is the only target we currently scrape coverage from (due to e2e test being run by prow and not sending data to codecov), it would be good if the unit tests in controllers/app would also run there. 

Currently there are no unit tests defined, but there might be in the future PR. 

### How

- Move the test code that requires envtest out of directory structure for operator code
- Rename e2e test to integration test and add Makefile target

### Verification

Checked that makefile targets work as before. 
Ran the test-integration test locally against the Fyre cluster, it ran as usual.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,10 +101,10 @@ PROMETHEUS_RULES_NAMESPACE ?= "__NAMESPACE__"
 all: manager
 
 # Run all tests
-test: test-unit test-e2e test-crds test-manifests-version
+test: test-unit test-integration test-e2e test-crds test-manifests-version
 
 # Run unit tests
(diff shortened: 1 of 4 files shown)


tkan145:
As the test is not really e2e, can we have another folder called `/test/integration` instead? Then add a new entry inside Makefile to run the integration test

urbanikb:
> As the test is not really e2e, can we have another folder called `/test/integration` instead? Then add a new entry inside Makefile to run the integration test

Just to rephrase what you are asking is to rename e2e to integration, is this correct? There is no "other" e2e test that would be left.

tkan145:
Yes correct, and we need a new entry in the Makefile for the integration tests, I would also like to keep `make test-e2e` but pointing to the e2e directory, which will be empty so the test will return early. Once mergerd, I will update prow to run both [e2e](https://github.com/openshift/release/blob/main/ci-operator/config/3scale/3scale-operator/3scale-3scale-operator-master.yaml#L26) (current) and integration.

tkan145 (approved):
/lgtm
/approve

PR #1167, opened 2026-04-22
WIP: THREESCALE-14224 - Fix perpetual reconcile loops: deployment churn, status writes, HPA deletes

borisurbanik:
## What

Fixes three independent sources of unnecessary reconcile work that caused either continuous API writes or missed reconcile triggers.

### 1. Deployment update churn (K8s default normalization)

The K8s API server applies defaults (probe integers, HTTPGet scheme, container termination fields, volume DefaultModes, pod-level policies, deployment strategy) to stored Deployments. Without normalization, every reconcile pass detected a false-positive diff between the in-memory desired state and the API-server-enriched existing state, issuing an unnecessary update on every cycle.

Introduces `normalizeDeploymentDefaults` (and its call chain through `normalizePodSpecDefaults`, `normalizeContainerDefaults`, `normalizeProbeDefaults`, `normalizeVolumeDefaults`) in `pkg/reconcilers/k8s_defaults.go`, called at the top of `DeploymentMutator` before any `DMutateFn` runs.

### 2. Status reconciler write loop and unreliable route watch

- Replaced unconditional `Status().Update()` with a no-op when `newStatus` equals the current status; the previous `&& apiManagerAvailable` guard prevented the short-circuit when routes were not ready, causing a write every reconcile cycle even when nothing had changed.
- Replaced `Requeue: true` (feeds the rate-limiter backoff queue) with `RequeueAfter: 30s` when Available is False, as a safety net for missed watch events.
- Replaced the namespace-wide route informer and unreliable ownerRef-chain mapper with a label-filtered watch (`3scale.net/created-by=zync`) and a domain-name mapper that enqueues the APIManager whose expected tenant hosts overlap with the changed route's host.

### 3. HPA reconciler unnecessary delete calls

`ReconcileHpa()` was calling `DeleteResource()` directly when HPA is disabled, issuing a `Client().Delete()` API call every reconcile cycle regardless of whether the HPA still existed. After the first successful delete, subsequent cycles hit a 404 indefinitely. Replaced with `TagObjectToDelete` + `ReconcileResource`, consistent with how PDB and monitoring resources are handled.

## Testing

- Added `test/envtest/` suite with 21 tests — one per normalized field — using a real API server to confirm each default is handled and no false-positive update is produced.
- Added `make test-envtest` target included in the GitHub Actions `unit-tests` job.
- Integration test (live cluster) confirmed: 0 deployment updates after initial create.

## Status

- [x] All three fixes implemented and tested
- [ ] Observed baseline churn on 2.16 cluster — verifying fix eliminates churn before marking ready for review

🤖 Generated with [Claude Code](https://claude.com/claude-code)


borisurbanik:
Converted to epic (https://redhat.atlassian.net/browse/THREESCALE-14224), will be implemented in smaller individual PRs.

PR #1173, opened 2026-05-05
THREESCALE-14777 Bump go to v1.25

tkan145:
## What

[THREESCALE-14777](https://redhat.atlassian.net/browse/THREESCALE-14777)

I also need to update both controller-gen and golangci-lint due to changes in go1.24, and this is causing CI to fail.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -259,7 +259,7 @@ fmt:
 vet:
 	$(GO) vet ./...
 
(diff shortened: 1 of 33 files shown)


borisurbanik (approved):
Checked that the makefile targets generate the same files, checked the rbac roles are unchanged - just reorganized in the file.

PR #1174, opened 2026-05-06
THREESCALE-14652 Prevent perpetual reconcile by hardcoding K8s-defaulted fields in component specs

borisurbanik:
https://redhat.atlassian.net/browse/THREESCALE-14652

## Summary

Fixes perpetual Deployment update loops caused by the K8s API server silently filling zero-value fields with defaults on write. Mutators using `reflect.DeepEqual` then always see a mismatch between the desired spec (Go zero values) and the live spec (K8s-filled defaults), triggering an update on every reconcile cycle.

**Root cause fields explicitly set in all component Deployment specs:**
- Probe fields: `Scheme`, `TimeoutSeconds`, `PeriodSeconds`, `SuccessThreshold`, `FailureThreshold`
- Container / init-container fields: `TerminationMessagePath`, `TerminationMessagePolicy`, `ImagePullPolicy` (init containers most sensitive — `DeploymentPodInitContainerMutator` does a full struct `DeepEqual`)
- Pod spec fields: `RestartPolicy`, `DNSPolicy`, `SecurityContext`, `TerminationGracePeriodSeconds`, `SchedulerName`
- Volume source fields: `DefaultMode` on `Secret`, `ConfigMap`, and `Projected` volume sources
- Use `nil` (not `[]T{}`) for optional volume/volumemount slices

**Observability:** `UpdateResource` now logs `namespace` and APIManager owner `name` as structured fields, enabling the integration test counter to attribute each Deployment update to the correct CR instance.

**Regression test:** `ReconcileCounter` + `verifyNoDeploymentUpdates` assert that total Deployment update calls per APIManager install stay within `[1, 50]`. The floor of 1 guards against a silent counter misconfiguration; the ceiling of 50 stays well below what the perpetual-reconcile bug produced (hundreds of updates per install).

## Test plan

- [x] Integration test passes: `WATCH_NAMESPACE=dummy make test-integration`
- [x] Deployment Update Report shows `Total` between 1 and 50
- [x] No backend/zync/system deployments cycling in cluster logs after APIManager becomes Available

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/3scale/amp/component/backend.go
+++ b/pkg/3scale/amp/component/backend.go
@@ -15,6 +15,7 @@ import (
 	policyv1 "k8s.io/api/policy/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
 )
 
 const (
@@ -256,6 +257,9 @@ func (backend *Backend) CronDeployment(ctx context.Context, k8sclient client.Cli
 								},
 								InitialDelaySeconds: 30,
 								PeriodSeconds:       5,
+								TimeoutSeconds:      1,
+								SuccessThreshold:    1,
+								FailureThreshold:    3,
 							},
 						},
 					},
@@ -330,10 +334,10 @@ func (backend *Backend) ListenerDeployment(ctx context.Context, k8sclient client
 									},
 								},
 								InitialDelaySeconds: 30,
-								TimeoutSeconds:      0,
+								TimeoutSeconds:      1,
 								PeriodSeconds:       10,
-								SuccessThreshold:    0,
-								FailureThreshold:    0,
+								SuccessThreshold:    1,
+								FailureThreshold:    3,
 							},
 							ReadinessProbe: &v1.Probe{
 								ProbeHandler: v1.ProbeHandler{
@@ -343,13 +347,14 @@ func (backend *Backend) ListenerDeployment(ctx context.Context, k8sclient client
 											Type:   intstr.Int,
 											IntVal: 3000,
 										},
+										Scheme: v1.URISchemeHTTP,
 									},
 								},
 								InitialDelaySeconds: 30,
 								TimeoutSeconds:      5,
-								PeriodSeconds:       0,
-								SuccessThreshold:    0,
-								FailureThreshold:    0,
+								PeriodSeconds:       10,
+								SuccessThreshold:    1,
+								FailureThreshold:    3,
 							},
 							ImagePullPolicy: v1.PullIfNotPresent,
 						},
@@ -668,7 +673,7 @@ func (backend *Backend) QueuesRedisTLSEnvVars() []v1.EnvVar {
 }
 
 func (backend *Backend) backendVolumes() []v1.Volume {
-	res := []v1.Volume{}
+	var res []v1.Volume
 	if backend.Options.BackendRedisTLS.Enabled {
 		items := []v1.KeyToPath{}
 		if backend.Options.BackendRedisTLS.HasCA() {
@@ -682,8 +687,9 @@ func (backend *Backend) backendVolumes() []v1.Volume {
 			Name: "backend-redis-tls",
 			VolumeSource: v1.VolumeSource{
 				Secret: &v1.SecretVolumeSource{
-					SecretName: BackendSecretBackendRedisSecretName,
-					Items:      items,
+					SecretName:  BackendSecretBackendRedisSecretName,
+					Items:       items,
+					DefaultMode: ptr.To(v1.SecretVolumeSourceDefaultMode),
 				},
 			},
 		}
@@ -703,8 +709,9 @@ func (backend *Backend) backendVolumes() []v1.Volume {
 			Name: "queues-redis-tls",
 			VolumeSource: v1.VolumeSource{
 				Secret: &v1.SecretVolumeSource{
-					SecretName: BackendSecretBackendRedisSecretName,
-					Items:      items,
+					SecretName:  BackendSecretBackendRedisSecretName,
+					Items:       items,
+					DefaultMode: ptr.To(v1.SecretVolumeSourceDefaultMode),
 				},
 			},
 		}
@@ -725,7 +732,7 @@ func (backend *Backend) backendListenerRunArgs() []string {
 }
 
 func (backend *Backend) backendContainerVolumeMounts() []v1.VolumeMount {
-	res := []v1.VolumeMount{}
+	var res []v1.VolumeMount
 	if backend.Options.BackendRedisTLS.Enabled {
 		res = append(res, backend.backendRedisContainerVolumeMounts())
 	}
--- a/pkg/3scale/amp/component/backend_redis_tls_test.go
+++ b/pkg/3scale/amp/component/backend_redis_tls_test.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/utils/ptr"
 
 	"github.com/3scale/3scale-operator/pkg/helper"
 )
@@ -163,7 +164,7 @@ func TestBackendComponentRedisTLSVolumes(t *testing.T) {
 				BackendRedisTLS:       TLSConfig{Enabled: false},
 				BackendRedisQueuesTLS: TLSConfig{Enabled: false},
 			},
-			[]v1.Volume{},
+			nil,
 		},
 		{
 			"StorageOnly_OneWayTLS",
@@ -183,6 +184,7 @@ func TestBackendComponentRedisTLSVolumes(t *testing.T) {
 							Items: []v1.KeyToPath{
 								{Key: "REDIS_SSL_CA", Path: "backend-redis-ca.crt"},
 							},
+							DefaultMode: ptr.To(v1.SecretVolumeSourceDefaultMode),
 						},
 					},
 				},
@@ -206,6 +208,7 @@ func TestBackendComponentRedisTLSVolumes(t *testing.T) {
 							Items: []v1.KeyToPath{
 								{Key: "REDIS_SSL_QUEUES_CA", Path: "backend-redis-queues-ca.crt"},
 							},
+							DefaultMode: ptr.To(v1.SecretVolumeSourceDefaultMode),
 						},
 					},
 				},
@@ -238,6 +241,7 @@ func TestBackendComponentRedisTLSVolumes(t *testing.T) {
 								{Key: "REDIS_SSL_CERT", Path: "backend-redis-client.crt"},
 								{Key: "REDIS_SSL_KEY", Path: "backend-redis-private.key"},
 							},
+							DefaultMode: ptr.To(v1.SecretVolumeSourceDefaultMode),
 						},
 					},
 				},
@@ -251,6 +255,7 @@ func TestBackendComponentRedisTLSVolumes(t *testing.T) {
 								{Key: "REDIS_SSL_QUEUES_CERT", Path: "backend-redis-queues-client.crt"},
 								{Key: "REDIS_SSL_QUEUES_KEY", Path: "backend-redis-queues-private.key"},
 							},
+							DefaultMode: ptr.To(v1.SecretVolumeSourceDefaultMode),
 						},
 					},
 				},
@@ -281,6 +286,7 @@ func TestBackendComponentRedisTLSVolumes(t *testing.T) {
 								{Key: "REDIS_SSL_CERT", Path: "backend-redis-client.crt"},
 								{Key: "REDIS_SSL_KEY", Path: "backend-redis-private.key"},
 							},
+							DefaultMode: ptr.To(v1.SecretVolumeSourceDefaultMode),
 						},
 					},
 				},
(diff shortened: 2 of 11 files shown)


pkg/reconcilers/base_reconciler.go
```diff
@@ -160,7 +160,17 @@ func (b *BaseReconciler) CreateResource(obj client.Object) error {
...
 func (b *BaseReconciler) UpdateResource(obj client.Object) error {
-	b.Logger().Info(fmt.Sprintf("Updated object '%s/%s'", strings.Replace(fmt.Sprintf("%T", obj), "*", "", 1), obj.GetName()))
+	typeName := strings.Replace(fmt.Sprintf("%T", obj), "*", "", 1)
+	crName := ""
+	for _, ref := range obj.GetOwnerReferences() {
+		if ref.Kind == "APIManager" {
+			crName = ref.Name
+			break
+		}
+	}
```
tkan145: What is this and why do we need this?

test/integration/apimanager_suite_test.go
```diff
@@ -25,6 +25,8 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	uberzap "go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
```
tkan145: Why do we need another logging lib?

borisurbanik:
@tkan145 the required tests on prow are failing due to CI already using go 1.25

```
# golang.org/x/tools/internal/tokeninternal
/go/pkg/mod/golang.org/x/tools@v0.16.1/internal/tokeninternal/tokeninternal.go:78:9: invalid array length -delta * delta (constant -256 of type int64)
```

Other than this the PR is ready for review. I will rebase this on master when https://github.com/3scale/3scale-operator/pull/1173 is merged to clear the CI errors.

controllers/apps/apimanager_controller.go
```diff
@@ -160,7 +160,7 @@ func (r *APIManagerReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return res, nil
 	}
 
-	specResult, specErr := r.reconcileAPIManagerLogic(instance)
+	specResult, specErr := r.reconcileAPIManagerLogic(r.WithRequest(req), instance)
```
tkan145: We are basically construct a new BaseReconciler just to add a new logger? Does not make a lot of sense to me.

If you want to log the namespace then pass in a logger that we construct at the top of the function
```
logger := r.BaseReconciler.Logger().WithValues("apimanager", req.NamespacedName)
```
Or take the namespace from apiManager when calling `NewBaseAPIManagerLogicReconciler`
```
func NewBaseAPIManagerLogicReconciler(b *reconcilers.BaseReconciler, apiManager *appsv1alpha1.APIManager) *BaseAPIManagerLogicReconciler {
	return &BaseAPIManagerLogicReconciler{
		BaseReconciler:       b,
		apiManager:           apiManager,
		logger:               b.Logger().WithValues("APIManager Controller", apiManager.Name),
		crdAvailabilityCache: &baseAPIManagerLogicReconcilerCRDAvailabilityCache{},
	}
}
```

pkg/3scale/amp/component/apicast.go
```diff
@@ -139,36 +140,49 @@ func (apicast *Apicast) StagingDeployment(ctx context.Context, k8sclient client.
...
-					ServiceAccountName: "amp",
-					Volumes:            apicast.stagingVolumes(),
+					Affinity:                      apicast.Options.StagingAffinity,
+					Tolerations:                   apicast.Options.StagingTolerations,
+					ServiceAccountName:            "amp",
+					RestartPolicy:                 v1.RestartPolicyAlways,
+					DNSPolicy:                     v1.DNSClusterFirst,
+					SecurityContext:               &v1.PodSecurityContext{},
+					TerminationGracePeriodSeconds: ptr.To(int64(v1.DefaultTerminationGracePeriodSeconds)),
+					SchedulerName:                 v1.DefaultSchedulerName,
```
tkan145: I just want to point out that most of the additional fields are purely cosmetic and do not affect how the reconcilers works; ApicastReconciler does not mutate any of these extra fields. Adding these fields and claiming that it reduces the number of reconcile calls is incorrect.

pkg/3scale/amp/component/apicast.go
```diff
@@ -219,15 +233,23 @@ func (apicast *Apicast) ProductionDeployment(ctx context.Context, k8sclient clie
...
-					ServiceAccountName: "amp",
-					Volumes:            apicast.productionVolumes(),
+					Affinity:                      apicast.Options.ProductionAffinity,
+					Tolerations:                   apicast.Options.ProductionTolerations,
+					ServiceAccountName:            "amp",
+					RestartPolicy:                 v1.RestartPolicyAlways,
+					DNSPolicy:                     v1.DNSClusterFirst,
+					SecurityContext:               &v1.PodSecurityContext{},
+					TerminationGracePeriodSeconds: ptr.To(int64(v1.DefaultTerminationGracePeriodSeconds)),
+					SchedulerName:                 v1.DefaultSchedulerName,
```
tkan145: These are redundant.

pkg/3scale/amp/component/backend_redis_tls_test.go
```diff
@@ -163,7 +164,7 @@ func TestBackendComponentRedisTLSVolumes(t *testing.T) {
 				BackendRedisTLS:       TLSConfig{Enabled: false},
 				BackendRedisQueuesTLS: TLSConfig{Enabled: false},
 			},
-			[]v1.Volume{},
```
tkan145: why?

pkg/3scale/amp/component/backend.go
```diff
@@ -725,7 +756,7 @@ func (backend *Backend) backendListenerRunArgs() []string {
 }
 
 func (backend *Backend) backendContainerVolumeMounts() []v1.VolumeMount {
-	res := []v1.VolumeMount{}
```
tkan145, line 728: Any reason why we need to change this? Seems a bit weird to only change the format here and leave everything else the same.

pkg/reconcilers/base_reconciler.go
```diff
@@ -67,6 +67,13 @@ func (b *BaseReconciler) Reconcile(ctx context.Context, req reconcile.Request) (
 	return reconcile.Result{}, nil
 }
 
+// WithRequest returns a shallow copy with a logger enriched with the request's namespace/name.
+func (b *BaseReconciler) WithRequest(req reconcile.Request) *BaseReconciler {
+	c := *b
+	c.logger = b.logger.WithValues("namespace", req.Namespace, "name", req.Name)
+	return &c
+}
+
```
tkan145: See above

test/integration/verify_no_perpetual_reconciliation_test.go
```diff
@@ -0,0 +1,76 @@
...
+	for _, n := range names {
+		fmt.Fprintf(w, "  %s: %d\n", n, updateCounts[n])
+	}
+	fmt.Fprintf(w, "=============================================================\n\n")
+
+	Expect(totalUpdates).To(BeNumerically(">=", minTotalDeploymentUpdates),
+		fmt.Sprintf("%s/%s: total deployment updates is 0 — counter is likely misconfigured (namespace/name fields not captured)", namespace, name))
+	Expect(totalUpdates).To(BeNumerically("<=", maxTotalDeploymentUpdates),
+		deploymentUpdateDetail(namespace, name, updateCounts, totalUpdates))
+}
```
tkan145: I really dislike this test; it's like we don't know how many times the reconciliation will happen, so we have to guess and hope the result falls within the acceptable range. It also doesn't tell us much except that the reconcile function was called x times, which function called it, and what caused that call are still unknown.

tkan145 (changes requested):

controllers/apps/apimanager_controller.go
borisurbanik: I have removed the logger modification, this was a leftover change from when I was trying to untangle the context of the receiver between per-request and per-controller. I've dropped that since and opened https://redhat.atlassian.net/browse/THREESCALE-14828.

pkg/3scale/amp/component/apicast.go
borisurbanik: Thanks for the review - this is a good point. I've removed fields that don't have a mutator across the PR.

test/integration/verify_no_perpetual_reconciliation_test.go
borisurbanik: I've updated to expect exactly 0 updates at reconcile and exactly 1 update after the part where update is triggered. Does this look better?

pkg/3scale/amp/component/backend.go
borisurbanik: There is a mutator for the volume that does DeepEqual and detects this as a change - we need to create object that has `nil` instead of empty array because that what Apiserver does...

pkg/reconcilers/base_reconciler.go
borisurbanik: Removed

controllers/apps/apimanager_controller.go
```diff
@@ -160,7 +160,7 @@ func (r *APIManagerReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return res, nil
 	}
 
-	specResult, specErr := r.reconcileAPIManagerLogic(instance)
+	specResult, specErr := r.reconcileAPIManagerLogic(r.BaseReconciler, instance)
```
tkan145: What are we trying to do here?

test/integration/verify_no_perpetual_reconciliation_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	names := make([]string, 0, len(updateCounts))
+	for n := range updateCounts {
+		names = append(names, n)
+	}
+	sort.Strings(names)
+	for _, n := range names {
+		fmt.Fprintf(w, "  %s: %d\n", n, updateCounts[n])
+	}
+	fmt.Fprintf(w, "==============================================\n\n")
```
tkan145: Is it for debug purpose? We do the same thing with deploymentUpdateDetail below.

tkan145 (commented):
Code wise, it's look ok to me, however I still see APIcast deployment are being update frequently, is it not part of this PR?
```
2026-07-09T14:48:29+10:00	INFO	controllers.APIManager	Updated object 'v1.Deployment/apicast-staging'
2026-07-09T14:48:29+10:00	INFO	controllers.APIManager	Updated object 'v1.Deployment/apicast-production'
```

I'm still not sure about the counting test; let me think about it a little more.

controllers/apps/apimanager_controller.go
borisurbanik: I'll revert that, this is probably a leftover from refactor that has been reverted in one of the commits. There are some structural issues with the controller, but I've opened a separate issue to describe it - https://redhat.atlassian.net/browse/THREESCALE-14828

test/integration/verify_no_perpetual_reconciliation_test.go
borisurbanik: Yeah, that's just for debug / reporting purposes. It's mostly just to follow the same pattern from other code in the test, each step has an informational message about what's being tested using GinkgoWriter. 

The func below is slightly different formatting and it's used for detail of a failed assertion.

borisurbanik:
> Code wise, it's look ok to me, however I still see APIcast deployment are being update frequently, is it not part of this PR?
> 
> ```
> 2026-07-09T14:48:29+10:00	INFO	controllers.APIManager	Updated object 'v1.Deployment/apicast-staging'
> 2026-07-09T14:48:29+10:00	INFO	controllers.APIManager	Updated object 'v1.Deployment/apicast-production'
> ```
> 
> I'm still not sure about the counting test; let me think about it a little more.

Thank you for the review!

I'm not sure why would it be updating apicast deployment, I tried to provision it now and it's not showing up any updates to apicast, must be something that only gets mutated in a specific configuration - do you think you can reconstruct what was the apimanager CR state at the time? The other "spammy" logs were from HPA which has been merged some time ago - https://github.com/3scale/3scale-operator/pull/1175

briangallagher:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/application-services-red-hat-openshift-api-management/pr-checks/f25fd3f7-eb68-418f-b704-9024e46bb21b?refs=c59a4c883233e283a57adb5deef653b6d55124cc&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/application-services-red-hat-openshift-api-management/pr-checks/f25fd3f7-eb68-418f-b704-9024e46bb21b/license?refs=c59a4c883233e283a57adb5deef653b6d55124cc&source=prChecksComment) |

test/integration/verify_no_perpetual_reconciliation_test.go
tkan145: The code below doing the exactly the same thing, so I don't see any good reason to have the same code in multiple places just so we have a different report format. And I think it's just noise; I'm only interested in the logs when something is wrong, not the debugging information output for each run.

tkan145 (approved):

PR #1177, opened 2026-05-07
THREESCALE-14654 Fix status reconciler requeue logic

borisurbanik:
Fixes https://issues.redhat.com/browse/THREESCALE-14654

## Summary

- Fixes `equalStatus && newAvailable` guard: the unavailable-but-equal case fell through to a no-op status write on every reconcile. Changed to `if equalStatus` which short-circuits both available and unavailable steady states.
- Switches both unavailable return paths from `Requeue: true` (immediate tight-loop) to `RequeueAfter: 30s`, giving components time to recover between checks and avoiding extending the backoff counter.

## Test plan

- [x] Unit tests: `go test ./controllers/apps/... -run TestAPIManagerStatusReconciler`
- [x] `TestAPIManagerStatusReconciler_Reconcile_requeueAfterWhenUnavailable` verifies RequeueAfter is non-zero on True→False transition

## Manual testing setup

Tested manually with following setup:

```shell
export NAMESPACE=3scale-test
make NAMESPACE=$NAMESPACE cluster/prepare/local

cat << EOF | oc create -f -
kind: Secret
apiVersion: v1
metadata:
  name: s3-credentials
  namespace: $NAMESPACE
data:
  AWS_BUCKET: c29tZXRoaW5nCg==
  AWS_REGION: dXMtd2VzdC0xCg==
type: Opaque
EOF

cat << EOF | oc create -f -
kind: APIManager
metadata:
  name: 3scale
  namespace: $NAMESPACE
spec:
  wildcardDomain: $DOMAIN
  system:
    fileStorage:
      simpleStorageService:
        configurationSecretRef:
          name: s3-credentials
  externalComponents:
    backend:
      redis: true
    system:
      database: true
      redis: true
EOF

sleep 120 && make run
```

After system provisions and stabilises, delete one of the route, to induce "Available: false", observe the logs for some time - the reconcile was triggered every 30s. 

🤖 Co-authored with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/controllers/apps/apimanager_controller.go
+++ b/controllers/apps/apimanager_controller.go
@@ -175,7 +175,7 @@ func (r *APIManagerReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		return specResult, nil
 	}
 
-	if statusResult.Requeue {
+	if statusResult.Requeue || statusResult.RequeueAfter > 0 {
 		logger.Info("Reconciling not finished. Requeueing.")
 		return statusResult, nil
 	}
--- a/controllers/apps/apimanager_controller.go
+++ b/controllers/apps/apimanager_controller.go
@@ -108,7 +108,7 @@ func (r *APIManagerReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		if statusErr != nil {
 			return ctrl.Result{}, statusErr
 		}
-		if statusResult.Requeue {
+		if statusResult.Requeue || statusResult.RequeueAfter > 0 {
 			logger.Info("Reconciling not finished. Requeueing.")
 			return statusResult, nil
 		}
@@ -138,7 +138,7 @@ func (r *APIManagerReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 				if statusErr != nil {
 					return ctrl.Result{}, statusErr
 				}
-				if statusResult.Requeue {
+				if statusResult.Requeue || statusResult.RequeueAfter > 0 {
 					logger.Info("Reconciling not finished. Requeueing.")
 					return statusResult, nil
 				}
@@ -175,7 +175,7 @@ func (r *APIManagerReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		return specResult, nil
 	}
 
-	if statusResult.Requeue {
+	if statusResult.Requeue || statusResult.RequeueAfter > 0 {
 		logger.Info("Reconciling not finished. Requeueing.")
 		return statusResult, nil
 	}
--- a/controllers/apps/apimanager_status_reconciler.go
+++ b/controllers/apps/apimanager_status_reconciler.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"time"
 
 	appsv1alpha1 "github.com/3scale/3scale-operator/apis/apps/v1alpha1"
 	subController "github.com/3scale/3scale-operator/controllers/subscription"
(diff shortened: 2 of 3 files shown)


controllers/apps/apimanager_status_reconciler.go
```diff
@@ -49,16 +50,15 @@ func (s *APIManagerStatusReconciler) Reconcile() (reconcile.Result, error) {
...
 	newAvailable := newStatus.Conditions.IsTrueFor(appsv1alpha1.APIManagerAvailableConditionType)
 
 	equalStatus := s.apimanagerResource.Status.Equals(newStatus, s.logger)
 	s.logger.V(1).Info("Status", "status is different", !equalStatus)
-	if equalStatus && newAvailable {
-		// Steady state
+	if equalStatus {
 		s.logger.V(1).Info("Status was not updated")
+		if !newAvailable {
+			return reconcile.Result{RequeueAfter: 30 * time.Second}, nil
```
tkan145: As discussed, we now return `reconcile.Result{RequeueAfter: 30 * time.Second}, nil`,  but inside apimanager_controller.go we only check Requeue, so the result is ignored.

```
if statusResult.Requeue {
	logger.Info("Reconciling not finished. Requeueing.")
	return statusResult, nil
}
```

tkan145 (changes requested):

controllers/apps/apimanager_controller.go
```diff
@@ -175,7 +175,7 @@ func (r *APIManagerReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return specResult, nil
 	}
 
-	if statusResult.Requeue {
```
tkan145, line 178: We only check the last one? `reconcileAPIManagerStatus` is called more than once

tkan145 (changes requested):

controllers/apps/apimanager_status_reconciler.go
```diff
@@ -49,33 +50,22 @@ func (s *APIManagerStatusReconciler) Reconcile() (reconcile.Result, error) {
...
-	if updateErr != nil {
-		// Ignore conflicts, resource might just be outdated.
-		if errors.IsConflict(updateErr) {
-			s.logger.Info("Failed to update status: resource might just be outdated")
-			return reconcile.Result{Requeue: true}, nil
+	if !equalStatus {
+		s.logger.V(1).Info("Status is different")
+		s.apimanagerResource.Status = *newStatus
+		if err := s.Client().Status().Update(s.Context(), s.apimanagerResource); err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to update status: %w", err)
```
tkan145: We now ignore the Conflict error?

tkan145 (changes requested):

briangallagher:
### :warning: **Snyk checks are incomplete.**

|---|---|---|---|---|---|---|
| :warning: | Open Source Security| 0 | 0 | 0 | 0 | [ See details ](https://app.snyk.io/org/application-services-red-hat-openshift-api-management/pr-checks/bc666959-5ebf-41ba-a0e9-c7b1e147ae4c?refs=e98339d837dd48e0df3724ae11175cd2c2fa07ec&source=prChecksComment) |

controllers/apps/apimanager_status_reconciler.go
borisurbanik: Fixed up checks in apimanager_controller.go, there are 3 places where status result is checked now.

controllers/apps/apimanager_controller.go
borisurbanik: Added 2 more for the early return branches, thanks for noticing!

controllers/apps/apimanager_status_reconciler.go
borisurbanik: I've restored the missing Conflict error check. I believe that code path is unreachable now (I mean in the current codebase, not after this change) that's why I removed it originally. It's better to leave it for consistency.

tkan145:
Reading the code again, I'm not sure what are we trying to solve here. If for whatever reason, we get into the unavailable-but-equal case, and if the ready status can't not be fixed and stay in unavailable state, with the current code, it will reconcile forever every 30s for no reason.

Also if status is equal but the `Available` is false, then the if branch will be ignore then it will get to the next section of the code and requeue, isn't it? 
```
if equalStatus && newAvailable { }  => true & false => false
```

If the available state is updated, the status equality check will return false and trigger the status update process.

borisurbanik:
> Reading the code again, I'm not sure what are we trying to solve here. If for whatever reason, we get into the unavailable-but-equal case, and if the ready status can't not be fixed and stay in unavailable state, with the current code, it will reconcile forever every 30s for no reason.
> 
> Also if status is equal but the `Available` is false, then the if branch will be ignore then it will get to the next section of the code and requeue, isn't it?
> 
> ```
> if equalStatus && newAvailable { }  => true & false => false
> ```
> 
> If the available state is updated, the status equality check will return false and trigger the status update process.

Thanks for looking again. The intent of this PR is to decouple the status check and availability check to separate conditions to make it easier to reason about it. It would also avoid one no-op `Status().Update()` when the status is not changed.

The requeue reduction to 30s was intended to speed up reconciliation when the backoff counter of the default requeue is already accumulated - again not a clear improvement. 

If you prefer leaving the current code in to avoid potential regression it's very understandable and I'm happy to close this PR.

PR #1178, opened 2026-05-08
Chore: fix ci workflow 

tkan145:
## What
* Bump builder image to go1.25
* Update GH action and pin version by using sha
* Ignore circleci test on master and all PR to master

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,5 +11,8 @@ jobs:
 workflows:
   version: 2
   noop:
+    when:
+      not:
+        equal: [ master, << pipeline.git.branch >> ]
     jobs:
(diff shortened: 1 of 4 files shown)


borisurbanik (approved):
No problem with the PR, approved. 

One question: 
> Update GH action and pin version by using sha
Would this not create maintenance burden / how were these `sha` pinned, is there a cli support for that in `gh` or something like that?

tkan145:
It's security practice to pin with SHA, see https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions 

Regarding maintenance, we probably only need to upgrade the action once a year at most, unless something really broken.

PR #1181, opened 2026-06-25
THREESCALE-10236 Custom CA certificate support for operator deployment

borisurbanik:
- https://redhat.atlassian.net/browse/THREESCALE-10236

The 3scale operator connects to the 3scale Admin API (and, via the Tenant controller, the Master API) to reconcile capabilities CRs (Backend, Product, Application, ActiveDoc, etc.). These connections use HTTPS.

In environments where 3scale is deployed with internal or self-signed CA, the operator's HTTP client rejects the connection because the CA is not in the system trust store. The existing workarounds — the `insecure_skip_verify` annotation and patching the operator Subscription to mount a CA via `SSL_CERT_FILE` ([Red Hat Solution 7049968](https://access.redhat.com/solutions/7049968)) — are either insecure or fragile (overwritten on upgrades).

Customers need a way to provide a custom CA bundle so the operator can trust internal CAs without disabling verification.

Sources of 3scale connections (each with different credential configuration):

1. **Explicit `providerAccountRef`**: the CR references a Secret containing `adminURL` and `token`.
2. **Default `threescale-provider-account` secret**: a well-known Secret name. Same schema as source 1 but discovered by convention.
3. **Local APIManager**: the operator discovers an APIManager CR in the namespace and derives the admin URL and token from `system-seed`.
4. **Tenant controller / Master API**: the Tenant CR references a `masterCredentialsRef` secret (defaulting to `system-seed`) containing master API credentials.

***This PR adds ability to configure a custom CA bundle when connecting to any of the above endpoints. The bundle will be loaded from a well-known configmap from the operator namespace (`threescale-ca-bundle`) and will be reloaded automatically if a change is detected.***

### Testing:

The testing follows 3 scenarios when using a standard deployment of APIManager on creating a backend resource without secret -> the backend reconciler will fall back to endpoint for local APIManager. 

1. make sure the request to 3scale admin URL is failing initially:

```
oc apply -n $NAMESPACE -f - <<EOF
kind: Backend
metadata:
  name: test-backend-noconfig
spec:
  name: "Test Backend No Config"
  systemName: "test-backend-noconfig"
  privateBaseURL: "https://httpbin.org"
EOF

# Wait a moment for reconcile, then check status
oc get backend test-backend-noconfig -n $NAMESPACE -o jsonpath='{.status.conditions}' | jq

# Check for the Warning event
oc get events -n $NAMESPACE --field-selector reason=ReconcileError,involvedObject.name=test-backend-noconfig
```

<details>
<summary>Result:</summary>

```
borisurbanik@MacBookPro rook_ceph_deploy % oc get backend test-backend-noconfig -n $NAMESPACE -o jsonpath='{.status.conditions}' | jq
[
  {
    "lastTransitionTime": "2026-06-30T22:28:51Z",
    "message": "Get \"https://3scale-admin.apps.burbanik-3scale.cp.fyre.ibm.com:443/admin/api/backend_apis.json?page=1&per_page=500\": tls: failed to verify certificate: x509: certificate signed by unknown authority",
    "status": "True",
    "type": "Failed"
  },
  {
    "lastTransitionTime": "2026-06-30T22:28:51Z",
    "status": "False",
    "type": "Invalid"
  },
  {
    "lastTransitionTime": "2026-06-30T22:28:51Z",
    "status": "False",
    "type": "Synced"
  }
]
borisurbanik@MacBookPro rook_ceph_deploy % oc get events -n $NAMESPACE --field-selector reason=ReconcileError,involvedObject.name=test-backend-noconfig
LAST SEEN   TYPE      REASON           OBJECT                          MESSAGE
5s          Warning   ReconcileError   backend/test-backend-noconfig   Get "https://3scale-admin.apps.burbanik-3scale.cp.fyre.ibm.com:443/admin/api/backend_apis.json?page=1&per_page=500": tls: failed to verify certificate: x509: certificate signed by unknown authority
```

</details>

2. Check that the annotation `insecure_skip_verify` would bypass the problem:

```
oc apply -n $NAMESPACE -f - <<EOF
kind: Backend
metadata:
  name: test-backend-insecure
  annotations:
    insecure_skip_verify: "true"
spec:
  name: "Test Backend Insecure"
  systemName: "test-backend-insecure"
  privateBaseURL: "https://httpbin.org"
EOF

oc wait --for=condition=Synced --timeout=60s backend/test-backend-insecure -n $NAMESPACE
oc get backend test-backend-insecure -n $NAMESPACE -o jsonpath='{.status.conditions}' | jq
```

<details>
<summary>Result:</summary>

```
borisurbanik@MacBookPro rook_ceph_deploy % oc wait --for=condition=Synced --timeout=60s backend/test-backend-insecure -n $NAMESPACE
backend.capabilities.3scale.net/test-backend-insecure condition met
borisurbanik@MacBookPro rook_ceph_deploy % oc get backend test-backend-insecure -n $NAMESPACE -o jsonpath='{.status.conditions}' | jq
[
  {
    "lastTransitionTime": "2026-06-30T22:31:26Z",
    "status": "False",
    "type": "Failed"
  },
  {
    "lastTransitionTime": "2026-06-30T22:31:26Z",
    "status": "False",
    "type": "Invalid"
  },
  {
    "lastTransitionTime": "2026-06-30T22:31:26Z",
    "status": "True",
    "type": "Synced"
  }
]
```

</details>

3. Check that custom CA will work as well

```
# Extract the ingress CA and create the ConfigMap
oc get secret router-ca -n openshift-ingress-operator \
  -o jsonpath='{.data.tls\.crt}' | base64 -d > /tmp/ingress-ca.crt

oc create configmap threescale-ca-bundle \
  --from-file=ca-bundle.crt=/tmp/ingress-ca.crt \
  -n $NAMESPACE

# Wait for the CABundleWatcher to reconcile (it's non-leader-election so picks up immediately)
sleep 3

oc apply -n $NAMESPACE -f - <<EOF
kind: Backend
metadata:
  name: test-backend-ca
spec:
  name: "Test Backend With CA"
  systemName: "test-backend-ca"
  privateBaseURL: "https://httpbin.org"
EOF

oc wait --for=condition=Synced --timeout=60s backend/test-backend-ca -n $NAMESPACE
oc get backend test-backend-ca -n $NAMESPACE -o jsonpath='{.status.conditions}' | jq
```

<details>
<summary>Result:</summary>

```
borisurbanik@MacBookPro rook_ceph_deploy % oc wait --for=condition=Synced --timeout=60s backend/test-backend-ca -n $NAMESPACE
backend.capabilities.3scale.net/test-backend-ca condition met
borisurbanik@MacBookPro rook_ceph_deploy % oc get backend test-backend-ca -n $NAMESPACE -o jsonpath='{.status.conditions}' | jq
[
  {
    "lastTransitionTime": "2026-06-30T22:34:32Z",
    "status": "False",
    "type": "Failed"
  },
  {
    "lastTransitionTime": "2026-06-30T22:34:32Z",
    "status": "False",
    "type": "Invalid"
  },
  {
    "lastTransitionTime": "2026-06-30T22:34:32Z",
    "status": "True",
    "type": "Synced"
  }
]
```

</details>

4. Test that invalid bundle will surface errors on the config map itself

```
oc apply -n $NAMESPACE -f - <<EOF
apiVersion: v1
kind: ConfigMap
metadata:
  name: threescale-ca-bundle
data:
  ca-bundle.crt: "this is not a valid PEM certificate"
EOF

sleep 3

oc get events -n $NAMESPACE \
  --field-selector reason=InvalidCABundle,involvedObject.name=threescale-ca-bundle
```

<details>

<summary>Result:</summary>

```
borisurbanik@MacBookPro rook_ceph_deploy % oc get events -n $NAMESPACE \
  --field-selector reason=InvalidCABundle,involvedObject.name=threescale-ca-bundle
LAST SEEN   TYPE      REASON            OBJECT                           MESSAGE
17s         Warning   InvalidCABundle   configmap/threescale-ca-bundle   InvalidCAFormat: No valid PEM-encoded certificates found in CA bundle
```

</details>

6. Test bring-your-own-ca + Test bundle with multiple CAs

```
echo "=== Step 1: Generate self-signed CA ==="
openssl genrsa -out /tmp/tc2-ca.key 2048
openssl req -x509 -new -nodes -key /tmp/tc2-ca.key -sha256 -days 365 \
  -out /tmp/tc2-ca.crt -subj "/CN=TestCA-TC2"

echo ""
echo "=== Step 1b: Generate server cert signed by TestCA-TC2 and patch route ==="
openssl genrsa -out /tmp/tc2-server.key 2048
openssl req -new -key /tmp/tc2-server.key \
  -out /tmp/tc2-server.csr \
  -subj "/CN=${ROUTE_HOST}" \
  -addext "subjectAltName=DNS:${ROUTE_HOST}"
openssl x509 -req -in /tmp/tc2-server.csr \
  -CA /tmp/tc2-ca.crt -CAkey /tmp/tc2-ca.key -CAcreateserial \
  -out /tmp/tc2-server.crt -days 365 -sha256 \
  -extfile <(echo "subjectAltName=DNS:${ROUTE_HOST}")
openssl verify -CAfile /tmp/tc2-ca.crt /tmp/tc2-server.crt

oc patch route "${ROUTE_NAME}" -n "${APIMANAGER_NS}" \
  --type=merge -p "{
    \"spec\": {
      \"tls\": {
        \"termination\": \"edge\",
        \"insecureEdgeTerminationPolicy\": \"Redirect\",
        \"certificate\": $(cat /tmp/tc2-server.crt | python3 -c 'import sys,json; print(json.dumps(sys.stdin.read()))'),
        \"key\": $(cat /tmp/tc2-server.key | python3 -c 'import sys,json; print(json.dumps(sys.stdin.read()))')
      }
    }
  }"

echo "=== Step 2: Concatenate second CA and update ConfigMap in-place ==="
openssl genrsa -out /tmp/tc2-ca2.key 2048
openssl req -x509 -new -nodes -key /tmp/tc2-ca2.key -sha256 -days 365 \
  -out /tmp/tc2-ca2.crt -subj "/CN=TestCA-TC2-Second"
cat /tmp/tc2-ca.crt /tmp/tc2-ca2.crt > /tmp/tc2-bundle.crt
oc create configmap threescale-ca-bundle \
  --from-file=ca-bundle.crt=/tmp/tc2-bundle.crt -n "${OPERATOR_NS}" \
  --dry-run=client -o yaml | oc apply -f -

echo "=== Step 3: Verify no warnings after update ==="
sleep 10
echo "--- Certificate blocks in ConfigMap ---"
oc get configmap threescale-ca-bundle \
  -o jsonpath='{.data.ca-bundle\.crt}' -n "${OPERATOR_NS}" | grep -c "BEGIN CERTIFICATE"
echo "--- InvalidCABundle events ---"
oc get events -n "${OPERATOR_NS}" --field-selector reason=InvalidCABundle --sort-by='.lastTimestamp'

echo ""
echo "=== Step 4: Trigger reconcile and verify Synced=True ==="
oc annotate backend "${BACKEND_NAME}" -n "${APIMANAGER_NS}" \
  test-reconcile-trigger="$(date -u +%Y%m%dT%H%M%SZ)" --overwrite
sleep 15
echo "--- Backend conditions ---"
oc get backend "${BACKEND_NAME}" -n "${APIMANAGER_NS}" \
  -o jsonpath='{range .status.conditions[*]}{.type}{"\t"}{.status}{"\t"}{.message}{"\n"}{end}'
```

<details>

<summary>Result:</summary>

```
=== Step 3: Verify no warnings after update ===

--- Certificate blocks in ConfigMap ---
2
--- InvalidCABundle events ---
LAST SEEN   TYPE      REASON            OBJECT                           MESSAGE
84m         Warning   InvalidCABundle   configmap/threescale-ca-bundle   InvalidCAFormat: No valid PEM-encoded certificates found in CA bundle
76m         Warning   InvalidCABundle   configmap/threescale-ca-bundle   InvalidCAFormat: No valid PEM-encoded certificates found in CA bundle
34m         Warning   InvalidCABundle   configmap/threescale-ca-bundle   InvalidCAFormat: No valid PEM-encoded certificates found in CA bundle
32m         Warning   InvalidCABundle   configmap/threescale-ca-bundle   InvalidCAFormat: No valid PEM-encoded certificates found in CA bundle
27m         Warning   InvalidCABundle   configmap/threescale-ca-bundle   InvalidCAFormat: No valid PEM-encoded certificates found in CA bundle
24m         Warning   InvalidCABundle   configmap/threescale-ca-bundle   InvalidCAFormat: No valid PEM-encoded certificates found in CA bundle
21m         Warning   InvalidCABundle   configmap/threescale-ca-bundle   InvalidCAFormat: No valid PEM-encoded certificates found in CA bundle

=== Step 4: Trigger reconcile and verify Synced=True ===
backend.capabilities.3scale.net/test-backend-ca annotated
--- Backend conditions ---
Failed	False
Invalid	False
Synced	True
```

</details>

7. Test setting up ca bundle after upgrade from 2.16.3 using OLM

Tested using live cluster with following configuration (with catalog image built from the PR branch): 

```
   oc apply -f - <<'EOF'
   apiVersion: operators.coreos.com/v1alpha1
   kind: CatalogSource
   metadata:
     name: threescale-dev
     namespace: openshift-marketplace
   spec:
     sourceType: grpc
     image: quay.io/burbanik/3scale-operator-catalog:v0.14.0
   EOF

   oc apply -f - <<'EOF'
   apiVersion: operators.coreos.com/v1
   kind: OperatorGroup
   metadata:
     name: threescale-og
     namespace: 3scale-test
   spec:
     targetNamespaces:
     - 3scale-test
   EOF

   oc apply -f - <<'EOF'
   apiVersion: operators.coreos.com/v1alpha1
   kind: Subscription
   metadata:
     name: 3scale-operator
     namespace: 3scale-test
   spec:
     source: threescale-dev
     sourceNamespace: openshift-marketplace
     name: 3scale-operator
     channel: threescale-2.16
     startingCSV: 3scale-operator.v0.13.3
     installPlanApproval: Manual
   EOF


cat <<'EOF' | oc apply -n 3scale-test -f -
  kind: Backend
  metadata:
    name: test-backend
    namespace: 3scale-test
  spec:
    name: "Test Backend"
    privateBaseURL: "https://echo-api.3scale.net:443"
  EOF)

# at this point the sync should be failing

# upgrade to 2.17 and create router-ca bundle

oc get secret router-ca -n openshift-ingress-operator -o jsonpath='{.data.tls\.crt}' | base64 -d > /tmp/ingress-ca.crt
oc create configmap threescale-ca-bundle --from-file=ca-bundle.crt=/tmp/ingress-ca.crt -n 3scale-test
```

<details>

<summary>Result:</summary>

Backend status before upgrade: 

```
borisurbanik@MacBookPro 3scale-operator % oc get backend test-backend -o json | jq -r '.status.conditions'
[
  {
    "lastTransitionTime": "2026-07-02T21:31:39Z",
    "message": "Get \"https://3scale-admin.apps.burbanik-3scale.cp.fyre.ibm.com:443/admin/api/backend_apis.json?page=1&per_page=500\": tls: failed to verify certificate: x509: certificate signed by unknown authority",
    "status": "True",
    "type": "Failed"
  },
  {
    "lastTransitionTime": "2026-07-02T21:31:39Z",
    "status": "False",
    "type": "Invalid"
  },
  {
    "lastTransitionTime": "2026-07-02T21:31:39Z",
    "status": "False",
    "type": "Synced"
  }
]
```

After upgrade with custom CA:

```
borisurbanik@MacBookPro 3scale-operator % oc get backend test-backend -o json | jq -r '.status.conditions'
[
  {
    "lastTransitionTime": "2026-07-02T21:53:44Z",
    "status": "False",
    "type": "Failed"
  },
  {
    "lastTransitionTime": "2026-07-02T21:31:39Z",
    "status": "False",
    "type": "Invalid"
  },
  {
    "lastTransitionTime": "2026-07-02T21:53:44Z",
    "status": "True",
    "type": "Synced"
  }
]
```

</details>

### DIFF
--- a/controllers/configuration/ca_bundle_watcher.go
+++ b/controllers/configuration/ca_bundle_watcher.go
@@ -1,0 +1,128 @@
+/*
+Copyright 2026 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package configuration
+
+import (
+	"context"
+	"crypto/x509"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/utils/ptr"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+const (
+	CABundleConfigMapName = "threescale-ca-bundle"
+	CABundleConfigMapKey  = "ca-bundle.crt"
+)
+
+// CABundleWatcher watches the threescale-ca-bundle ConfigMap and updates the
+// package-level CA pool via SetRootCAs on every successful reconcile.
+type CABundleWatcher struct {
+	client.Client
+	Recorder  record.EventRecorder
+	Namespace string
+}
+
+// SetupWithManager registers the controller with the manager.
+func (r *CABundleWatcher) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("cabundlewatcher").
+		// NeedLeaderElection is disabled so that every operator replica runs
+		// this controller, not just the leader.  Each replica must load the CA
+		// bundle into its own in-process TLS config.
+		//
+		// controller-runtime replays a synthetic Add event for every object
+		// already in the informer cache when a controller starts, so a newly
+		// promoted leader would pick up the bundle automatically.  Running on
+		// all replicas avoids the small window between manager start and leader
+		// promotion during which capability reconcilers could attempt outbound
+		// TLS calls before the bundle has been loaded.  Those calls would fail
+		// and requeue, but disabling leader election here eliminates the window
+		// entirely.
+		WithOptions(controller.Options{NeedLeaderElection: ptr.To(false)}).
+		For(&corev1.ConfigMap{}, builder.WithPredicates(predicate.NewPredicateFuncs(func(object client.Object) bool {
+			return object.GetNamespace() == r.Namespace && object.GetName() == CABundleConfigMapName
+		}))).
+		Complete(r)
+}
+
+// Reconcile fetches the CA bundle ConfigMap and updates the package-level
+// TLS config.  On an invalid bundle, the error is logged and recorded as a Warning event on
+// the ConfigMap; the existing TLS config is left unchanged so capability
+// controllers continue to operate with the last known good CA.
+func (r *CABundleWatcher) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := ctrl.LoggerFrom(ctx)
+
+	cm := &corev1.ConfigMap{}
+	err := r.Get(ctx, client.ObjectKey{Namespace: r.Namespace, Name: CABundleConfigMapName}, cm)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			SetRootCAs(nil)
+			logger.Info("CA bundle ConfigMap not found; using system default CAs", "namespace", r.Namespace, "configmap", CABundleConfigMapName)
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	pool, parseErr := parseBundleFromConfigMap(cm)
+	if parseErr != nil {
+		logger.Error(parseErr, "CA bundle ConfigMap contains an invalid certificate bundle; keeping previous TLS config")
+		if r.Recorder != nil {
+			r.Recorder.Eventf(cm, corev1.EventTypeWarning, "InvalidCABundle", "%v", parseErr)
+		}
+		return ctrl.Result{}, nil
+	}
+
+	if pool == nil {
+		SetRootCAs(nil)
+		logger.Info("CA bundle ConfigMap key absent; using system default CAs", "namespace", r.Namespace, "configmap", CABundleConfigMapName, "key", CABundleConfigMapKey)
+		return ctrl.Result{}, nil
+	}
+
+	SetRootCAs(pool)
+	logger.Info("CA bundle updated and applied", "namespace", r.Namespace, "configmap", CABundleConfigMapName)
+	return ctrl.Result{}, nil
+}
+
+// parseBundleFromConfigMap parses the CA bundle from a ConfigMap and returns a
+// *x509.CertPool.  Returns nil, nil if the key is absent (no custom bundle
+// configured).  Returns an error if the PEM data is present but invalid.
+func parseBundleFromConfigMap(cm *corev1.ConfigMap) (*x509.CertPool, error) {
+	val, exists := cm.Data[CABundleConfigMapKey]
+	if !exists {
+		return nil, nil // no bundle configured — use system roots
+	}
+
+	if len(val) == 0 {
+		return nil, fmt.Errorf("InvalidCAFormat: key %q in ConfigMap %s/%s is empty", CABundleConfigMapKey, cm.Namespace, cm.Name)
+	}
+
+	certPool := x509.NewCertPool()
+	if !certPool.AppendCertsFromPEM([]byte(val)) {
+		return nil, fmt.Errorf("InvalidCAFormat: no valid PEM-encoded certificates found in CA bundle")
+	}
+
+	return certPool, nil
+}
--- a/controllers/configuration/ca_bundle_watcher_test.go
+++ b/controllers/configuration/ca_bundle_watcher_test.go
@@ -1,0 +1,316 @@
+/*
+Copyright 2026 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package configuration_test
+
+import (
+	"context"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"errors"
+	"math/big"
+	"strings"
+	"testing"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	fakectrlclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/3scale/3scale-operator/controllers/configuration"
+	"github.com/3scale/3scale-operator/pkg/testhelper"
+)
+
+// errorInjectingClient makes r.Get return a non-NotFound error so the
+// reconciler propagates it to the workqueue.
+type errorInjectingClient struct {
+	client.Client
+	getErr error
+}
+
+func (e *errorInjectingClient) Get(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
+	return e.getErr
+}
+
+// generateSelfSignedCert creates a minimal self-signed RSA certificate and
+// returns its PEM-encoded DER bytes. It is not a real secret and carries no
+// trust outside the test suite.
+func generateSelfSignedCert(t *testing.T, cn string) []byte {
+	t.Helper()
+	key, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		t.Fatalf("generate RSA key: %v", err)
+	}
+	tmpl := &x509.Certificate{
+		SerialNumber:          big.NewInt(1),
+		Subject:               pkix.Name{CommonName: cn},
+		NotBefore:             time.Now().Add(-time.Hour),
+		NotAfter:              time.Now().Add(24 * time.Hour),
+		IsCA:                  true,
+		KeyUsage:              x509.KeyUsageCertSign | x509.KeyUsageCRLSign,
+		BasicConstraintsValid: true,
+	}
+	der, err := x509.CreateCertificate(rand.Reader, tmpl, tmpl, &key.PublicKey, key)
+	if err != nil {
+		t.Fatalf("create certificate: %v", err)
+	}
+
+	return der
+}
+
+func encodePEM(der []byte) []byte {
+	return pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: der})
+}
+
+func TestCABundleWatcher_Reconcile(t *testing.T) {
+	certDER := generateSelfSignedCert(t, "test-ca")
+	certPEM := encodePEM(certDER)
+
+	seedDER := generateSelfSignedCert(t, "test-ca-at-start")
+	seedPool := x509.NewCertPool()
+	seedCert, err := x509.ParseCertificate(seedDER)
+	if err != nil {
+		t.Fatalf("parse seed cert: %v", err)
+	}
+	seedPool.AddCert(seedCert)
+
+	validCM := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{Namespace: "test-ns", Name: configuration.CABundleConfigMapName},
+		Data:       map[string]string{configuration.CABundleConfigMapKey: string(certPEM)},
+	}
+	emptyCM := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{Namespace: "test-ns", Name: configuration.CABundleConfigMapName},
+		Data:       map[string]string{configuration.CABundleConfigMapKey: ""},
+	}
+	invalidCM := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{Namespace: "test-ns", Name: configuration.CABundleConfigMapName},
+		Data:       map[string]string{configuration.CABundleConfigMapKey: "not-a-pem"},
+	}
+	keyAbsentCM := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{Namespace: "test-ns", Name: configuration.CABundleConfigMapName},
+		Data:       map[string]string{},
+	}
+
+	tests := []struct {
+		name         string
+		objects      []runtime.Object
+		seedPool     *x509.CertPool
+		injectGetErr bool
+		wantErr      bool
+		wantErrMsg   string
+		check        func(t *testing.T, rec *record.FakeRecorder)
+	}{
+		{
+			name:     "ConfigMap not found — SetRootCAs(nil), no error, no event",
+			objects:  []runtime.Object{},
+			seedPool: seedPool,
+			wantErr:  false,
+			check: func(t *testing.T, rec *record.FakeRecorder) {
+				if configuration.GetRootCAs() != nil {
+					t.Error("expected GetRootCAs() == nil after not-found")
+				}
+				assertNoEventW(t, rec)
+			},
+		},
+		{
+			name:         "generic Get error — error returned, CAs unchanged, no event",
+			objects:      []runtime.Object{},
+			seedPool:     seedPool,
+			injectGetErr: true,
+			wantErr:      true,
+			wantErrMsg:   "synthetic server error",
+			check: func(t *testing.T, rec *record.FakeRecorder) {
+				pool := configuration.GetRootCAs()
+				if pool == nil {
+					t.Error("expected GetRootCAs() to remain non-nil (seeded value)")
+				} else if !pool.Equal(seedPool) {
+					t.Error("loaded CA pool does not contain the expected certificate")
+				}
+				assertNoEventW(t, rec)
+			},
+		},
+		{
+			name:     "ConfigMap present, key present but empty — Warning event, no error, CAs unchanged",
+			objects:  []runtime.Object{emptyCM},
+			seedPool: seedPool,
+			wantErr:  false,
+			check: func(t *testing.T, rec *record.FakeRecorder) {
+				pool := configuration.GetRootCAs()
+				if pool == nil {
+					t.Error("expected GetRootCAs() to remain non-nil (kept previous config)")
+				} else if !pool.Equal(seedPool) {
+					t.Error("loaded CA pool does not contain the expected certificate")
+				}
+
+				msg := drainOneEventW(t, rec)
+				for _, sub := range []string{"Warning", "InvalidCABundle", "InvalidCAFormat: key \"ca-bundle.crt\" in ConfigMap test-ns/threescale-ca-bundle is empty"} {
+					if !strings.Contains(msg, sub) {
+						t.Errorf("event %q missing expected substring %q", msg, sub)
+					}
+				}
+				assertNoEventW(t, rec)
+			},
+		},
+		{
+			name:     "ConfigMap present, PEM invalid — Warning event, no error, CAs unchanged",
+			objects:  []runtime.Object{invalidCM},
+			seedPool: seedPool,
+			wantErr:  false,
+			check: func(t *testing.T, rec *record.FakeRecorder) {
+				pool := configuration.GetRootCAs()
+				if pool == nil {
+					t.Error("expected GetRootCAs() to remain non-nil (kept previous config)")
+				} else if !pool.Equal(seedPool) {
+					t.Error("loaded CA pool does not contain the expected certificate")
+				}
+				msg := drainOneEventW(t, rec)
+				for _, sub := range []string{"Warning", "InvalidCABundle", "InvalidCAFormat: no valid PEM-encoded certificates found in CA bundle"} {
+					if !strings.Contains(msg, sub) {
+						t.Errorf("event %q missing expected substring %q", msg, sub)
+					}
+				}
+				assertNoEventW(t, rec)
+			},
+		},
+		{
+			name:     "ConfigMap present, key absent — SetRootCAs(nil), no error, no event",
+			objects:  []runtime.Object{keyAbsentCM},
+			seedPool: seedPool,
+			wantErr:  false,
+			check: func(t *testing.T, rec *record.FakeRecorder) {
+				if configuration.GetRootCAs() != nil {
+					t.Error("expected GetRootCAs() == nil after key-absent")
+				}
+				assertNoEventW(t, rec)
+			},
+		},
+		{
+			name:    "ConfigMap present, valid PEM — SetRootCAs(pool), no error, no event",
+			objects: []runtime.Object{validCM},
+			wantErr: false,
+			check: func(t *testing.T, rec *record.FakeRecorder) {
+				pool := configuration.GetRootCAs()
+				if pool == nil {
+					t.Fatal("expected GetRootCAs() != nil after valid PEM")
+				}
+				cert, err := x509.ParseCertificate(certDER)
+				if err != nil {
+					t.Fatalf("parse test certificate: %v", err)
+				}
+				expected := x509.NewCertPool()
+				expected.AddCert(cert)
+				if !pool.Equal(expected) {
+					t.Error("loaded CA pool does not contain the expected certificate")
+				}
+				assertNoEventW(t, rec)
+			},
+		},
+		{
+			name:     "ConfigMap present, valid PEM, replace seed case — SetRootCAs(pool), no error, no event",
+			objects:  []runtime.Object{validCM},
+			seedPool: seedPool,
+			wantErr:  false,
+			check: func(t *testing.T, rec *record.FakeRecorder) {
+				pool := configuration.GetRootCAs()
+				if pool == nil {
+					t.Fatal("expected GetRootCAs() != nil after valid PEM")
+				}
+				cert, err := x509.ParseCertificate(certDER)
+				if err != nil {
+					t.Fatalf("parse test certificate: %v", err)
+				}
+				expected := x509.NewCertPool()
+				expected.AddCert(cert)
+				if !pool.Equal(expected) {
+					t.Error("loaded CA pool does not contain the expected certificate")
+				}
+				assertNoEventW(t, rec)
+			},
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			// Seed a known non-nil pool so "nil after Reconcile" assertions are unambiguous.
+			testhelper.SeedRootCAsForTest(t, tc.seedPool)
+
+			s := runtime.NewScheme()
+			if err := corev1.AddToScheme(s); err != nil {
+				t.Fatalf("AddToScheme corev1: %v", err)
+			}
+
+			cl := fakectrlclient.NewClientBuilder().
+				WithScheme(s).
+				WithRuntimeObjects(tc.objects...).
+				Build()
+
+			var cl2 client.Client = cl
+			if tc.injectGetErr {
+				cl2 = &errorInjectingClient{Client: cl, getErr: errors.New("synthetic server error")}
+			}
+
+			rec := record.NewFakeRecorder(10)
+			watcher := &configuration.CABundleWatcher{
+				Client:    cl2,
+				Recorder:  rec,
+				Namespace: "test-ns",
+			}
+
+			_, err := watcher.Reconcile(context.Background(), reconcile.Request{
+				NamespacedName: types.NamespacedName{Namespace: "test-ns", Name: configuration.CABundleConfigMapName},
+			})
+
+			if (err != nil) != tc.wantErr {
+				t.Errorf("wantErr=%v, got err=%v", tc.wantErr, err)
+			}
+			if tc.wantErrMsg != "" && (err == nil || !strings.Contains(err.Error(), tc.wantErrMsg)) {
+				t.Errorf("expected error containing %q, got %v", tc.wantErrMsg, err)
+			}
+			tc.check(t, rec)
+		})
+	}
+}
+
+// drainOneEventW reads one event from the recorder and returns it.
+// Calls t.Fatal immediately if the channel is empty.
+func drainOneEventW(t *testing.T, rec *record.FakeRecorder) string {
+	t.Helper()
+	select {
+	case msg := <-rec.Events:
+		return msg
+	default:
+		t.Fatal("expected one event on recorder, but channel was empty")
+		return ""
+	}
+}
+
+// assertNoEventW fails the test if any event remains on the recorder.
+func assertNoEventW(t *testing.T, rec *record.FakeRecorder) {
+	t.Helper()
+	select {
+	case msg := <-rec.Events:
+		t.Errorf("unexpected event: %q", msg)
+	default:
+	}
+}
--- a/controllers/configuration/tls_config.go
+++ b/controllers/configuration/tls_config.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2026 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package configuration
+
+import (
+	"crypto/x509"
+	"sync"
+)
+
+// currentRootCAs is the package-level CA pool derived from the most recently
+// reconciled threescale-ca-bundle ConfigMap.
+// nil means "no custom CA bundle" — callers should use system roots.
+var currentRootCAs *x509.CertPool
+
+var tlsConfigMu sync.RWMutex
+
+// SetRootCAs replaces the current package-level CA pool.
+// Pass nil to revert to system roots (e.g. when the ConfigMap is deleted).
+func SetRootCAs(pool *x509.CertPool) {
+	tlsConfigMu.Lock()
+	currentRootCAs = pool
+	tlsConfigMu.Unlock()
+}
+
+// GetRootCAs returns the current package-level CA pool, or nil if no bundle
+// has been configured. Callers construct their own tls.Config and set RootCAs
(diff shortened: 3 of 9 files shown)


pkg/controller/helper/threescale_api.go
```diff
@@ -33,24 +37,36 @@ func PortaClientFromURLString(adminURLStr, token string, insecureSkipVerify bool
...
+		TLSClientConfig: tlsConfig,
 	}
-
 	if helper.GetEnvVar(HTTP_VERBOSE_ENVVAR, "0") == "1" {
 		transport = &helper.Transport{Transport: transport}
 	}
-
-	return threescaleapi.NewThreeScale(adminPortal, token, &http.Client{Transport: transport}), nil
+	return &http.Client{Transport: transport}
 }
```
tkan145: Any reason why we want to split it to another function here?

controllers/configuration/ca_bundle_watcher.go
```diff
@@ -0,0 +1,170 @@
...
+	client.Client
+	Recorder  record.EventRecorder
+	Namespace string
+}
+
+// SetupWithManager registers the controller with the manager.
+func (r *CABundleWatcher) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("cabundlewatcher").
+		WithOptions(controller.Options{NeedLeaderElection: ptr.To(false)}).
```
tkan145, line 64: Why do we need this?

controllers/configuration/ca_bundle_watcher.go
```diff
@@ -0,0 +1,170 @@
...
+					e.ObjectNew.GetNamespace() == r.Namespace
+			},
+			DeleteFunc: func(e event.DeleteEvent) bool {
+				return e.Object.GetName() == CABundleConfigMapName &&
+					e.Object.GetNamespace() == r.Namespace
+			},
+			GenericFunc: func(e event.GenericEvent) bool {
+				return e.Object.GetName() == CABundleConfigMapName &&
+					e.Object.GetNamespace() == r.Namespace
+			},
```
tkan145: You can replace all of this with 
```
CABundleConfigMapPredicate := predicate.NewPredicateFuncs(func(object client.Object) bool {
		return object.GetNamespace() == common.OperatorNamespace &&
			object.GetName() == CABundleConfigMapName
})
```

controllers/configuration/ca_bundle_watcher.go
```diff
@@ -0,0 +1,170 @@
...
+		})).
+		WithLogConstructor(func(_ *reconcile.Request) logr.Logger {
+			return mgr.GetLogger().WithValues("controller", "cabundlewatcher")
+		}).
```
tkan145: Do we need this? iirc, If the logger is nil, it will construct a default one with controller name

controllers/configuration/ca_bundle_watcher.go
```diff
@@ -0,0 +1,170 @@
...
+
+func (e *CAValidationError) Unwrap() error {
+	return e.Err
+}
+
+const (
+	CAValidationReasonMissingSecret = "MissingCASecret"
+	CAValidationReasonMissingKey    = "MissingCAKey"
+	CAValidationReasonInvalidFormat = "InvalidCAFormat"
+)
```
tkan145: Why do we need to wrap the error instead of just return it directly? Also seems like only CAValidationReasonInvalidFormat is being used.

pkg/controller/helper/threescale_api.go
```diff
@@ -33,24 +37,36 @@ func PortaClientFromURLString(adminURLStr, token string, insecureSkipVerify bool
...
-// PortaClientFromURL instantiates porta_client.ThreeScaleClient from admin url object
+// PortaClientFromURL instantiates a ThreeScaleClient from an admin URL
+// and access token.  When insecureSkipVerify is true, the CA bundle is ignored.
 func PortaClientFromURL(url *url.URL, token string, insecureSkipVerify bool) (*threescaleapi.ThreeScaleClient, error) {
+	var httpClient *http.Client
+	if insecureSkipVerify {
+		httpClient = buildHTTPClient(&tls.Config{InsecureSkipVerify: true}) //nolint:gosec
+	} else {
+		httpClient = buildHTTPClient(configuration.GetTLSConfig())
+	}
```
tkan145: Aren't it a lot simpler to write

```
tlsConfig := configuration.GetTLSConfig()
if tlsConfig == nil {
     tlsConfig = &tls.Config{}
}
tlsCOnfig.InsecureSkipVerify = insecureSkipVerify
```

doc/operator-ca-bundle-trust-manager.md
```diff
@@ -0,0 +1,98 @@
+# Combining CA sources with trust-manager
+
+Use this guide when you need the operator to trust both a cluster-managed CA (via OpenShift CA injection) and a separately managed custom CA. Common cases include multiple 3scale instances signed by different CAs, or a remote OpenAPI spec endpoint that uses a CA not in the cluster bundle.
```
tkan145: I don't think we handle the OpenAPI spec enpoint yet, it use different type of client

doc/operator-ca-bundle-trust-manager.md
```diff
@@ -0,0 +1,98 @@
...
+
+Once the `config.openshift.io/inject-trusted-cabundle=true` label is set, the Cluster Network Operator (CNO) owns the ConfigMap and overwrites any manual edits, so you cannot add extra certificates directly. [trust-manager](https://cert-manager.io/docs/trust/trust-manager/) solves this by assembling a merged bundle from multiple sources and writing it to the target ConfigMap.
```
tkan145: This sentence is very confusing. What is Cluster Network Operator (CNO)? and why do user need to use label injection? what if I have a different source for CA? I think we can be a bit more generic here.

doc/operator-ca-bundle-trust-manager.md
```diff
@@ -0,0 +1,98 @@
...
+## Step 1 — Create a ConfigMap for the cluster CA source
+
+Label a ConfigMap in the `cert-manager` namespace (the trust-manager trust namespace) for OpenShift CA injection. trust-manager will read this as one of its sources:
+
+```bash
+kubectl create configmap openshift-ca -n cert-manager
+kubectl label configmap openshift-ca \
+  config.openshift.io/inject-trusted-cabundle=true \
+  -n cert-manager
+```
```
tkan145: Do we really need this?

doc/operator-ca-bundle-trust-manager.md
```diff
@@ -0,0 +1,98 @@
...
+
+```yaml
+    - secret:
+        name: my-ca-secret        # cert-manager Secret in the cert-manager namespace
+        key: tls.crt              # prefer tls.crt over ca.crt — see note below
+```
+
+> **Which key to use — `tls.crt` or `ca.crt`?**
+>
+> Use `tls.crt`. cert-manager's `ca.crt` field is populated on a best-effort basis and may be incomplete or missing for some issuer types. `tls.crt` contains the full certificate chain. Where possible, include only root certificates in your trust bundle — trusting intermediates turns them into de facto roots, which makes CA rotation unsafe. See the trust-manager docs on [Bundling Intermediates](https://cert-manager.io/docs/trust/trust-manager/#bundling-intermediates) if you need to trim a chain.
```
tkan145: Aren't they different thing? My take on that docs is that you need to provide the full chain with tls.cert/tls.key so that it can create/sign a new CA cert with key ca.crt.

doc/operator-ca-bundle.md
```diff
@@ -0,0 +1,175 @@
...
+
+```bash
+kubectl create configmap threescale-ca-bundle -n <operator-namespace>
+kubectl label configmap threescale-ca-bundle \
+  config.openshift.io/inject-trusted-cabundle=true \
+  -n <operator-namespace>
+```
+
+> **Note:** The CNO-injected bundle contains the proxy CA and system trusted CAs, but does **not** include the ingress router CA. If you need both, use [trust-manager](operator-ca-bundle-trust-manager.md) to combine the ingress CA with the proxy trust bundle.
+
```
tkan145, line 130: Can we move the trust-manager doc here?

controllers/configuration/ca_bundle_watcher.go
borisurbanik: Added a comment - this controller needs to update CA in backup replicas as well as the leader. We're not updating anything in cluster.

doc/operator-ca-bundle.md
borisurbanik: Done - starts on line 54

pkg/controller/helper/threescale_api.go
borisurbanik: I've changed as suggested. The reason used to be stronger in earlier implementations when the client was injected, now it would be just for slightly better readability.

controllers/configuration/ca_bundle_watcher.go
tkan145: again, why do we want this? Only the leader can run at any given time, if a replica is promoted to a leader it will then start the controller work and load the CA. So why do we need to have a replica sitting idle and react to even for no reason?

controllers/configuration/tls_config.go
```diff
@@ -36,12 +36,14 @@ func SetTLSConfig(cfg *tls.Config) {
...
 func GetTLSConfig() *tls.Config {
 	tlsConfigMu.RLock()
-	cfg := currentTLSConfig
-	tlsConfigMu.RUnlock()
-	return cfg
+	defer tlsConfigMu.RUnlock()
+	if currentTLSConfig == nil {
+		return nil
+	}
+	return currentTLSConfig.Clone()
```
tkan145: Why are we returning a clone here?

pkg/controller/helper/threescale_api.go
tkan145: sorry but I don't see any reason to make unnecessary code change to basically do the same thing and now we have to check if tlsConfig is nil in both places.

doc/operator-ca-bundle.md
```diff
@@ -26,26 +26,21 @@ kubectl create configmap threescale-ca-bundle \
...
 
-kubectl create configmap threescale-ca-bundle \
-  --from-file=ca-bundle.crt=/tmp/router-ca.crt \
-  -n <operator-namespace>
-```
+If the replacement certificate is signed by a **public CA** (e.g., Let's Encrypt, DigiCert) that is included in the RHEL base image trust store, no custom CA bundle is needed for the operator.
 
-This covers the common case where the 3scale Admin API is served behind an OpenShift Route signed by the cluster's ingress CA.
+If the replacement certificate is signed by a **corporate or private CA**, you must still provide that CA to the operator using one of the methods on this page. The operator runs in a container and only trusts CAs present in the container image's system trust store or explicitly configured via `threescale-ca-bundle`. Corporate CAs are not automatically injected into operator pods — even when configured at the cluster level via the proxy trust mechanism.
```
tkan145: > Corporate CAs are not automatically injected into operator pods — even when configured at the cluster level via the proxy trust mechanism.

Are we sure about this? AFAIK, the operator CA is managed by the OLM.

doc/operator-ca-bundle.md
```diff
@@ -26,26 +26,21 @@ kubectl create configmap threescale-ca-bundle \
...
-```bash
-oc get secret router-ca -n openshift-ingress-operator \
-  -o jsonpath='{.data.tls\.crt}' | base64 -d > /tmp/router-ca.crt
+The recommended solution is to replace the default ingress certificate with one signed by a trusted CA. See [Replacing the default ingress certificate](https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/security_and_compliance/configuring-certificates#replacing-default-ingress) in the OpenShift documentation (adjust the version in the URL to match your cluster).
 
-kubectl create configmap threescale-ca-bundle \
-  --from-file=ca-bundle.crt=/tmp/router-ca.crt \
-  -n <operator-namespace>
-```
+If the replacement certificate is signed by a **public CA** (e.g., Let's Encrypt, DigiCert) that is included in the RHEL base image trust store, no custom CA bundle is needed for the operator.
```
tkan145: This make it sounds like I need to replace the cert for this to work, why do we need to care about whether the cert is default or new?

doc/operator-ca-bundle.md
borisurbanik: Decided to remove this section - treating of proxy CA is well documented in official docs and it's also documented below from our end with correct scope - simply explaining how to attach this CA.

controllers/configuration/ca_bundle_watcher.go
borisurbanik: After our discussion I have checked on this in code, what is expected from the situation when there are 2 managers starting up in different replicas:

- all caches are started up and synced https://github.com/kubernetes-sigs/controller-runtime/blob/v0.17.2/pkg/manager/internal.go#L409
- this controller will be started and reconcile in all replicas because its runnable is started before leader election https://github.com/kubernetes-sigs/controller-runtime/blob/v0.17.2/pkg/manager/internal.go#L416
- leader election starts after https://github.com/kubernetes-sigs/controller-runtime/blob/v0.17.2/pkg/manager/internal.go#L423

_If we didn't run the CA watcher in all replicas, the cache wouldn't re-sync when the lease expires._ 

The only handling for the leader election (on first acquire or failover) is to start more runnables, the cache isn't re-synced here https://github.com/kubernetes-sigs/controller-runtime/blob/v0.17.2/pkg/manager/internal.go#L568-L571

pkg/controller/helper/threescale_api.go
borisurbanik: Ahh, sorry about this - I have implemented the other suggestion and just assumed that would address this comment as well but this was specifically about removing the indirection, I have inlined the `buildHTTPClient` back the way it was before

Having the package-level tls config was making the code more verbose than it originally  was so I've changed it to only manage the root CA and it fits in the original codebase much better this way.

controllers/configuration/tls_config.go
borisurbanik: It was so that consumers can make modifications to the tls.Config (set the InsecureSkipVerify), but when adding comment for that I realized that is very complicated contract and instead it should either manage all of the tls.Config or none of it. 

I have changed the package to have a RootCA pointer - which isn't using clone as nobody should be modifying that. This also made the client constructor change less into fewer LOC.

controllers/configuration/ca_bundle_watcher.go
tkan145: I'm just curious where you got the information that `the cache isn't re-synced here` from? Because from the source code, I see that they all initialize the cache in the same way and nothing that say only leader is allowed to update the cache? Each replica run a full Manager. This means non-leader pods continuously establish ListAndWatch connections to the API server, it just not execute the actual reconcile loop. 

Also how does it affect our TLS cache? as soon as you run the reconcile loop it will reread the configmap isn't it?

borisurbanik: >I'm just curious where you got the information that the cache isn't re-synced here from? Because from the source code, I see that they all initialize the cache in the same way and nothing that say only leader is allowed to update the cache? Each replica run a full Manager. This means non-leader pods continuously establish ListAndWatch connections to the API server, it just not execute the actual reconcile loop.

Yes exactly, the cache is done as opaque list of runnables, and it's initialized even before leader election starts. 

>Also how does it affect our TLS cache? as soon as you run the reconcile loop it will reread the configmap isn't it?

Yes agreed, the re-sync is irrelevant to the TLS cache because the controller is going to start for the first time after leader election from whatever is the last observed CM. There is still benefit to leaving this enabled for all replicas - the TLS will be warmed up already when the failover happens so any queued processing could start right away.

pkg/controller/helper/threescale_api_test.go
```diff
@@ -1,34 +1,136 @@
...
+	ok(t, err)
+
+	setRootCAsForTest(t, nil)
+
+	c, err := PortaClientFromURL(srvURL, "token", false)
+	ok(t, err)
+
+	_, reqErr := c.ListBackendApis()
+	assert(t, reqErr != nil, "expected TLS certificate error, got nil")
+	assert(t, strings.Contains(reqErr.Error(), "certificate"), "expected certificate error, got: %v", reqErr)
```
tkan145: Hm, what type of error are we expecting here?

test/unitcontrollers/activedoc_controller_test.go
```diff
@@ -0,0 +1,72 @@
...
+	tests := []struct {
+		name           string
+		objects        []runtime.Object
+		conditionCheck func(err error, cl client.Client) bool
+	}{
+		{
+			// ActiveDocReconciler.reconcileSpec() calls LookupProviderAccount and then PortaClientFromAccount after
+			// checkExternalRefs. SystemName must be pre-set so that SetDefaults()
+			// returns false and no early requeue occurs.
+			name: "InvalidCABundle",
```
tkan145: Just curious, why are we only test the invalid CA case (this and other tests) and not valid or skip verify cases?

briangallagher:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/application-services-red-hat-openshift-api-management/pr-checks/1a157d5c-cfcc-4d4d-89af-b99e03dfa3e4?refs=4690e33bce7cabd97ea13dbbb252f3cea0fec3ab&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/application-services-red-hat-openshift-api-management/pr-checks/1a157d5c-cfcc-4d4d-89af-b99e03dfa3e4/license?refs=4690e33bce7cabd97ea13dbbb252f3cea0fec3ab&source=prChecksComment) |

pkg/controller/helper/threescale_api_test.go
```diff
@@ -1,34 +1,137 @@
...
+	t.Cleanup(func() { configuration.SetRootCAs(prev) })
+}
+
+// TestPortaClientFromAccount_InvalidURL verifies that an unparseable admin URL
+// is rejected before any network I/O.
+func TestPortaClientFromAccount_InvalidURL(t *testing.T) {
 	providerAccount := &ProviderAccount{AdminURLStr: ":foo", Token: "some token"}
 	_, err := PortaClient(providerAccount, false)
 	assert(t, err != nil, "error should not be nil")
 }
```
tkan145: What is going on with all of these extra comment? If you are going to test different cases, use table base setup instead of a function for each case.

tkan145 (commented):
Do we really need all of those CR tests and mock just to test API call?

The most important thing is that ca_bundler_watcher is not tested anywhere.

pkg/controller/helper/threescale_api_test.go
borisurbanik: Updated the tests to be table tests 👍 There are 3 tables - one for each Porta constructor variant.

borisurbanik:
> Do we really need all of those CR tests and mock just to test API call?
> 
> The most important thing is that ca_bundler_watcher is not tested anywhere.

As discussed, added the ca_bundler_watcher reconciler test and removed the mocked tests to be replaced in functional test in the future to eliminate the mocks.

controllers/configuration/ca_bundle_watcher.go
```diff
@@ -0,0 +1,130 @@
...
+
+const (
+	CABundleConfigMapName    = "threescale-ca-bundle"
+	CABundleConfigMapKey     = "ca-bundle.crt"
+	errReasonInvalidCAFormat = "InvalidCAFormat"
+)
+
+// CABundleWatcher watches the threescale-ca-bundle ConfigMap and writes the
+// parsed *tls.Config to the package-level variable via SetRootCAs on every
```
tkan145: Says "writes *tls.Config" but it actually returns *x509.CertPool.

controllers/configuration/ca_bundle_watcher.go
```diff
@@ -0,0 +1,130 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+const (
+	CABundleConfigMapName    = "threescale-ca-bundle"
+	CABundleConfigMapKey     = "ca-bundle.crt"
+	errReasonInvalidCAFormat = "InvalidCAFormat"
```
tkan145: Is this necessary? 😅

controllers/configuration/ca_bundle_watcher_test.go
```diff
@@ -0,0 +1,265 @@
...
+fF8LlnjeVx9V3qDG90Pn6Qd4+92hpOK6rbGFAgMBAAGjUzBRMB0GA1UdDgQWBBSA
+LkhTgxGzp3/BQblBQyG+9h2DjjAfBgNVHSMEGDAWgBSALkhTgxGzp3/BQblBQyG+
+9h2DjjAPBgNVHRMBAf8EBTADAQH/MA0GCSqGSIb3DQEBCwUAA4IBAQAo+P1usqGo
+ScTguZ5ASyg8+/0cgH7olhrW/rdlpuqgicZ0wNExMZyUoIqJtUEgVvJECBd7s96I
+aQCwQBAsvVSbtn2IbP4xGZ2BGpjyqSlhjlvwtWi/Vif5IXEWAgCzZQA02li2edz3
+OU1vy7/ImHme7AX59SZU9GnrzyU5Px0kCDTwUISoQXby15Cutyc9Zxx7+yi1X4Of
+XjxYYbHY8P8+yNREe2bApOdsNDPhqxkONN23LRWt7KsXc4Ixyc02vGlalHau+LPy
+ucShOI+5t8rImb6lXsRT9u+0nOCp7282OAWinXwq22AdY9UjfMXR+jfhLwupCvWv
+Iqm1o0IL33Eb
+-----END CERTIFICATE-----
```
tkan145: Use go function to generate temp cert instead of handwritten one here.

controllers/configuration/ca_bundle_watcher_test.go
```diff
@@ -0,0 +1,265 @@
...
+// calls this helper MUST NOT call t.Parallel().  t.Setenv is the mechanism
+// that enforces this: it panics when called from inside a parallel test,
+// making the constraint mechanical rather than advisory.
+func seedRootCAsForTest(t *testing.T) {
+	t.Helper()
+	t.Setenv("_TEST_TLS_GUARD", "1")
+	prev := GetRootCAs()
+	SetRootCAs(x509.NewCertPool())
+	t.Cleanup(func() { SetRootCAs(prev) })
+}
```
tkan145: Similar function exist inside `pkg/controller/helper/threescale_api_test.go`, can we extract it out to helper

controllers/configuration/ca_bundle_watcher_test.go
```diff
@@ -0,0 +1,265 @@
...
+			injectGetErr: true,
+			wantErr:      true,
+			wantErrMsg:   "synthetic server error",
+			check: func(t *testing.T, rec *record.FakeRecorder) {
+				if GetRootCAs() == nil {
+					t.Error("expected GetRootCAs() to remain non-nil (seeded value)")
+				}
+				assertNoEventW(t, rec)
+			},
+		},
```
tkan145, line 152: What does this even do? Why are we injecting random error here?

controllers/configuration/ca_bundle_watcher_test.go
```diff
@@ -0,0 +1,265 @@
...
+// newTestScheme builds a fresh *runtime.Scheme for each test, avoiding mutation
+// of the shared global scheme.Scheme.
+func newTestScheme(t *testing.T) *runtime.Scheme {
+	t.Helper()
+	s := runtime.NewScheme()
+	if err := corev1.AddToScheme(s); err != nil {
+		t.Fatalf("AddToScheme corev1: %v", err)
+	}
+	return s
+}
```
tkan145: Is it necessary to extract this to a function when you only call it once?

controllers/configuration/ca_bundle_watcher_test.go
```diff
@@ -0,0 +1,265 @@
...
+-----END CERTIFICATE-----
+`
+
+func (e *errorInjectingClient) Get(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
+	return e.getErr
+}
+
+func reqForWatcher(ns, name string) reconcile.Request {
+	return reconcile.Request{NamespacedName: types.NamespacedName{Namespace: ns, Name: name}}
+}
```
tkan145: You really like to use function for a single line code 😅

controllers/configuration/ca_bundle_watcher_test.go
```diff
@@ -0,0 +1,265 @@
...
+		},
+		{
+			name:    "ConfigMap present, valid PEM — SetRootCAs(pool), no error, no event",
+			objects: []runtime.Object{validCM},
+			wantErr: false,
+			check: func(t *testing.T, rec *record.FakeRecorder) {
+				if GetRootCAs() == nil {
+					t.Error("expected GetRootCAs() != nil after valid PEM")
+				}
+				assertNoEventW(t, rec)
```
tkan145, line 248: Don't you need to check if the CA has been set up correctly here? Otherwise, how would we know if the key is set with the same content as in the config map?

pkg/controller/helper/threescale_api_test.go
```diff
@@ -1,34 +1,196 @@
...
-	_, err := PortaClientFromURL(url, "some token", false)
-	assert(t, err != nil, "error should not be nil")
+	tests := []struct {
+		name                  string
+		buildURL              func(t *testing.T, srv *httptest.Server) *url.URL
+		insecureSkipVerify    bool
+		setupCA               func(t *testing.T, srv *httptest.Server)
+		wantClientErrContains string
+		wantRequestErr        bool
+		requestErrContains    string
```
tkan145: Can we fold this together? Below we check `if wantClientErrContains != ""` but then for request error we need an extra bool?

test/unitcontrollers/activedoc_controller_test.go
borisurbanik: Most of the tests are now testing valid and insecure cases as well.

controllers/configuration/ca_bundle_watcher.go
borisurbanik: Changed to "CA Pool", knowing the exact struct is not necessary for the overall comment

controllers/configuration/ca_bundle_watcher_test.go
borisurbanik: Done 👍

controllers/configuration/ca_bundle_watcher_test.go
borisurbanik: Extracted to a new testhelper package, we didn't have a test helper package before so it's a bit awkward to only have this one function there. Plus I needed to turn the configuration test into a black-box test.

However, I think this is still the right way (the other way being to have this inlined in tests) - it has a good documentation function, because it shows that we're using package level variable to anyone who checks the test helper package.

controllers/configuration/ca_bundle_watcher_test.go
borisurbanik: To make sure the error handling doesn't by mistake behave the same as "not found".

If "not found" -> reset the package level CA bundle to nil. 
If "any other error" -> requeue (expect error); preserve the CA bundle

controllers/configuration/ca_bundle_watcher_test.go
borisurbanik: It can be inlined - done

controllers/configuration/ca_bundle_watcher_test.go
borisurbanik: Good catch, that's what the test would look like if it was written before implementation :-) Added assertion for the CA to match what's passed to the configmap.

pkg/controller/helper/threescale_api_test.go
borisurbanik: Done

controllers/configuration/ca_bundle_watcher_test.go
borisurbanik: It's because of so many refactors in this PR ;-)

controllers/configuration/ca_bundle_watcher_test.go
tkan145: So, currently, this code doesn't check what you described, right? We pass in an empty object and then compare the CA to nil (the default value); it doesn't check if the CA is still intact if provided, or whether the operator reverts it to nil.

borisurbanik: Partially - the test body always seeds an initial CA pool that's not nil - the test above was testing that when NotFound error is returned by API, the pool ends up nil. This test was simply asserting the path from that test was not taken (the pool will not be set to nil), but yes, it wasn't checking the positive case that the pool was preserved. 

I have added verification of seedPool in each test that was meant to preserve the seed pool (which is all of them, except the valid case, missing key or missing CM). 

As always, thank you for the review and for checking the tests!

tkan145 (approved):

PR #1182, opened 2026-06-29
(chore) Add basic .bobignore

borisurbanik:
Add a basic `.bobignore` file.

The purpose of this file is to:

- exclude sensitive files
- filter what is included in context to reduce token usage - build artifacts, images, large files

Note this only hides the files from tools that support it, it is not 100% guaranteed that files won't leak to the LLM

### DIFF
--- a/.bobignore
+++ b/.bobignore
@@ -1,0 +1,40 @@
+# Sensitive information
+.env
+secrets/
+*password*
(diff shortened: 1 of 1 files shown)


tkan145:
As discussed, we don't want this. Closing

PR #1190, opened 2026-08-28
THREESCALE-15877: Fix File.exists? in a default system ConfigMap

mdujava:
With upgrade to Ruby 3.3, `File.exists` was deprecated.

Reference: https://github.com/3scale/porta/pull/4129

### DIFF
--- a/pkg/3scale/amp/component/system.go
+++ b/pkg/3scale/amp/component/system.go
@@ -1458,7 +1458,7 @@ func (system *System) getSystemRollingUpdatesConfData() string {
 
 func (system *System) getSystemServiceDiscoveryData() string {
 	return `production:
-  enabled: <%= cluster_token_file_exists = File.exists?(cluster_token_file_path = '/var/run/secrets/kubernetes.io/serviceaccount/token') %>
+  enabled: <%= cluster_token_file_exists = File.exist?(cluster_token_file_path = '/var/run/secrets/kubernetes.io/serviceaccount/token') %>
   server_scheme: 'https'
   server_host: 'kubernetes.default.svc.cluster.local'
   server_port: 443


briangallagher:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/application-services-red-hat-openshift-api-management/pr-checks/523cf66f-952a-42eb-a099-5cad86b09457?refs=a615c8e5ff77495660a9f5d7d7e6a803d7287a15&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/application-services-red-hat-openshift-api-management/pr-checks/523cf66f-952a-42eb-a099-5cad86b09457/license?refs=a615c8e5ff77495660a9f5d7d7e6a803d7287a15&source=prChecksComment) |

borisurbanik (approved):
@mdujava the fix here LGTM for new installation

I had a look at the JIRA issue, it seems to come from alpha testing.

From another look at the reconciler - and when you update the "model" here - so on new deployment you will get the right configmap, but the reconciler is very specific for this system map and it only handles creating configmap where it doesn't exist, or updating Zync endpoint (see https://github.com/3scale/3scale-operator/blob/18b6909ab6c1df45db4498f173d4021e91df45fa/pkg/3scale/amp/operator/system_reconciler.go#L543-L592)

I can approve this change to unblock testing - but it would be ideal to address also the upgrade path in the same PR.

mdujava:
> but it would be ideal to address also the upgrade path in the same PR.

Do you think its safe to update this automatically or would we require manual step instead?

borisurbanik:
> > but it would be ideal to address also the upgrade path in the same PR.
> 
> Do you think its safe to update this automatically or would we require manual step instead?

This file looks like it's a template that needs to be filled out by customers to make it fully work - does our test fixture do any modifications to it after deployment / would that form normally work even without customers setting up client_id/secret for example? Could you test whether the functionality changes for customers that have the default configmap who never edited it?

The [docs on this](https://docs.redhat.com/en/documentation/red_hat_3scale_api_management/2.16/html/admin_portal_guide/service-discovery_service-discovery) instruct customers go generate a configuration file that doesn't have any of this so I'd say manual note should be sufficient.

